PR #300, opened 2013-05-14
Add 2 new TCP flags per RFC 3168.

delphij:
Add ECE, CWR and NS bits for TCP header, defined in RFC 3168 and RFC 3540.


mcr:
What I got from this discussion was that we can have ECE and CWR for free, but that the 8th bit presents a problem.  Would you like to split this pull request (and rebase it)?

delphij:
Rebased and amended to include RFC 3168 only.

This should read:

"Add ECE and CWR bits for TCP header, defined in RFC 3168."

infrastation:
`tcp-ece` and `tcp-cwr` are in the master branch now, thank you.

infrastation:
I would suggest to discuss `tcp-ns` on tcpdump-workers first.

PR #320, opened 2013-09-19
fix typo on manpage install

busterb:
This causes errors when doing make install twice because the manpage is not properly removed.


mcr:
We drifted from the question of the man pages.
I agree that it's annoying, the alternative is that people can not compile because they can not install the "right" autoconf tools.  I am open to a post-commit hook, but given that development is now on github, that won't work.
I am manually pulling this up and closing it.

mcr:
This tree contains some 32/64bit tpacket fixes, which I've omitted.

mcr:
I can not find any unmerged pieces from this pull request.
The 32-bit userland on 64-bit kernel is likely still an issue though.

busterb:
Sorry for the scope leakage - I should have submitted this PR from a feature branch.

PR #358, opened 2014-06-05
Add the --disable-packet-ring configure option.

bonsaiviking:
This allows the creation of 32-bit binaries that are compatible with
64-bit Linux kernels before version 2.6.27.

Original patch credit: David Fifield of the Nmap project
(https://svn.nmap.org/nmap/libpcap/NMAP_MODIFICATIONS/0002-Add-the-disable-packet-ring-configure-option.patch)

### DIFF
--- a/configure.in
+++ b/configure.in
@@ -1655,6 +1655,24 @@ linux*)
 	;;
 esac
 
+dnl The packet ring capture facility of Linux, described in
+dnl Documentation/networking/packet_mmap.txt, is not 32/64-bit compatible before
+dnl version 2.6.27. A 32-bit kernel requires a 32-bit userland, and likewise for
+dnl 64-bit. The effect of this is that a 32-bit libpcap binary will not run
+dnl correctly on a 64-bit kernel (the binary will use the wrong offsets into a
+dnl kernel struct). This problem was solved in Linux 2.6.27. Use
+dnl --disable-packet-ring whenever a 32-bit application must run on a 64-bit
+dnl target host, and either the build host or the target host run Linux 2.6.26
+dnl or earlier.
+AC_ARG_ENABLE([packet-ring],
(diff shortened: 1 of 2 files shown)


fxlb:
@bonsaiviking: After the merge we get on Travis:
```
./pcap-linux.c:5659:1: warning: ‘iface_get_offload’ defined but not used [-Wunused-function]
```
Could you have a look?

guyharris:
`iface_get_offload()`, and the stuff it calls, isn't needed if we aren't using the TPACKET code; I've checked in a change not to compile if if `HAVE_PACKET_RING` isn't defined (whether that's because the user explicitly disabled it *or* if you're building against a too-old kernel).

*However*, that shouldn't be happening on Travis, as the default *should* be "use TPACKET".

dmiller-nmap:
@fxlb Sorry I took so long to get back to you. Looks like it's taken care of in c2c1d3f32d20ad1b899223cd623cb65c53c67371

guyharris:
Note that a number of changes:

    6069bd00dba709733be49d2e6785781d59b17a3b

    35c404821c70b8b48d9d8a9f4066db98d878da7e

    3fcb830b7609225b93fce206a3855004ce3b9ee2

    1573a7360972ec845854f2b49ed3ce6b0c6aa1b5

    7beef37c555bff4003c8532195841b663fcab6d7

    03baffab46e1fd3bde48d8d70c7e62d7f6a231c2

should handle the "So are there any APIs that let userland code determine whether it's running on a 32-bit or 64-bit kernel (preferably non-x86-specific APIs)? If so, it might be possible to have the code compensate for the problems with TPACKET_V1."  The API in question is uname() and it isn't ideal as it requires libpcap to know, for instruction sets that have both 32-bit and 64-bit variants and that were both supported by Linux prior to 2.6.27, what uname returns for the 32-bit and 64-bit version, and relies on uname() not to lie and report to 32-bit programs that the system's instruction set is the 32-bit variant, but it should handle at least some of the cases.

PR #359, opened 2014-06-05
Move the inclusion of config.h to the top of scanner.l.

bonsaiviking:
This works around _LARGE_FILES difficulties on AIX. See
http://seclists.org/nmap-dev/2012/q1/459 for example.

Original patch credit David Fifield of the Nmap project
(https://svn.nmap.org/nmap/libpcap/NMAP_MODIFICATIONS/pre-configure.patch)


guyharris:
I decided that, for various other reasons, it's time to start requiring newer versions of Flex, so I backed out #360 and merged this one instead.

PR #370, opened 2014-07-29
Update config

LocutusOfBorg:
I updated the config files with the newer autoconf and automake, to support recent architectures


mcr:
which version of autoconf did you use?
I will update with autoconf 2.69 on our bpf.tcpdump.org, since we need to be able to do this from known version.

mcr:
It appears that you must have used a version of autoconf newer than what's in debian stable, so I won't be updating it, as I won't be able to maintain it easily.

LocutusOfBorg:
I used the version of debian/testing, that is the same 2.69 as the debian/stable one just one is 2.69-1 and the testing one is 2.69-8.

I presume you are talking about autotools-dev, not autoconf.

I'm using also the debian/testing version and that "testing" is gonna be called "jessie" in a few months (it is already an RC)

https://packages.qa.debian.org/a/autotools-dev.html

fxlb:
I got `checking build system type... ./config.guess: unable to guess system type` on:
```
UNAME_MACHINE = ppc64le
UNAME_RELEASE = 3.17.4-301.fc21.ppc64le
UNAME_SYSTEM  = Linux
UNAME_VERSION = #1 SMP Mon Dec 1 07:51:01 UTC 2014
```
Thus, I commit test branches on my forks with updated `config.{sub,guess}`, timestamp=2015-01-01, for libpcap and tcpdump.
Tests on Travis are OK.
I am going to realize tests on various systems before a *possible* merge in the master branch.
Feel free to test these branches ...
libpcap:  git clone --branch config https://github.com/fxlb/libpcap.git libpcap-config
tcpdump: git clone --branch config https://github.com/fxlb/tcpdump.git tcpdump-config

fxlb:
The `config.{sub,guess}` used are the latest.

fxlb:
Systems tested OK (for now):
```
i386-pc-solaris2.9
i386-pc-solaris2.10
i386-pc-solaris2.11
sparc-sun-solaris2.9
sparc-sun-solaris2.10
sparc-sun-solaris2.11
powerpc64-unknown-linux-gnu
powerpc64le-unknown-linux-gnu (the system where the problem was detected)
x86_64-unknown-netbsd5.1
```

fxlb:
Does anybody think that the update of these 2 files (`config.{sub,guess}`) could be a problem?

fxlb:
Updates done.
@LocutusOfBorg Thanks for pointing this out.

LocutusOfBorg:
thanks to you!

PR #371, opened 2014-07-29
Add hurd support, from debian

LocutusOfBorg:
This is the hurd support, from debian patch

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -657,6 +657,7 @@ else
 	AC_CHECK_HEADERS(net/nit.h sys/net/nit.h)
 	AC_CHECK_HEADERS(linux/socket.h net/raw.h sys/dlpi.h)
 	AC_CHECK_HEADERS(config/HaikuConfig.h)
+	AC_CHECK_HEADERS(hurd.h)
 
 	if test "$ac_cv_lbl_bpf_h_defines_biocsetif" = yes; then
 		#
@@ -718,6 +719,11 @@ else
 		# Haiku.
 		#
 		V_PCAP=haiku
+	elif test "$ac_cv_header_hurd_h" = yes; then
+		#
+		# Hurd.
+		#
+		V_PCAP=hurd
 	else
 		#
 		# Nothing we support.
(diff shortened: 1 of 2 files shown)


mcr:
Does this patch depend upon a new autoconf?

LocutusOfBorg:
nope, it should work also without the new autoconf

guyharris:
From a quick look at the current Hurd code in Git, the return value of a BPF filter is used only as a pass/fail indication; it doesn't limit the amount of packet data copied.  For now, it's probably best to use `MAXIMUM_SNAPLEN` rather than 1500 in the filter.

infrastation:
@LocutusOfBorg, would you like to rebase these changes on the current master branch?

infrastation:
Thank you. Do not mind FreeBSD and illumos failures, which are caused by recent Valgrind changes. To try the new code, is the simplest way to setup a VM as described [here](https://www.gnu.org/software/hurd/hurd/running/qemu.html)?

LocutusOfBorg:
Hello, it should be, but also going on irc OFTC/#debian-hurd helps :)

sthibaul:
FI, it builds fine on my hurd box, provided the additional change:

[patch.txt](https://github.com/the-tcpdump-group/libpcap/files/9600838/patch.txt)

sthibaul:
(and autoreconf of course)

sthibaul:
otherwise one gets:

```
gcc  -shared libpcap.so.$MAJOR_VER \
   -o libpcap.so.$VER pcap-hurd.o fad-getad.o  pcap-dbus.o  pcap.o gencode.o optimize.o nametoaddr.o etherent.o fmtutils.o savefile.o sf-pcap.o sf-pcapng.o pcap-common.o bpf_image.o bpf_filter.o bpf_dump.o scanner.o grammar.o strlcat.o strlcpy.o  -lrt -ldbus-1
/usr/bin/ld: cannot find libpcap.so.1: No such file or directory
```

because it's missing the `-soname` parameter

infrastation:
It took 8 years to resolve this due to a number of reasons, one of which is that none of the project maintainers have a working GNU Hurd system.  I tried to rig one up today, but without success.  However, @sthibaul has compile-tested the code and confirmed that the current libpcap binary package (`libpcap0.8:hurd-i386 1.10.1-4`) is functional for the current tcpdump binary package, so the original patch (which has been stable since 2014) and the two follow-up improvements are now in the master branch.

guyharris:
This didn't include CMake support  I took the configure.ac changes and tried to translate them to CMake in #1267 - new platform changes should support both configuration mechanisms.  Please test that as well.

PR #376, opened 2014-08-06
Add with-libnl switch to configure so that libnl may be found in other locations for cross-compiles

mcr:


mcr:
I developed this patch for a Linux-based appliance that was based upon a system that predates libnl3, so I had to add it, so that's almost the same thing...

PR #378, opened 2014-08-19
Add tstamp_type/precision support for DAG capture

sfd:
Add tstamp_precision and tstamp_type support to pcap-dag.c.

Update contact details.

The DAG hardware clock may or may not be synchronsied to either the host clock or an external clock reference, so we can't determine which of PCAP_TSTAMP_ADAPTER or PCAP_TSTAMP_ADAPTER_UNSYNCED is accurate. Either will produce the same result.

The only potential concern is that we don't advertise PCAP_TSTAMP_HOST_* so anyone requesting those explicitly will not be able to capture.

### DIFF
--- a/pcap-dag.c
+++ b/pcap-dag.c
@@ -1,5 +1,5 @@
 /*
- * pcap-dag.c: Packet capture interface for Endace DAG card.
+ * pcap-dag.c: Packet capture interface for Emulex EndaceDAG cards.
  *
  * The functionality of this code attempts to mimic that of pcap-linux as much
  * as possible.  This code is compiled in several different ways depending on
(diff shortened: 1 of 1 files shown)


guyharris:
Well, GitHub lies when they [claim that you can "review the cumulative diff of all file changes across all commits"](https://help.github.com/articles/using-pull-requests/#reviewing-proposed-changes) - there's nothing obvious to click on to do that - so the easiest way to review is multi-commit pull request is to just merge it and then do whatever changes you deem appropriate right after that.  I'll look at this when it pops up on bpf.tcpdump.org.

sfd:
Doesn't the top level of the pull request show you the cumulative commit changes?

In any case thanks for the merge, appreciated.

guyharris:
By "the top level of the pull request" do you mean "the 'Files changed' tab for the page for the pull request"?  Not exactly the most intuitive interface....

(Although part of the issue is that this pull request does at least 3 different things - adds support for a description string for DAG devices, fixes a packet-parsing bug, and adds time stamp resolution support; if those had been separate pull requests, it would have been a bit clearer.)

PR #404, opened 2015-01-09
Use size_t for buffer sizes instead of int

jer-gentoo:
I'm seeing all kinds of problems on a PARISC system and I'm not here to solve all of them. Running tcpdump crashes the system (a kernel problem not fixed here, of course) on a 64-bit system with a 32-bit userland and with more than 4 gigabytes RAM. The kernel does not crash with only 2 gigabytes of RAM, but does give a clue as pcap prints:
corrupted frame on kernel ring mac offset 23712 + caplen -1163995555 > frame len 262144

Obviously caplen should not be negative.

Oh, did I mention that Linux on PARISC64 doesn't merely use a 32-bit userland (which very likely complicates talking to the kernel and interpreting what it returns) but that it's also big-endian (along with HP-UX) and that its stack direction is up, not down (as with HP-UX)?

With these changes, what it at least gets right is caplen:
corrupted frame on kernel ring mac offset 9562 + caplen 839049610 > frame len 262144

even if caplen is still way off the chart.

That aside, tcpdump just keeps thrashing it until the kernel dies a sorry death. wireshark neatly stops capturing when this happens. I haven't looked into the actual problem much yet, the one where caplen gets mangled beyond recognition, and I suspect that's hidden in pcap-linux.c somewhere.

### DIFF
--- a/pcap-int.h
+++ b/pcap-int.h
@@ -161,7 +161,7 @@ struct pcap {
 	/*
 	 * Read buffer.
 	 */
-	int bufsize;
+	size_t bufsize;
 	u_char *buffer;
 	u_char *bp;
 	int cc;
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -298,7 +298,7 @@ struct pcap_linux {
 	size_t	mmapbuflen;	/* size of region */
 	int	vlan_offset;	/* offset at which to insert vlan tags; if -1, don't insert */
 	u_int	tp_version;	/* version of tpacket_hdr for mmaped ring */
-	u_int	tp_hdrlen;	/* hdrlen of tpacket_hdr for mmaped ring */
+	size_t	tp_hdrlen;	/* hdrlen of tpacket_hdr for mmaped ring */
 	u_char	*oneshot_buffer; /* buffer for copy of packet */
 #ifdef HAVE_TPACKET3
 	unsigned char *current_packet; /* Current packet within the TPACKET_V3 block. Move to next block if NULL. */
@@ -4220,7 +4220,7 @@ static int pcap_handle_packet_mmap(
 	if (tp_mac + tp_snaplen > handle->bufsize) {
 		snprintf(handle->errbuf, PCAP_ERRBUF_SIZE,
 			"corrupted frame on kernel ring mac "
-			"offset %d + caplen %d > frame len %d",
+			"offset %d + caplen %d > frame len %zd",
 			tp_mac, tp_snaplen, handle->bufsize);
 		return -1;
 	}
--- a/pcap-win32.c
+++ b/pcap-win32.c
@@ -149,7 +149,7 @@ pcap_stats_win32(pcap_t *p, struct pcap_stat *ps)
 
 /* Set the dimension of the kernel-level capture buffer */
 static int
-pcap_setbuff_win32(pcap_t *p, int dim)
+pcap_setbuff_win32(pcap_t *p, size_t dim)
 {
 	if(PacketSetBuff(p->adapter,dim)==FALSE)
 	{
--- a/pcap.c
+++ b/pcap.c
@@ -659,7 +659,7 @@ pcap_set_immediate_mode(pcap_t *p, int immediate)
 }
 
 int
-pcap_set_buffer_size(pcap_t *p, int buffer_size)
+pcap_set_buffer_size(pcap_t *p, size_t buffer_size)
 {
 	if (pcap_check_activated(p))
 		return (PCAP_ERROR_ACTIVATED);
@@ -1656,7 +1656,7 @@ pcap_setmode_dead(pcap_t *p, int mode)
 }
 
 int
-pcap_setmintocopy(pcap_t *p, int size)
+pcap_setmintocopy(pcap_t *p, size_t size)
 {
 	return (p->setmintocopy_op(p, size));
 }
@@ -1668,7 +1668,7 @@ pcap_get_adapter(pcap_t *p)
 }
 
 static int
-pcap_setmintocopy_dead(pcap_t *p, int size)
+pcap_setmintocopy_dead(pcap_t *p, size_t size)
 {
 	snprintf(p->errbuf, PCAP_ERRBUF_SIZE,
 	    "The mintocopy parameter cannot be set on a pcap_open_dead pcap_t");
@@ -1849,7 +1849,7 @@ pcap_open_dead(int linktype, int snaplen)
  * XXX - what if we get a short write?
  */
 int
-pcap_sendpacket(pcap_t *p, const u_char *buf, int size)
+pcap_sendpacket(pcap_t *p, const u_char *buf, size_t size)
 {
 	if (p->inject_op(p, buf, size) == -1)
 		return (-1);
--- a/pcap/pcap.h
+++ b/pcap/pcap.h
@@ -282,7 +282,7 @@ int	pcap_set_rfmon(pcap_t *, int);
 int	pcap_set_timeout(pcap_t *, int);
 int	pcap_set_tstamp_type(pcap_t *, int);
 int	pcap_set_immediate_mode(pcap_t *, int);
-int	pcap_set_buffer_size(pcap_t *, int);
+int	pcap_set_buffer_size(pcap_t *, size_t);
 int	pcap_set_tstamp_precision(pcap_t *, int);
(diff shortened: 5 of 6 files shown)


guyharris:
Changing the data types in APIs changes the ABI of libpcap, so that cannot and will not be done.  (There are OSes that offer libpcap as a documented system library and that have third-party applications expected not to fail on future OSes.)

The problem with "corrupted frame on kernel ring mac offset 23712 + caplen -1163995555 > frame len 262144" is that it's printing an unsigned value with %d; I've checked in a fix for that (printing both the MAC offset and the captured length with %u) as fc17773272c10a4da045b84013ff3bdc3eb31b61 and backported it all the way back to the libpcap 1.1 branch (for the benefit of proudly trailing-edge Linuxes).

Neither the MAC offset nor the packet length look at all valid, so, yes, there is an underlying problem, and *that's* what needs to be fixed.

> Oh, did I mention that Linux on PARISC64 doesn't merely use a 32-bit userland (which very likely complicates talking to the kernel and interpreting what it returns)

I would hope that it would support both 32-bit and 64-bit userlands.

The first version of the "turbopacket" memory-mapped capturing code, TPACKET_V1, shared a data structure between the kernel and userland that had an `unsigned long` as the first member (see `struct tpacket_hdr` in `if_packet.h`), which means that if the kernel and userland aren't both ILP32 or both LP64, the structures will be laid out differently between the kernel and userland, and the userland won't work.  Unfortunately, there doesn't seem to be any good way for userland to figure out the kernel's programming model and compensate for that.

TPACKET_V2 changed that so that the _uN types are used, and the kernel and userland structures match (see `struct tpacket2_hdr` in `if_packet.h`).

If your kernel supports only TPACKET_V1, and you have a 64-bit kernel and 32-bit userland, Bad Things Will Happen with packet capture (and not just on PA-RISC or on big-endian machines in general).

Neither the endianness nor the stack growth depth should matter here.

jer-gentoo:
There is  no 64-bit userland for Linux on HPPA. Nobody ever sat down to write it.

guyharris:
("It" presumably meaning "the assembler-language code, and what few bits of non-portable C code, necessary to make it work, along with whatever configure script etc. changes are needed", as that's all that should be necessary to make it work.)

So that means you have a 64-bit kernel and 32-bit userland.  What kernel version are you using?  2.6.27 introduced TPACKET_V2, so with a 2.6.27 or later kernel, non-ancient libpcap should use TPACKET_V2, with the kernel and userland data structures being identical.

jer-gentoo:
Linux 3.17.7 / parisc64 PA8800 (Mako) 9000/785/C8000 GNU/Linux

It should be using TPACKET_V2/3.

jer-gentoo:
I'm using git head libpcap and seeing slightly nicer error messages (this is from wireshark/dumpcap):

10:36:18.664  Dbg  FIX: vsimple_dialog t: 3 bm: 1 m: Error while capturing packets: corrupted frame on kernel ring mac offset 51490 + caplen 928074546 &gt; frame len 262144

txgcwm:
corrupted frame on kernel ring mac offset 61396 + caplen 3053520940 > frame len 262144

I really can‘t understand why tp_snaplen so big,  somethings error in kernel?

infrastation:
@jer-gentoo, are you willing to troubleshoot this bug any more?

jer-gentoo:
> corrupted frame on kernel ring mac offset 61396 + caplen 3053520940 > frame len 262144
> 
> I really can‘t understand why tp_snaplen so big, somethings error in kernel?

PARISC is a bigendian architecture. Additionally its stack grows up. Perhaps this helps. I assume some calculation on tp_snaplen is done in a non-portable way.

guyharris:
> > corrupted frame on kernel ring mac offset 61396 + caplen 3053520940 > frame len 262144
> > I really can‘t understand why tp_snaplen so big, somethings error in kernel?
> 
> PARISC is a bigendian architecture.

So are:

- PowerPC/Power ISA, except for ppc64le;
- System/390 and z/Architecture;
- MIPS non-le;
- SPARC;

etc., and all of them run Linux, so it's not as if PA-RISC is the first big-endian architecture on which the memory-mapped capture code runs.

One difference, though, is that, as far as I know, there is a 64-bit userland for all of those. 

> Additionally its stack grows up. Perhaps this helps. I assume some calculation on tp_snaplen is done in a non-portable way.

The `tp_snaplen` member of the turbopacket header is set by `tpacket_rcv()` to the value of `skb->len`, where `skb` is an skbuff pointer that's the first argument to `tpacket_rcv()`.  It is later set to the value of `res` if it's greater than `res`; `res` is the return value of `run_filter()` - this implements the "the return value of the filter program is the maximum number of bytes of packet data to capture" behavior.  After that, there's some code that does

	if (po->tp_version <= TPACKET_V2) {
		if (macoff + snaplen > po->rx_ring.frame_size) {
			if (po->copy_thresh &&
			    atomic_read(&sk->sk_rmem_alloc) < sk->sk_rcvbuf) {
				if (skb_shared(skb)) {
					copy_skb = skb_clone(skb, GFP_ATOMIC);
				} else {
					copy_skb = skb_get(skb);
					skb_head = skb->data;
				}
				if (copy_skb)
					skb_set_owner_r(copy_skb, sk);
			}
			snaplen = po->rx_ring.frame_size - macoff;
			if ((int)snaplen < 0) {
				snaplen = 0;
				do_vnet = false;
			}
		}
	} else if (unlikely(macoff + snaplen >
			    GET_PBDQC_FROM_RB(&po->rx_ring)->max_frame_len)) {
		u32 nval;

		nval = GET_PBDQC_FROM_RB(&po->rx_ring)->max_frame_len - macoff;
		pr_err_once("tpacket_rcv: packet too big, clamped from %u to %u. macoff=%u\n",
			    snaplen, nval, macoff);
		snaplen = nval;
		if (unlikely((int)snaplen < 0)) {
			snaplen = 0;
			macoff = GET_PBDQC_FROM_RB(&po->rx_ring)->max_frame_len;
			do_vnet = false;
		}
	}

The machine code generated by the compiler for all that code is obviously non-portable - PA-RISC code is unlikely to run on an x86 processor without binary-to-binary translation - but the C code doesn't look particularly unportable, unless the code that sets the length field of the skbuff is unportable.

jer-gentoo:
> > PARISC is a bigendian architecture.
> 
> So are:
> 
> * PowerPC/Power ISA, except for ppc64le;
> * System/390 and z/Architecture;
> * MIPS non-le;
> * SPARC;
> 
> etc., and all of them run Linux, so it's not as if PA-RISC is the first big-endian architecture on which the memory-mapped capture code runs.
> 
> One difference, though, is that, as far as I know, there is a 64-bit userland for all of those.

Forgot to mention that as the third major difference with many other operating systems. ISTR that some of the operating systems based on the architectures mentioned above also have a 32-bit userland with a 64-bit kernel. Perhaps they convert some of the system calls in the kernel?

> The machine code generated by the compiler for all that code is obviously non-portable

Obviously. :-)

infrastation:
Which of the following can you provide?
* exact steps to reproduce, including OS, compiler, tcpdump and libpcap versions
* stack trace, core file or the likes
* access to the machine
* confirmation whether the problem reproduces with the latest master branches of libpcap and tcpdump
* git bisection commit id that fixed the problem, if it does not reproduce on new code and does reproduce on old code

guyharris:
> > > PARISC is a bigendian architecture.
> > 
> > 
> > So are:
> > 
> > * PowerPC/Power ISA, except for ppc64le;
> > * System/390 and z/Architecture;
> > * MIPS non-le;
> > * SPARC;
> > 
> > etc., and all of them run Linux, so it's not as if PA-RISC is the first big-endian architecture on which the memory-mapped capture code runs.
> > One difference, though, is that, as far as I know, there is a 64-bit userland for all of those.
> 
> Forgot to mention that as the third major difference with many other operating systems. ISTR that some of the operating systems based on the architectures mentioned above also have a 32-bit userland with a 64-bit kernel.

I don't believe any of those, with the *possible* exception of MIPS non-le, have *only* a 32-bit userland.  (I'm assuming here that "the operating systems" means "the Linux distributions"; the only one of those platforms that I know of where a *non-Linux-based OS* exists that has anything 32-bit-only about its userland is z/Architecture, where z/OS supports only 32-bit *code* addressing, but AIX, z/OS, and Solaris don't count here.)

PR #407, opened 2015-01-12
Add support for MPTCP filtering

gdetal:
Multipath TCP (MPTCP) is a standardized extension to TCP (see http://tools.ietf.org/html/rfc6824).This pull request add the support for filtering on packets containing at least one MPTCP option.

The BPF code generated is complex and large due to the limitations of BPF (i.e. no loop) so any feedback on how to reduce the code is welcome !

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -277,6 +277,8 @@ struct block *gen_portop6(int, int, int);
 static struct block *gen_port6(int, int, int);
 struct block *gen_portrangeop6(int, int, int, int);
 static struct block *gen_portrange6(int, int, int, int);
+static struct block *gen_tcpopt(int, int);
+static struct block *gen_mptcp(void);
 static int lookup_proto(const char *, int);
 static struct block *gen_protochain(int, int, int);
 static struct block *gen_proto(int, int, int);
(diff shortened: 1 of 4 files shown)


fxlb:
@gdetal, This PR adds two warnings:
```
./gencode.c: In function ‘gen_tcpopt.constprop.9’:
./gencode.c:5145:9: warning: ‘b0’ may be used uninitialized in this function [-Wuninitialized]
./gencode.c:4999:9: warning: ‘s0’ may be used uninitialized in this function [-Wuninitialized]
```
Could you check?

gdetal:
@fxlb thanks! I've updated my patch such that there shouldn't be a warning anymore (I haven't been able to test as my compiler don't show this warning even with `-Wall`).
y patch such that there shouldn't be a warning anymore (I haven't been able to test as my compiler don't show this warning even with `-Wall`).

fxlb:
to get more warnings, if the compiler handle them: 
```
touch .devel
./configure
make
```

fxlb:
Thanks for the updated patch.
I leave a new review to developers having a thorough knowledge of libpcap.

mcr:
You have/will provided a pcap with example packets for tcpdump's make check?

mcr:
I merged this, but it resulted in:
     ./gencode.c: In function 'gen_tcpopt':
     ./gencode.c:4865:12: error: 'off_macpl' undeclared (first use in this function)
     ./gencode.c:4865:12: note: each undeclared identifier is reported only once for each function it appears in

I have reverted the merge.

guyharris:
Yes, this will need to be updated to match the current gencode.c, which has been modified to be a bit more general when handling encapsulations and to have slightly less confusing variable names.

That might be a bit of work; note that many of these offsets are *not* constant, so you can't just stuff them into the `k` field of a BPF instruction (so code that was doing that was wrong even *before* the changes to gencode.c).

gdetal:
I'll update the pull request.

@guyharris can you point me to specific functions that I need to look at ? Basically, I need to maintain a pointer to the start and end of the TCP option space. I can easily retrieve the TCP data offset, IP header len but what should I use to retrieve the LL header len ?

PR #410, opened 2015-01-27
Corrective: pcap-linux.c

peugnezb:
If tp version is not valid, then we don't enter in the switch condition and we access to an uninitialized variable (req.tp_frame_size).


infrastation:
Superseded by #414.

PR #411, opened 2015-01-31
Add support for Geneve (Generic Network Virtualization Encapsulation).

jessegross:
"geneve" can be used to filter on Geneve encapsulated packets. It
also allows later filters to look at frames inside the tunnel
headers.

The Geneve protocol is documented here:
http://tools.ietf.org/html/draft-gross-geneve-02

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -195,6 +195,7 @@ enum e_offrel {
 	OR_PREVLINKHDR,		/* previous link-layer header */
 	OR_LLC,			/* 802.2 LLC header */
 	OR_PREVMPLSHDR,		/* previous MPLS header */
+	OR_LINKTYPE,		/* link-layer type */
 	OR_LINKPL,		/* link-layer payload */
 	OR_LINKPL_NOSNAP,	/* link-layer payload, with no SNAP header at the link layer */
 	OR_TRAN_IPV4,		/* transport-layer header, with IPv4 network layer */
@@ -834,10 +835,8 @@ static bpf_abs_offset off_linkpl;
 
 /*
  * "off_linktype" is the offset to information in the link-layer header
- * giving the packet type.  This offset is relative to the beginning
- * of the link-layer header - i.e., it doesn't include off_linkhdr.constant_part - so
- * loads with an offset that includes "off_linktype" should use
- * OR_LINKHDR.
+ * giving the packet type. This is an absolute offset from the beginning
+ * of the packet.
  *
  * For Ethernet, it's the offset of the Ethernet type field; this
  * means that it must have a value that skips VLAN tags.
@@ -854,9 +853,10 @@ static bpf_abs_offset off_linkpl;
  *
  * For Linux cooked sockets, it's the offset of the type field.
  *
- * It's set to -1 for no encapsulation, in which case, IP is assumed.
+ * off_linktype.constant_part is set to -1 for no encapsulation,
+ * in which case, IP is assumed.
  */
-static u_int off_linktype;
+static bpf_abs_offset off_linktype;
 
 /*
  * TRUE if the link layer includes an ATM pseudo-header.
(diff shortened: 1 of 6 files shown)


mcr:
Will you provide some pcaps and test cases fro the tcpdump repo, in order to test this code?

jessegross:
I sent a pull request for tcpdump with expanded test cases for Geneve to exercise this (https://github.com/the-tcpdump-group/tcpdump/pull/424).

I noticed a bug in my implementation here while doing that so I also updated this pull request as well.

mcr:
Jesse Gross <notifications@github.com> wrote:
    > I noticed a bug in my implementation here while doing that so I also
    > updated this pull request as well.

Cool!

gencode.c
```diff
@@ -1511,7 +1544,7 @@ gen_load_llrel(offset, size)
 		 * do an indirect load, to use the X register as an offset.
 		 */
 		s2 = new_stmt(BPF_LD|BPF_IND|size);
-		s2->s.k = offset;
+		s2->s.k = offset + off_ll;
```
guyharris: "off_ll", in this case, is presumably not the offset of the beginning of the link-layer header from the beginning of the raw packet data, as said offset is in the X register.  A different variable should be used there, even if that means that a new variable is introduced.

gencode.c
```diff
@@ -1552,7 +1585,7 @@ gen_load_macplrel(offset, size)
 		 * as an offset.
 		 */
 		s2 = new_stmt(BPF_LD|BPF_IND|size);
-		s2->s.k = offset;
+		s2->s.k = off_macpl + offset;
```
guyharris: `off_macpl` is the offset of the MAC-layer *payload*, and points *past* the link-layer header, so presumably this should also be a different variable.

jessegross: Perhaps I should have pulled this out into a separate commit but this is actually fixing a bug with all variable length headers and should not change the intended meaning of what is there currently.

The problem is that some filters (like VLAN) increment off_macpl as they are parsed. In the fixed MAC header case, this is handled properly since it is just incorporated into the header length. However, currently with variable length headers we only look at the variable component and ignore the fixed part, so the VLAN stack can't be handled. This is changing it so that we look at both pieces for variable headers; in cases where there are no VLAN filters, off_macpl is zero so things still work.

I don't know of any filter that increments off_ll but it seemed right to apply the same logic there.

jessegross: It's not hard to move it to a separate variable, it just wasn't clear to me whether that was actually more robust.

guyharris:
Could you break the commit into at least three commits:

1. the changes to `gen_vlan_no_bpf_extensions()` (as those fix a problem unrelated to Geneve - they're necessary to handle VLANs on non-Ethernet networks, and are obviously correct);
2. the `off_macpl+offset` change (that one needs some more thinking - is it really an indication that there needs to be another variable)?;
3. the Geneve changes.

If there are any other non-Geneve-specific changes, split them into separate commits as well.

Put the CHANGES file change into the first of those.

jessegross:
Sure, I just updated the pull request broken apart into multiple commits. I left the off_macpl change the way it was since it seems a little more logical to me that way but added some comments to make it clearer and dropped the off_ll part since it was somewhat more difficult to justify.

guyharris:
I've picked up your first two changes manually, and I've redone a lot of gencode.c to introduce the notion of encapsulated link-layers, using it both for ATM LANE and PPPoE.  This might be usable, perhaps with some additional changes, for Geneve, when transporting Ethernet frames, as well.

The second two changes will probably not apply to the current state of gencode.c.

jessegross:
Awesome, thanks. It looks like your work isn't yet on master (unless I'm missing something) so I'll update the second two patches on top of this when it is available.

guyharris:
There's "master" and there's "master", i.e. there's "master on bpf.tcpdump.org" and there's "master on GitHub".  I do my work on "master on bpf.tcpdump.org"; changes on one of the masters get propagated to the other master by, I'm guessing, some timed script, so they get there after a delay.

jessegross:
Ah, thanks for pointing that out. I see your work now.

infrastation:
It is on GH now.

infrastation:
That's a lot of work.

jessegross:
Thanks again for your work. I've converted my code to run on top of your new infrastructure which definitely makes the patch less invasive and updated the pull request. I also fixed some bugs related to variable length headers that I noticed along the way.

PR #416, opened 2015-02-13
add very simple and basic implementation for can injection

kairenken:
I've added a very simple can injection feature to libpcap it is based on linux can tools. What do you think about it? Would it be ready to pull? What suggestions do you have for further improvements? Thank's


kairenken:
Try to pass Tracis check first

PR #347, opened 2014-04-04
Remove BlueZ dependence

MichalLabedzki:
Libpcap should not use BlueZ userspace headers, because it does not
use BlueZ at all. It used only Linux kernel, but it does not
provide userspace headers. For compatible reason we should use own copy
of headers and try to dynamically detect Linux kernel version and use
proper structures.

This fixes #343.

PR #417, opened 2015-02-15
Remove BlueZ dependence

### LINKED ISSUE #343: current libcap does not compile when bluez5 is present
The bluetooth code was written for bluez4. It checks for the presence of bluez and just assumes it being bleuz4. As the API changed with bluez5 the code no longer compiles when the bluez5 development package is installed.

MichalLabedzki:
Libpcap should not use BlueZ userspace headers, because it does not
use BlueZ at all. It use only Linux kernel, but it does not
provide userspace headers. For compatible reason we should use own copy
of headers and try to dynamically detect Linux kernel version and use
proper structures.

This fixes #343.

This is the second try of pull request: https://github.com/the-tcpdump-group/libpcap/pull/347

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -256,9 +256,6 @@
 /* target host supports Bluetooth sniffing */
 #undef PCAP_SUPPORT_BT
 
-/* target host supports Bluetooth Monitor */
-#undef PCAP_SUPPORT_BT_MONITOR
-
 /* target host supports CAN sniffing */
 #undef PCAP_SUPPORT_CAN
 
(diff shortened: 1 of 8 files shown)


guyharris:
Does copying the data structures etc. from the Linux kernel headers make the Bluetooth code GPLv2-licensed?  If so, that's a problem, as the rest of libpcap is BSD-licensed (and, unfortunately, licensed with the old 4-clause BSD license).

mcr:
guy, I'm trying to see where the copying of headers occured, and I didn't see it. Maybe I missed it.

guyharris:
I guess my concern is whether the "our own copy" could get people complaining that we just took GPLed code from Linux or not; if it was a reimplementation from scratch (I don't think "clean room" is necessary), that should suffice.

PR #425, opened 2015-03-27
Fix building Bluetooth Linux Monitor support with BlueZ 5.1+

jsitnicki:
Starting from version 5.1 BlueZ no longer exports the mgmt.h header
or any other header that declares the structure of packets passed
over HCI sockets set to use the HCI monitor channel.

Declare the structure locally and give it the same name as in the
Linux kernel 3.4+.

### DIFF
--- a/pcap-bt-monitor-linux.c
+++ b/pcap-bt-monitor-linux.c
@@ -34,12 +34,12 @@
 #endif
 
 #include <errno.h>
+#include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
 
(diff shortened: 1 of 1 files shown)


jsitnicki:
While the changes to remove libpcap's dependency on BlueZ are cooking (#347, #417), perhaps a fix for just the immediate cause of build against BlueZ 5 failure could remedy the situation that some distros (Fedora) are shipping libpcap with support for Bluetooth disabled altogether.

Fixes #343.

fxlb:
Merged. Thank you!

PR #429, opened 2015-04-20
bpf: increase snaplen if doing cooked mode userspace filtering

msekletar:
This commit should address the issue when bpf_filter_with_auxdata returned 0 for valid
packets because offset in filter exceeded tp_snaplen as returned by kernel. If
we filter in cooked mode filter offsets are adjusted because sll_header, we
should do the same for snaplen.


msekletar:
Just FYI, I closed and again opened PR in order to re-trigger Travis build. Previous one failed to complete on OS X for some reason.

PR #431, opened 2015-04-29
[RFC] bpf: if generating code for vlan keyword also check for non-offloaded case

msekletar:
Our bpf compiler should also handle cases if vlan rx-offload is turned off on
the device or it did not happen because 802.1q encapsulated frame contained
non-standard vlan tpid value. Under described circumstances BPF code which
checks if tag is present and has desired value fails, thus packet is dropped.

This commit alters behavior of compiler such that fallback case is generated. If
first check fails then we check for vlan tpid and vlan tci in the packet it
self. If they are present we then store displacement to memory cell 0, which
needs to be applied to all offsets generated from later vlan checks.

Also note that this patch makes libpcap _segfault in optimizer code_. Once I track down the problem I will send separate pull request.

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -259,6 +259,7 @@ static struct block *gen_ncmp(enum e_offrel, bpf_u_int32, bpf_u_int32,
 static struct slist *gen_load_absoffsetrel(bpf_abs_offset *, u_int, u_int);
 static struct slist *gen_load_a(enum e_offrel, u_int, u_int);
 static struct slist *gen_loadx_iphdrlen(void);
+static struct block *gen_load_vlan_indirect(const enum e_offrel, bpf_u_int32, bpf_u_int32);
 static struct block *gen_uncond(int);
 static inline struct block *gen_true(void);
 static inline struct block *gen_false(void);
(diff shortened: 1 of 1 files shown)


msekletar:
I've figured out why the optimizer is crashing. Following code will fix the crash (I will open separate pull request with the patch)

```C
diff --git a/optimize.c b/optimize.c
index ada2019..fff85aa 100644
--- a/optimize.c
+++ b/optimize.c
@@ -902,6 +902,10 @@ opt_peep(struct block *b)
 			v &= b->s.k;
 			break;
 
+                case BPF_JA:
+                        v = 1;
+                        break;
+
 		default:
 			abort();
 		}
```

But after I apply above patch and run tcpdump to produce optimized code, it will produce BPF code  which is not equivalent in behavior to non-optimized variant.

Here is not optimized BPF code
```
./tcpdump -O -d -i lo -nn 'vlan 1 && vlan 2'

(000) ld       #0x0
(001) st       M[0]
(002) ldb      [-4048]
(003) jeq      #0x1             jt 4    jf 6
(004) ldb      [-4052]
(005) jeq      #0x1             jt 18   jf 6
(006) ldb      [-4048]
(007) jeq      #0x0             jt 8    jf 35
(008) ldh      [12]
(009) jeq      #0x8100          jt 12   jf 10
(010) ldh      [12]
(011) jeq      #0x9100          jt 12   jf 35
(012) ld       #0x4
(013) st       M[0]
(014) ja       15
(015) ldh      [14]
(016) and      #0xfff
(017) jeq      #0x1             jt 18   jf 35
(018) ld       M[0]
(019) add      #12
(020) tax      
(021) ldh      [x + 0]
(022) jeq      #0x8100          jt 23   jf 26
(023) ldh      [x + 2]
(024) and      #0xfff
(025) jeq      #0x2             jt 34   jf 26
(026) ld       M[0]
(027) add      #12
(028) tax      
(029) ldh      [x + 0]
(030) jeq      #0x9100          jt 31   jf 35
(031) ldh      [x + 2]
(032) and      #0xfff
(033) jeq      #0x2             jt 34   jf 35
(034) ret      #262144
(035) ret      #0
```

and here is optimized version

```
(000) ldb      [-4048]
(001) jeq      #0x1             jt 2    jf 4
(002) ldb      [-4052]
(003) jeq      #0x1             jt 13   jf 22
(004) jeq      #0x0             jt 5    jf 22
(005) ldh      [12]
(006) jeq      #0x8100          jt 8    jf 7
(007) jeq      #0x9100          jt 8    jf 22
(008) ld       #0x4
(009) ja       10
(010) ldh      [14]
(011) and      #0xfff
(012) jeq      #0x1             jt 13   jf 22
(013) add      #12
(014) tax      
(015) ldh      [x + 0]
(016) jeq      #0x8100          jt 18   jf 17
(017) jeq      #0x9100          jt 18   jf 22
(018) ldh      [x + 2]
(019) and      #0xfff
(020) jeq      #0x2             jt 21   jf 22
(021) ret      #262144
(022) ret      #0
```

For some reason optimizer thinks that trick I did with storing offset in scratch pad memory is redundant/deadcode and it optimizes that away.

/me goes to read Steve's paper from SIGCOMM to get better understanding how the damn thing works.

luky37:
Since this is [RFC] and - if I understand correctly - there are problems with the produced BPF code, maybe this should be discussed in the tcpdump-workers mailing-list?

msekletar:
> Since this is [RFC] and - if I understand correctly - there are problems with the produced BPF code, maybe this should be discussed in the tcpdump-workers mailing-list?

Any particular reason why we shouldn't continue discussion here? I think it only makes sense to keep discussion about patches and code on Github since code changes should be preferably posted using Pull requests.

mcr:
> Since this is [RFC] and - if I understand correctly - there are
    > problems with the produced BPF code, maybe this should be
    > discussed in the tcpdump-workers mailing-list?

Michal Sekletar <notifications@github.com> wrote:
    > Any particular reason why we shouldn't continue discussion here? I
    > think it only makes sense to keep discussion about patches and code
    > on Github since code changes should be preferably posted using Pull
    > requests.

Because there is a much wider audience on the list.

luky37:
I just suggested it because I feel like the discussion stagnated here and since we have a blocking technical issue, I thought the mailing list would provide a broader audience and someone may be able to suggest a solution to this problem. Instead it seems you are the only one thinking about this problem (I'm very interested in this fix - but unable to help with the code).

I'm not saying the move the discussion about actual commit to the mailing list. I'm suggesting to ask the mailing list for help regarding the specific optimizer problem.

In OSS projects specifically, the discussion about code is very important and saves (you in this case) a lot of time. But it doesn't seem there is a lot of discussion between developers going on on github for libpcap specifically.

msekletar:
Ok then, I will send a post on the mailing list.

mkubecek:
I somehow missed this pull request when I was checking if there is already one so I ended up doing the same.

The vlan filter improvement is now merged as commit d739b068ac29. As for the optimizer issue, I believe it's the same as issue #581 (proposed fix included in pull request #582).

msekletar:
> The vlan filter improvement is now merged as commit d739b06. As for the optimizer issue, I believe it's the same as issue #581 (proposed fix included in pull request #582).

Very nice! Thank you for working on this.

PR #432, opened 2015-05-04
Fix memory leak in pcap_create_interface

bhickey:
Free opt.source in pcap_create_common when freeing handle.

Signed-off-by: Brendan Hickey <bhickey@google.com>


guyharris:
Fixed a bit differently in in cd46a74cf49299d92640bfe63b4caf66e0cce65c; that fixes more modules than just the Linux PF_PACKET socket module.  It also reduces the amount of code in the individual `pcap_create_handle()` routines.

bhickey:
Thanks :)

PR #433, opened 2015-05-07
Give credit to Ani Sinha

msekletar:
Ani was mentioned as the author of initial implementation which resulted in PR #391. However PR description is not visible in git repository. Let's fix that and give him credit in the comment instead.


infrastation:
Done.

PR #435, opened 2015-05-15
Allow pcap_set_buffer_size() to set SNF_DATARING_SIZE in pcap-snf.c.

myri:


guyharris:
So the ring size argument to snf_open() is in bytes?

myri:
The dataring_sz argument to snf_open() is considered to be in bytes if it is greater than or equal to 1048576.  If it is less than that, it is considered in megabytes. If the value is zero or less than 0, the Sniffer library tries to choose a sensible default unless the SNF_DATARING_SIZE variable is set in the environment.

guyharris:
> The dataring_sz argument to snf_open() is considered to be in bytes if it is greater than or equal to 1048576. If it is less than that, it is considered in megabytes.

Then perhaps if the value provided by `pcap_set_buffer_size()` is < 1048576, you should multiply it by 1048576 before passing it to `snf_open()`, to ensure that the value, in bytes (that's what the `pcap_set_buffer_size()` says are the units of the argument), is used as the ring size.

guyharris:
Or just pass 1048576 if the value is < 1048576, as you'd have to round up the value after dividing it by 1048576.

myri:
Thanks. I'll treat the value as in bytes and convert it to the minimum of 1048576 if it < 1MB.

guyharris:
So the only part of this change that survived your most recent commit is the "Update snf_findalldevs() ..." part; did you really intend not to "Allow pcap_set_buffer_size() to set SNF_DATARING_SIZE in pcap-snf.c"?

myri:
Right.  I may add pcap_set_buffer_size() support back in later after some more review and discussion. Thanks for noticing.

PR #440, opened 2015-06-02
Fix uclinux

baruchsiach:
These two patches fix two different problems causing together a build failure on noMMU Linux (uClinux) systems like Blackfin.

### DIFF
--- a/configure.in
+++ b/configure.in
@@ -149,7 +149,7 @@ struct rtentry;
 fi
 
 case "$host_os" in
-linux*)
+linux*|uclinux*)
 	AC_CHECK_HEADERS(linux/sockios.h linux/if_bonding.h,,,
 	[
 #include <sys/socket.h>
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -955,7 +955,8 @@ added:
 static int
 is_bonding_device(int fd, const char *device)
 {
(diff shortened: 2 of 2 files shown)


guyharris:
So why, if there's no if_bonding.h, is one or both of `BOND_INFO_QUERY_OLD` or `SIOCBONDINFOQUERY` being defined?  Are either of those ioctls supported?  If so, where is the `ifbond` structure defined?

baruchsiach:
@guyharris `SIOCBONDINFOQUERY` is defined in `linux/sockios.h, which is included unconditionally (and again when `HAVE_LINUX_SOCKIOS_H` is defined).

guyharris:
Well, I guess putting a huge pile of ioctls in `sockios.h` was better than just putting them into `ioctl.h`, as it at least keeps ioctls done on sockets separate from ioctls done on arbitrary file descriptors or on file descriptors for serial lines/pseudo-terminals or....

Still, it would have been better to split them up further.

So I guess a system could have the bonding ioctls in `linux/sockios.h` even if it doesn't support them.

guyharris:
`BOND_INFO_QUERY_OLD` appears to be defined in `linux/if_bonding.h`, so if there is no `linux/if_bonding.h` it won't get defined.

Issue #438, opened 2015-05-21
packets are not truncated by using pcap_open_dead()


ikuzar:
I write a program which takes as input several pcap files. These pcap files have the same origin (they are captured from the same interface). I have to filter and merge them before dumping towards an output file. Packets must be truncated to 64 bytes. But it is not the case: packets keep their initial length. I describe below the steps. I loop on these steps with the number of input files indexed by int i:
0) I open input file with  pcap_open_offline:
inputs_pcap[i] = pcap_open_offline(inputs_file[i], pcap_errbuf);
1) I retrieve once the link-layer type from the first input file, by calling pcap_datalink:
input_link_layer = pcap_datalink(inputs_pcap[0]);
2) I create once the output pcap_t  (for dumping purpose) with pcap_open_dead:
output_pcap = pcap_open_dead(input_link_layer, 64);
3) Then, I create once a dumper with output_pcap:
dumper = pcap_dump_open(output_pcap, out_file_name);
4) Then, I retrieve packets from current input pcap with pcap_loop:
res = pcap_loop(inputs_pcap[i], -1, dump_packet, (u_char *) dumping_info);

typedef struct dumping_info {
	pcap_dumper_t *dumper; // dumper returned by 3)
	pcap_t *current_input_pcap; // inputs_pcap[i]
} * dumping_info_t;

static void
dump_packet(u_char *user_data, const struct pcap_pkthdr *pkt_hdr, const u_char *pkt_data)
{
	dumping_info_t dumping_info = (dumping_info_t) user_data;
        ...
	pcap_dump((u_char *)(dumping_info->dumper), pkt_hdr, pkt_data);	
}

5) At the end of the loop (on variable "i"), I flush the dumper
pcap_dump_flush(dumper)

Is it a bug of pcap_open_dead( ) ? this routine seems to ignore snaplen setting... 

My configurations:
- Ubuntu 10.04 LTS, Linux 2.6.32-74-generic
- ldd result on the executable: libpcap.so.0.8 => /usr/lib/libpcap.so.0.8 (0x00007fb0de6ed000)
- dpkg -l | grep libpcap result: 
ii  libpcap0.8                           1.0.0-6                                         system interface for user-level packet capture
ii  libpcap0.8-dev                       1.0.0-6                                         development library and header files for libpcap0.8

Thanks for your help

Astolfi:
[...]

PR #444, opened 2015-06-24
Truncate packet length for offline captures based on BPF return value

Astolfi:
Seems related to this ticket:

https://github.com/the-tcpdump-group/libpcap/issues/438

### DIFF
--- a/savefile.c
+++ b/savefile.c
@@ -376,6 +376,7 @@ pcap_offline_read(pcap_t *p, int cnt, pcap_handler callback, u_char *user)
 	int status = 0;
 	int n = 0;
 	u_char *data;
+	u_int filter_out = -1;
 
 	while (status == 0) {
 		struct pcap_pkthdr h;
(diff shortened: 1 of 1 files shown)


infrastation:
Fixes #438. Needs a trivial rebase because since commit 1131a7c the check uses `pcap_filter()` instead of `bpf_filter()`.

savefile.c
```diff
@@ -405,7 +406,11 @@ pcap_offline_read(pcap_t *p, int cnt, pcap_handler callback, u_char *user)
 		}
 
 		if ((fcode = p->fcode.bf_insns) == NULL ||
-		    bpf_filter(fcode, data, h.len, h.caplen)) {
+		    (filter_out = bpf_filter(fcode, data, h.len, h.caplen))) {
+		   /* Google-specific fix for:
```
fxlb, line 410: Google-specific fix ???

infrastation: In the context of the bug this probably means "nobody bothered to use the exact return value of this function until it broke some Google code, which I had to fix".

I wonder if it would be right to assign `h.caplen = -1` when `p->fcode.bf_insns == NULL`, because that's one of the effects of this change.

fxlb: Perhaps...
Without context, this comment is unnecessary and should be deleted.

guyharris:
Any particular reason not to handle #438 by doing the truncation in the dump routine, i.e. just adding

    /* Truncate packet to a maximum of 64 bytes */
    if (pkt_hdr->caplen > 64)
        pkt_hdr->capen = 64;

to `dump_packet()` before the call to `pcap_dump()`.  This requires *no* changes to libpcap.

PR #453, opened 2015-08-31
WIN32 and _WIN32 macros

ujos:
Hi

There is issue with libpcap on Windows. When you create new application and include pcap.h file, you have compilation error like:

fatal error C1083: Cannot open include file: 'sys/time.h': No such file or directory

The reason is "#if defined(WIN32)". This macro is not standard macro and compiler does not create it by default. However there is another one "_WIN32" which is defined by Visual C++ compiler by default. I believe we should use it instead of WIN32. 

In my fork I've fixed this also fixed some ADAPTER/Adapter type issues. Code successfully compiles with WinPcap 4.1.3 (with some changes inside WinPcap)

Thanks
Dima


ujos:
Also there is issue with strdup + Visual Studio 2015 + __STDC__ macro. On VS2015 is __STDC__ is defined, there is no function "strdup". I did not commit fix yet.

guyharris:
I've converted all checks for WIN32 in libpcap and tcpdump to check for _WIN32 instead, as that's Microsoft's official #define for "32-bit and wider Windows".

guyharris:
I've fixed all references to "Adapter" to refer instead to "ADAPTER".

guyharris:
I've cleaned up the pcap_stats() code (it needs more work, but this should be sufficient to fix compilation errors for the stats).

guyharris:
> Also there is issue with strdup + Visual Studio 2015 + ____STDC____ macro. On VS2015 if ____STDC____ is defined, there is no function "strdup". I did not commit fix yet.

Do you mean `____STDC____` or `__STDC__`?

ujos:
> I've fixed all references to "Adapter" to refer instead to "ADAPTER".

I guess I've submitted Pull request too early. I've realized that `struct _ADAPTER` and `LPADAPTER` should be used instead. Please check this commit: https://github.com/ujos/libpcap/commit/faec94d32bae31a2d7e266c33e5f7a9caf123c6c

> I've cleaned up the pcap_stats() code (it needs more work, but this should be sufficient to fix compilation errors for the stats).

Sorry... I've forgot to remove it from pcap_t struct. But then I've realized that I can add it to `pcap_win` struct in the pcap-win32.c. Please check https://github.com/ujos/libpcap/commit/1a6059186c74f0ca2a91fd742cf40abbd7a46270

> Do you mean ____STDC____ or __STDC__?

Yes, I did no know how to write two underscores :)

Should I post new Pull request for those two commits?

guyharris:
Please rebase your changes against the current GitHub top-of-branch.

ujos:
Ok... It will take time. I'm new in Git. Need to read how to do it first.

ujos:
Looks like rebasing is done. There are issues with spaces-tabs. I do not know how to handle it right now. If I run `git diff eec6bc9 d41141e -w >patch` I see following differences between your and mine repos:
1. You use `ADAPTER` while in my code I use `_ADAPTER` and `LPADAPTER`. `ADAPTER` is `typedef` in the WinPcap.

2. I've removed `#define snprintf  _snprintf`. VS2015 complains on this. 

3. You have `_WIN32_DEFAULT_KERNEL_BUFFER_SIZE` while correct name I guess is `WIN32_DEFAULT_KERNEL_BUFFER_SIZE`

4. `struct bpf_hdr {` can be removed

5. Added includes (io.h and stdio.h) and macros (isatty and fileno) to scanner.l to solve compilation warnings

gvanem:
> I've removed #define snprintf _snprintf. VS2015 complains on this.

I too use VS2015 and the Windows-Kit, but libpcap should support older MSVC compilers & SDK which doesn't have `snprintf()`. So the define should be in there since VS2015 do have `_snprintf()` too. You'll just have to ensure `<stdio.h>` gets included before that `#define` statement.

guyharris:
> 1. You use `ADAPTER` while in my code I use `_ADAPTER` and `LPADAPTER`. `ADAPTER` is `typedef` in the WinPcap, AFAIK you cannot use typedef in forward declaration expression.

The problem isn't the typedef (note that `LPADAPTER` is also a typedef), it's the requirement that `<Packet32.h>` be included in order to declare `pcap_get_adapter()`.  That routine is needed in order to keep `pcap_t` opaque; I've changed pcap/pcap.h in b6977c56508c684ec70b2518b68f33cfe711f43f to declare it only if `<Packet32.h>` has previously been included.

Giving an incomplete declaration of `struct _ADAPTER` in `pcap/pcap.h` won't help people who write code assigning the result of `pcap_get_adapter()` to a variable of type `ADAPTER *` or `LPADAPTER`.

> 2. I've removed `#define snprintf  _snprintf`. VS2015 complains on this.

See Gisle's comment.

> 3. You have `_WIN32_DEFAULT_KERNEL_BUFFER_SIZE` while correct name I guess is `WIN32_DEFAULT_KERNEL_BUFFER_SIZE`

Fixed in 6f35acdba117098cf5e3dfc40602ea4358b36506.  (I thought I'd fixed both of the `WIN32_DEFAULT_..._BUFFER_SIZE` #defines, but I'd only fixed `WIN32_DEFAULT_USER_BUFFER_SIZE`.)

> 4. `struct bpf_hdr {` can be removed in pcap-win32.c

Fixed in aebbd1c17111e590c1be72c148e00501bb204bee.  (WinPcap should really declare it in one and only one header, so the same structure is shared by the drivers, Packet.dll, and WinPcap itself.)

> 5. Added includes (`io.h` and `stdio.h`) and macros (`isatty` and `fileno`) to scanner.l to solve compilation warnings

What version of Flex are you using?  I'd expect the Windows version of Flex either not to use UN\*X-only APIs or to make them "available" in Windows by using the right #defines.)

ujos:
> You'll just have to ensure <stdio.h> gets included before that #define statemen

OK. I will add #if _MSC_VER >= 1900

> The problem isn't the typedef (note that LPADAPTER is also a typedef)

As for me I do not like when order of includes is important. IMHO it is more user friendly if you use forward declaration. If user need this function, he will include Packet32.h where it is required by compiler. I used LPADAPTER in the implementation part of the libpcap (not interface files) where I can control which header files are included.

ujos:
BTW, I'm thinking on creation CMake file for libpcap. It will simplify support of different MS Visual Studio's.

ujos:
> What version of Flex are you using? I'd expect the Windows version of Flex either not to use UN*X-only APIs or to make them "available" in Windows by using the right #defines.)

flex 2.5.39 (cygwin)

gvanem:
@guyharris   it's the requirement that <Packet32.h> be included in order to declare pcap_get_adapter()

Then why not include `<Packet32.h>` unconditionally for `_WIN32`? That would solve some confusion. And there are still lots of that in libpcap + WinPcap.

@ujos Added includes (io.h and stdio.h) and macros (isatty and fileno) to scanner.l to solve compilation warnings

Have you added `-DYY_NO_UNISTD_H` to your `CFLAGS`? The only warning I get when compiling `scanner.c` is:
```
<stdout>(4403): warning C4273: 'isatty': inconsistent dll linkage
f:\ProgramFiler-x86\WindowsKits\10\Include\10.0.10150.0\ucrt\corecrt_io.h(489): note: see previous definition of 'isatty'
scanner.l(391): warning C4090: 'function': different 'const' qualifiers
```

Because of:
```
#ifndef __cplusplus
extern int isatty (int );
#endif /* __cplusplus */
```

I think that warning is harmless.

ujos:
> Have you added -DYY_NO_UNISTD_H to your CFLAGS?

Yes, I have this macro and without macros isatty and fileno I have following warnings:
```
1><stdout>(4433): warning C4013: 'isatty' undefined; assuming extern returning int
1><stdout>(4433): warning C4013: 'fileno' undefined; assuming extern returning int
```

ujos:
> Then why not include <Packet32.h> unconditionally for _WIN32? That would solve some confusion. And there are still lots of that in libpcap + WinPcap.

Works for me :). However I prefer less includes where it is possible.

guyharris:
> The problem isn't the typedef (note that `LPADAPTER` is also a typedef), it's the requirement that `<Packet32.h>` be included in order to declare `pcap_get_adapter()`.  That routine is needed in order to keep `pcap_t` opaque;

Another alternative is to provide routines that take a `pcap_t` as the first argument and do "get" and "set" operations on NDIS OIDs; I've checked in changes to do that, and get rid of pcap_get_adapter()`.

ujos:
Added CMake support dd98f6fa846b5fc66167341e377c30cb35283a3f

guyharris:
Please remove all the white space changes - if a file uses tabs for indentation, don't change them to spaces.  That will make it easier to look at the diffs and see what's changed (if anything).

ujos:
Unfortunately, my editor supports this mode (use tabs if tabs are present). I've configured it according to policy on my project in company.

Please try following command `git diff -w  e19fa9b 17a2a3e`.  It will show difference between previous and last commit excluding whitespace changing.

For future I will create AStyle scheme and will reformat the code. So next commit will not have this issue... But it will be not easy to merge with your master branch.

fxlb:
Please revert this change, for the space changes, the spaces at end of lines and newlines at end of lines ...

fxlb:
Some of your previous changes have also newlines at end of lines.

For changes like `s/snprintf/pcap_snprintf/`, a shell line with a for loop and a sed command is very effective.

guyharris:
> Please revert this change,

Well, it's in his own repository, not the tcpdump.org repository, but...

> for the space changes, the spaces at end of lines and newlines at end of lines

...we don't want the space changes, added spaces at the ends of lines, or CRs added to the ends of lines  either.

(I presume you meant "CRs", not "newlines" - whether it's Windows or UN*X, there's a line feed/newline at the end of the line; the difference between Multics-inspired systems such as UN*X and DEC-operating-system-inspired systems such as Windows is that the former end lines with just line feed/newline and the latter end them with carriage return+line feed.)

For the CR-LF vs. LF, that's probably a Git configuration issue somewhere; I think Git can arrange that checked-out files have CR-LF at the ends of lines on Windows and just LF on UN*X, and that a file checked in with CR-LF line endings on Windows or LF-only line endings on UN*X will both work and not remove CRs on Windows or add them on UN*X.

fxlb:
> For the CR-LF vs. LF, that's probably a Git configuration issue somewhere; I think Git can arrange that checked-out files have CR-LF at the ends of lines on Windows and just LF on UN*X, and that a file checked in with CR-LF line endings on Windows or LF-only line endings on UN*X will both work and not remove CRs on Windows or add them on UN*X.

You are right, the git configuration is fixed by e1f993b39b60bf01f69d9a8ab05d9a674a899d8a.

infrastation:
So what exactly is the focus of this pull request?

ujos:
Initial purpose of this pull request was _WIN32 macro usage. 

On CR-LF and spaces/tabs. Guys, sorry, but I cannot change configuration of my IDE and text editors as it violate policy of my primary project I'm working on.

What I'd like suggest for this case is to require contributes to use AStyle (or another tools like AStyle) before commit. In this way you will have code formatting standardized and there will be no issues with spaces.

guyharris:
OK, I've checked in 735f1f9d3318693f0096be4198d34e9ac0985777 and 6d35f01601154022c4568addf44af1338bc9da4c, which changes uses of `snprintf()` and `vsnprintf()` to use `pcap_snprintf()` and `pcap_vsnprintf()`, #defines `pcap_snprintf` as `snprintf` on UN\*Xes that have `snprintf()`, defines our own `pcap_snprintf()` and `pcap_vsnprintf()` on UN\*Xes that don't have `snprintf()`, and defines `pcap_snprintf()` and `pcap_vsnprintf()` wrappers around `_snprintf()` and `_vsnprintf()` that ensure null-termination on Windows.

ujos:
Looks good

fxlb:
Any reason to keep this issue open ?

ujos:
No. Thanks.

PR #458, opened 2015-10-09
Add the pcap_file_type() function and the file_type member in pcap_t

fxlb:
[FOR COMMENTS ONLY; no man pages updates currently]

This function is used to get the pcap file type when reading a pcap file.
The types are currently pcap and pcap-ng (PCAP_FILE_TYPE and
PCAP_NG_FILE_TYPE).

The first use of this function is to allow tcpdump to know the file type
and to do relevant version checks based on these types.

For that purpose, we need to move PCAP_NG_VERSION_MAJOR from sf-pcap-ng.c
to pcap/pcap.h.

### DIFF
--- a/pcap-int.h
+++ b/pcap-int.h
@@ -195,6 +195,8 @@ struct pcap {
 	int version_major;
 	int version_minor;
 
+	int file_type;		/* pcap or pcap-ng (for now) */
+
 	int snapshot;
 	int linktype;		/* Network linktype */
 	int linktype_ext;       /* Extended information stored in the linktype field of a file */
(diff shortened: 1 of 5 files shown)


fxlb:
I am thinking to add: `PCAP_NG_VERSION_MINOR 0`

guyharris:
> The first use of this function is to allow tcpdump to know the file type and to do relevant version checks based on these types.

Why does tcpdump - or any other application using libpcap - need to do version checks?  Are there checks that can't be done in libpcap?

fxlb:
>Why does tcpdump - or any other application using libpcap - need to do version checks? Are there checks that can't be done in libpcap?

I wil try the libpcap way ...

fxlb:
Libcap way to come. Closing.

PR #466, opened 2015-12-22
Avoiding conditional directives that break statements.

RomeroMalaquias:


infrastation:
This is going to introduce a declaration in the middle of a code block, which will make some compilers unhappy. Could you find a better way to do the same?

guyharris:
> Could you find a better way to do the same?

Such as e627474b63517588367ea9c7faec822b2059f36b?

PR #467, opened 2015-12-22
Add module definition file support for building on Windows

blazeable:
Without a module definition file (and without __declspec annotations in the source code), no symbols are exported in the compiled DLL. This small patch adds a module definition file for pcap.dll, and updates CMakeLists to pass the /DEF flag to the linker when building with MSVC.

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -665,11 +665,6 @@ pcap_compile(pcap_t *p, struct bpf_program *program,
 	u_int len;
 	int  rc;
 
-	/*
-	 * XXX - single-thread this code path with pthread calls on
-	 * UN*X, if the platform supports pthreads?  If that requires
-	 * a separate -lpthread, we might not want to do that.
-	 */
 #ifdef _WIN32
 	static int done = 0;
 


guyharris:
> (and without __declspec annotations in the source code)

Do the _declspec annotations I just added to the trunk in fa79dd4c6f651fb853efe95681c9843b16a40df8 suffice here?

blazeable:
Should be enough, thanks. However there are a couple compile errors introduced by fa79dd4:
* The prototypes for ```bpf_validate()```, ```bpf_filter()``` and ```bpf_filter_aux_data()``` in pcap/bpf.h don't match those in pcap/pcap.h, causing error C2375 'redefinition; different linkage'. The ```PCAP_API``` macro needs to be used in pcap/bpf.h. However, the macro isn't defined in bpf_filter.c, since it doesn't include pcap/pcap.h (only pcap-stdinc.h on Windows).
* The prototype for ```pcap_vsnprintf()``` in pcap/pcap-int.h has two storage class specifiers, causing error C2159.

blazeable:
WinDump uses ```pcap_next_etherent``` from etherent.c, so it should be exported as well.

guyharris:
> The prototype for `pcap_vsnprintf()` in pcap/pcap-int.h has two storage class specifiers

Fixed in 26bb2e5d087d306088f8c36987d4c34d98d9ed0b.

guyharris:
> WinDump uses `pcap_next_etherent` from `etherent.c`, so it should be exported as well.

The comment in top-of-trunk tcpdump says

     * XXX pcap_next_etherent doesn't belong in the pcap interface, but
     * since the pcap module already does name-to-address translation,
     * it's already does most of the work for the ethernet address-to-name
     * translation, so we just pcap_next_etherent as a convenience.

so we either need to 1) provide another official way of doing that or 2) say that it *does* belong in the pcap interface and publishing it, including documenting it and `pcap/namedb.h`.

gvanem:
> WinDump uses pcap_next_etherent from etherent.c, so it should be exported as well.

Do you actual get ether-names in a WinDump capture? (use `windump -e ip`to check). I'm not sure you can since `PCAP_ETHERS_FILE "/etc/ethers"` (in `pcap/namedb.h`) most likely does *not* exist on Windows. But in WinDump, the file `win32/src/ether_ntohost.c` should work. 

So I'm confused as to why it needs to be exported as the situation is now; the ether-stuff in `nametoaddr.c` seems to be used for generating a pcap-filter only. IMHO there should be no need for *two* ether-lookup APIs.

guyharris:
The comment for `init_etherarray()` in tcpdump says:

> Initialize the ethers hash table.  We take two different approaches depending on whether or not the system provides the ethers name service.  If it does, we just wire in a few names at startup, and `etheraddr_string()` fills in the table on demand.  If it doesn't,then we suck in the entire `/etc/ethers` file at startup.  The idea is that parsing the local file will be fast, but spinning through all the ethers entries via NIS & next_etherent might be very slow.

> XXX `pcap_next_etherent` doesn't belong in the pcap interface, but since the pcap module already does name-to-address translation, it's already does most of the work for the ethernet address-to-name translation, so we just `pcap_next_etherent` as a convenience.

It appears to be assuming that if the system has `ether_ntohost()` it uses YP/NIS; that's a bit of a bogus assumption.

In any case, the "suck in the entire `/etc/ethers` file" code uses `pcap_next_etherent()` from libpcap/WinPcap to avoid duplicating it.

guyharris:
Well, I tried playing with this change, but it's a bit uglier than I expected when I tried using the symbol list for other platforms, so I reverted to the `PCAP_API` scheme.

guyharris:
> WinDump uses `pcap_next_etherent` from `etherent.c`, so it should be exported as well.

Fixed in af0dd17e5c43793b5b8770ab341209e10b46f951.

guyharris:
> The prototypes for `bpf_validate()`, `bpf_filter()` and `bpf_filter_aux_data()` in `pcap/bpf.h` don't match those in `pcap/pcap.h`, causing error C2375 'redefinition; different linkage'. The PCAP_API macro needs to be used in `pcap/bpf.h`. However, the macro isn't defined in `bpf_filter.c`, since it doesn't include `pcap/pcap.h` (only pcap-stdinc.h on Windows).

Fixed in af0dd17e5c43793b5b8770ab341209e10b46f951.  I moved the definition of `PCAP_API` to a separate header file, included by `pcap/pcap.h`, `pcap/bpf.h`, and `pcap/namedb.h`, and used it for both declarations of `bpf_validate()` and `bpf_filter()`.  `bpf_filter_with_aux_data()` is an internal routine, only declared in `pcap/bpf.h`, so I didn't mark it; I should put it in some private header file so that it's not visible outside the libpcap source.

PR #470, opened 2015-12-26
[Win32] savefile.c

gvanem:
As comment on elsewhere (I lost that reference), the functions:
`pcap_fopen_offline_with_tstamp_precision()` and `pcap_fopen_offline()` must only be coded when `LIBPCAP_EXPORTS` is defined. They are macros in `pcap/pcap.h`otherwise.

### DIFF
--- a/savefile.c
+++ b/savefile.c
@@ -75,8 +75,10 @@
  * a HANDLE (which is OS-defined, not CRT-defined, and is part of the Win32
  * and Win64 ABIs).
  */
-static pcap_t *pcap_fopen_offline_with_tstamp_precision(FILE *, u_int, char *);
-static pcap_t *pcap_fopen_offline(FILE *, char *);
+#if defined(LIBPCAP_EXPORTS)
+  static pcap_t *pcap_fopen_offline_with_tstamp_precision(FILE *, u_int, char *);
+  static pcap_t *pcap_fopen_offline(FILE *, char *);
+#endif
 #endif
 
 /*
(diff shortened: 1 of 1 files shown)


guyharris:
> must only be coded when `LIBPCAP_EXPORTS` is defined.

When libpcap is being compiled, `LIBPCAP_EXPORTS` should be defined; if it's ever *not* defined when compiling `savefile.c` (or any other source file in libpcap), that's a bug in the build process.

gvanem:
> When libpcap is being compiled, LIBPCAP_EXPORTS should be defined

That's completely non-intuitive. It should IMHO be `LIBPCAP_BUILDING` instead.

> that's a bug in the build process.

Is it a bug when I want to create a *static library*. Hence I removed `LIBPCAP_EXPORTS` from my Makefile. But (as stated elsewhere), I feel it would be more intuitive to add a `LIBPCAP_STATIC`. Then there would be no doubt. Now we're forced to use libpcap as a .DLL.

guyharris:
> > When libpcap is being compiled, `LIBPCAP_EXPORTS` should be defined

> That's completely non-intuitive. It should IMHO be `LIBPCAP_BUILDING` instead.

More to the point, it makes sense to have #ifdefs such as that in *public header* files, which are read both when building libpcap and when building code using libpcap, but not in the source to libpcap itself, which is read only when building libpcap.

guyharris:
OK, it's now called `BUILDING_LIBPCAP`, and is defined iff we're building libpcap.

gvanem:
> OK, it's now called BUILDING_LIBPCAP.

Or `BUILDING_PCAP`. I see. I'll delete this pull request then.

PR #471, opened 2015-12-27
Add module definition file for building on Windows

guyharris:
Update CMake build script to add linker arguments to use the definition
file.


guyharris:
Using `PCAP_API` is getting a bit complex, as not all files that need to be in the API are declared in pcap/pcap.h.  I think I can turn the .def file into instructions to various UN\*X linkers to export *only* symbols in that file, and that'll even work on AIX and HP-UX, where the compilers don't support saying "don't export this from the shared library" or "export this from the shared library" at compile time.

guyharris:
Reverted - using `PCAP_API` turned out to be less complex than turning the .def file into instructions to the linkers on various UN\*Xes.

rathnaTech:
how to run the libpcap through command line to get dumped packets.

PR #474, opened 2016-01-12
media word type from ifmediareq struct differs on OpenBSD

devnexen:


guyharris:
> media word type from ifmediareq struct differs on OpenBSD

Except on versions where it doesn't differ; older versions of OpenBSD used int.

devnexen:
true. did a small update then.

devnexen:
does it look good ? Just I maintain nmap for OpenBSD which depends on libpcap and would be grateful if this change ended up in the repository. Thanks.

guyharris:
Fixed in f2b77b74364547a5f319bf780f13932e0937f82b.

PR #491, opened 2016-04-17
[API] add functions to be able to get packet's savefile offset

alexandretea:
I noticed there was no way through the API to get a packet's savefile offset, so I added those functions in order to resolve this problem :
`pcap_next_with_sf_offset()` and `pcap_next_ex_with_sf_offset()`

They work the same as `pcap_next()` and `pcap_next_ex()` but takes an additional pointer on a long, which will be set to the packet's savefile offset. If the packets are read from a live capture, then this long will be set to -1.

This feature would be useful for a project I work on, which aim is to retrieve packets from pcap[-ng] files and then process them. I need to keep their savefile offsets to be able to retrieve their raw format at any time, after the processing.
It may also be useful for others :)

### DIFF
--- a/pcap-int.h
+++ b/pcap-int.h
@@ -200,6 +200,9 @@ struct pcap {
 	int activated;		/* true if the capture is really started */
 	int oldstyle;		/* if we're opening with pcap_open_live() */
 
+    long lastpkt_offset;    /* last packet offset in save file */
+    long current_offset;    /* current offset in save file */
+
 	struct pcap_opt opt;
 
 	/*
--- a/pcap.c
+++ b/pcap.c
@@ -303,6 +303,26 @@ pcap_next_ex(pcap_t *p, struct pcap_pkthdr **pkt_header,
 	return (p->read_op(p, 1, p->oneshot_callback, (u_char *)&s));
 }
 
+const u_char *
+pcap_next_with_file_offset(pcap_t *p, struct pcap_pkthdr *pkt_header,
+        long *file_offset)
+{
+    const u_char *ret = pcap_next(p, pkt_header);
+
+    *file_offset = p->lastpkt_offset;
+    return ret;
+}
+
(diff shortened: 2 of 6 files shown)


pcap-int.h
```diff
@@ -200,6 +200,9 @@ struct pcap {
 	int activated;		/* true if the capture is really started */
 	int oldstyle;		/* if we're opening with pcap_open_live() */
 
+    long lastpkt_offset;    /* last packet offset in save file */
```
guyharris: Please use tabs, not 4 spaces, for indentation.

pcap.c
```diff
@@ -303,6 +303,26 @@ pcap_next_ex(pcap_t *p, struct pcap_pkthdr **pkt_header,
 	return (p->read_op(p, 1, p->oneshot_callback, (u_char *)&s));
 }
 
+const u_char *
+pcap_next_with_file_offset(pcap_t *p, struct pcap_pkthdr *pkt_header,
```
guyharris: This is a version of `pcap_next()` with additional information, and inherit's `pcap_next()`s broken-by-design problems.  Don't bother with it.

sf-pcap-ng.c
```diff
@@ -875,6 +875,8 @@ pcap_ng_check_header(bpf_u_int32 magic, FILE *fp, u_int precision, char *errbuf,
 	bhdrp->block_type = magic;
 	bhdrp->total_length = total_length;
 	shbp->byte_order_magic = byte_order_magic;
+    p->lastpkt_offset = bhdrp->total_length;
```
guyharris: Again, use tabs for indentation.

alexandretea:
Sorry about the indentation, I fixed it :) I also removed `pcap_next_with_sf_offset()`.

mcr:
Please rebase.

PR #494, opened 2016-05-31
Add some error checking to pcap_dump().

shane-kerr:
We noticed that pcap files were corrupted after a full disk event. This patch tries to help a bit.

Note that probably it would be better if `pcap_dump()` returned an error code, rather than being a void function. The best way to transition to that might be to add a new function that returned an error, and had `pcap_dump()` just call that and throw away the error code, then declare `pcap_dump()` deprecated in a future version of the ABI and ultimately removing it.

Who thought a function that doesn't return an error code was a good idea? :wink:

### DIFF
--- a/sf-pcap.c
+++ b/sf-pcap.c
@@ -648,13 +648,39 @@ pcap_dump(u_char *user, const struct pcap_pkthdr *h, const u_char *sp)
 	struct pcap_sf_pkthdr sf_hdr;
 
 	f = (FILE *)user;
+	/*
+	 * If the output file handle is in an error state, don't write
+	 * anything.
+	 *
+	 * While in principle a file handle can return from an error state
+	 * to a normal state (for example if a disk that is full has space
+	 * freed), we have possibly left a broken file already, and won't
+	 * be able to clean it up. The safest option is to do nothing.
+	 *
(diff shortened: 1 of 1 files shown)


infrastation:
This rebases on the current master branch fine after a trivial conflict resolution, but the pull request predates the "push into the contributor's branch" feature, so I cannot update it in place. In any case, the suggested change looks useful.

guyharris:
> Who thought a function that doesn't return an error code was a good idea?

Who thought that the pcap callback functions for `pcap_dispatch()` and `pcap_loop()` shouldn't be able to return an error code, so that those routines could exit the loop and return the error code?

The right fix is to add new routines where the callback can return a status code, and then a dump routine that does return an error code could be added and used with those routines.

That could be done as part of the process of adding full pcapng support.

(Note that changes that modify the API or ABI are not allowed.)

infrastation:
This isn't a perfect fix, but it adds a comment with the rationale and an extra `ferror()` check that is not in pull request #1048, so in order to credit the much earlier effort I am going to rebase and to merge this change in a couple days (unless anyone objects).

infrastation:
Merged as commit e6d1929c, thank you.

PR #497, opened 2016-06-14
Change the way NFLOG version is determined

W0naN0w:
The NOT operation is inside the brackets because it only applies to the variable nfhdr->nflog_version, not to the expression.
This check is pretty straightforward but silences a compiler warning. The operator != is not the best usage here, so I replaced it with a solution that made his proof in Nmap project (I am currently working for Nmap organisation).

### DIFF
--- a/pcap-common.c
+++ b/pcap-common.c
@@ -1338,7 +1338,7 @@ swap_nflog_header(const struct pcap_pkthdr *hdr, u_char *buf)
 		return;
 	}
 
-	if (nfhdr->nflog_version != 0) {
+	if ((!nfhdr->nflog_version) == 0) {
 		/* Unknown NFLOG version */
 		return;
 	}


hsluoyz:
I still don't understand where the difference is between ``a != 0`` and ``!a == 0``? And we usually write it in the former style.

pcap-common.c
```diff
@@ -1338,7 +1338,7 @@ swap_nflog_header(const struct pcap_pkthdr *hdr, u_char *buf)
 		return;
 	}
 
-	if (nfhdr->nflog_version != 0) {
+	if ((!nfhdr->nflog_version) == 0) {
```
guyharris: The code is checking whether the NFLOG version, which is stored in the `nflog_version` member of the `nflog_hdr_t` structure pointed to by `nfhdr`, is 0, which is the version we support.

The way you check whether the `nflog_version` member of the `nflog_hdr_t` structure pointed to by `nfhdr`, is 0 is by doing `nfhdr->nflog_version != 0`.  If your compiler can't handle that, get a better compiler.

guyharris:
This is the wrong thing to do, and we will not do it.

guyharris:
Or is your change backwards, and applying to an older version of libpcap?  The *correct* change was made in f94cd1eb7d0ab32bf1db1dd029acc0c68b054363, which is the exact *opposite* of the change shown here.   The

    if (!(nfhdr->nflog_version) == 0) {

code was getting warnings; fixing it to do

    if (nfhdr->nflog_version != 0) {

made the warnings go away.

W0naN0w:
Hi guys, actually, the check before my commit was 	if (!(nfhdr->nflog_version) == 0) and it was working fine. The only thing is the compiler was complaining because it was à bit unclear to him which check is intended: is it the variable you want the NOT operator on or is it the whole expression ? Thought != Is not appropriate here and to silence warning you just have to put the NOT op. "!" Inside the brackets like this :
	if ((!nfhdr->nflog_version) == 0)

guyharris:
> the check before my commit was if (!(nfhdr->nflog_version) == 0)

Then your commit is, as I suggested above, to an older version of libpcap; the code in the current version is

    if (nfhdr->nflog_version != 0) {

not

    if (!(nfhdr->nflog_version) == 0) {

> is it the variable you want the NOT operator on or is it the whole expression

We don't want the NOT operator *at all*.  We want the "not equals" operator, because we want to check whether `nfhdr->nflog_version` is "not equal" to the only version of the NFLOG header that we support, which is version 0.

PR #499, opened 2016-06-25
Fix all build errors on MSVC 2013 together with Npcap source code

hsluoyz:
libpcap repo needs to be placed into Npcap repo's "\wpcap" folder.

### DIFF
--- a/Win32/Prj/libpcap.vcxproj
+++ b/Win32/Prj/libpcap.vcxproj
@@ -123,7 +123,6 @@
     <ClCompile Include="..\..\bpf_image.c" />
     <ClCompile Include="..\..\etherent.c" />
     <ClCompile Include="..\..\fad-helpers.c" />
-    <ClCompile Include="..\..\fad-win32.c" />
     <ClCompile Include="..\..\gencode.c" />
     <ClCompile Include="..\..\grammar.c" />
     <ClCompile Include="..\..\inet.c" />
@@ -140,6 +139,7 @@
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="..\..\pcap-common.h" />
+    <ClInclude Include="..\..\pcap-int.h" />
     <ClInclude Include="..\..\pcap-stdinc.h" />
     <ClInclude Include="..\..\pcap-tc.h" />
     <ClInclude Include="..\..\pcap.h" />
--- a/Win32/Prj/libpcap.vcxproj.filters
+++ b/Win32/Prj/libpcap.vcxproj.filters
@@ -13,9 +13,6 @@
     <ClCompile Include="..\..\etherent.c">
       <Filter>Source Files</Filter>
     </ClCompile>
-    <ClCompile Include="..\..\fad-win32.c">
-      <Filter>Source Files</Filter>
-    </ClCompile>
     <ClCompile Include="..\..\gencode.c">
       <Filter>Source Files</Filter>
     </ClCompile>
@@ -78,6 +75,9 @@
       <Filter>Header Files</Filter>
     </ClInclude>
     <ClInclude Include="..\..\pcap.h">
+      <Filter>Header Files</Filter>
+    </ClInclude>
+    <ClInclude Include="..\..\pcap-int.h">
       <Filter>Header Files</Filter>
     </ClInclude>
   </ItemGroup>
--- a/pcap-int.h
+++ b/pcap-int.h
@@ -409,6 +409,11 @@ int	pcap_offline_read(pcap_t *, int, pcap_handler, u_char *);
  * #defining them to be snprintf or vsnprintf, respectively, or by
  * defining our own versions and exporting them.
  */
+
+#if defined(WIN32) && _MSC_VER < 1900
+#define snprintf _snprintf
+#endif
+
 #ifdef HAVE_SNPRINTF
 #define pcap_snprintf snprintf
 #else


pcap/pcap.h
```diff
@@ -37,6 +37,10 @@
 
 #include <pcap/export-defs.h>
 
+#if defined(WIN32) && _MSC_VER < 1900
```
guyharris: We shouldn't be using snprintf() anywhere - we should be using pcap_snprintf().

_snprintf(), unfortunately, does *not* work the same way as C99 snprintf() - it does not guarantee that the resulting string is null-terminated, so we don't use it directly.  Instead, missing/win_snprintf.c has a pcap_vsnprintf() routine, which calls _vsnprintf() and then forcibly null-terminates the string; pcap_snprintf() uses pcap_vsnprintf().

So this change shouldn't be necessary.

fad-win32.c
```diff
@@ -42,6 +42,107 @@
 
 #include <errno.h>
 
+/*
+* Add an entry to the list of addresses for an interface.
+* "curdev" is the entry for that interface.
+*/
+static int
```
guyharris: Why is this routine necessary?  What does it do that add_addr_to_dev() doesn't do (other than depend on struct sockaddr_storage existing)?  There are some things add_addr_to_dev() does that this routine does *NOT* do, such as:

- properly free previously-allocated memory if a later allocation fails;

- use `pcap_snprintf()` rather than `snprintf()`.

Furthermore, it doesn't have the size arguments, but your later call to it *does* pass the size arguments, so the code won't even compile.

Win32/Prj/libpcap.vcxproj
```diff
@@ -109,26 +113,41 @@
...
+    <ClCompile Include="..\..\..\Win32-Extensions\grammar.c" />
     <ClCompile Include="..\..\inet.c" />
     <ClCompile Include="..\..\nametoaddr.c" />
     <ClCompile Include="..\..\optimize.c" />
+    <ClCompile Include="..\..\pcap-common.c" />
     <ClCompile Include="..\..\pcap-tc.c" />
     <ClCompile Include="..\..\Pcap-win32.c" />
     <ClCompile Include="..\..\pcap.c" />
     <ClCompile Include="..\..\savefile.c" />
-    <ClCompile Include="..\..\scanner.c" />
```
guyharris: Again, the build process should run Flex and Bison, which should put those source files in the main source directory, so that change shouldn't be necessary.

Win32/Prj/libpcap.vcxproj
```diff
@@ -109,26 +113,41 @@
...
+    <ClCompile Include="..\..\..\Win32-Extensions\scanner.c" />
+    <ClCompile Include="..\..\sf-pcap-ng.c" />
+    <ClCompile Include="..\..\sf-pcap.c" />
   </ItemGroup>
   <ItemGroup>
+    <ClInclude Include="..\..\..\Win32-Extensions\bitset.h" />
```
guyharris: Nothing uses bitset.h, so this change shouldn't be necessary.

Win32/Prj/libpcap.vcxproj
```diff
@@ -109,26 +113,41 @@
...
+    <ClInclude Include="..\..\..\Win32-Extensions\grammar.h" />
+    <ClInclude Include="..\..\..\Win32-Extensions\scanner.h" />
```
guyharris: Again, the build process should run Flex and Bison, which should put those source files in the main source directory, so those two changes shouldn't be necessary.

Win32/Prj/libpcap.vcxproj
```diff
@@ -109,26 +113,41 @@
...
+    <ClInclude Include="..\..\..\Win32-Extensions\TcApi.h" />
```
guyharris: The TurboCap library isn't part of libpcap/WinPcap/Npcap, it's a separate library provided by Riverbed, so it should be looking for that header externally.

Win32/Prj/libpcap.vcxproj
```diff
@@ -109,26 +113,41 @@
...
+    <ClInclude Include="..\..\..\Win32-Extensions\tokdefs.h" />
+    <ClInclude Include="..\..\pcap-common.h" />
```
guyharris: Again, the build process should run Flex and Bison, which should put those source files in the main source directory, so that change shouldn't be necessary.

Win32/Prj/libpcap.vcxproj.filters
```diff
@@ -19,9 +19,6 @@
     <ClCompile Include="..\..\gencode.c">
       <Filter>Source Files</Filter>
     </ClCompile>
-    <ClCompile Include="..\..\grammar.c">
```
guyharris: Again, the build process should run Flex and Bison, which should put those source files in the main source directory, so that change shouldn't be necessary.

Win32/Prj/libpcap.vcxproj.filters
```diff
@@ -43,7 +40,22 @@
     <ClCompile Include="..\..\savefile.c">
       <Filter>Source Files</Filter>
     </ClCompile>
-    <ClCompile Include="..\..\scanner.c">
+    <ClCompile Include="..\..\..\Win32-Extensions\grammar.c">
+      <Filter>Source Files</Filter>
+    </ClCompile>
+    <ClCompile Include="..\..\..\Win32-Extensions\scanner.c">
```
guyharris: Again, the build process should run Flex and Bison, which should put those source files in the main source directory, so those two changes shouldn't be necessary.

Win32/Prj/libpcap.vcxproj.filters
```diff
@@ -59,5 +71,29 @@
     <ClInclude Include="..\..\pcap-tc.h">
       <Filter>Header Files</Filter>
     </ClInclude>
+    <ClInclude Include="..\..\..\Win32-Extensions\bitset.h">
```
guyharris: Again, othing uses bitset.h, so this change shouldn't be necessary.

Win32/Prj/libpcap.vcxproj.filters
```diff
@@ -59,5 +71,29 @@
...
+      <Filter>Header Files</Filter>
+    </ClInclude>
+    <ClInclude Include="..\..\..\Win32-Extensions\grammar.h">
```
guyharris: Again, the build process should run Flex and Bison, which should put those source files in the main source directory, so those two changes shouldn't be necessary.

Win32/Prj/libpcap.vcxproj.filters
```diff
@@ -59,5 +71,29 @@
...
+      <Filter>Header Files</Filter>
+    </ClInclude>
+    <ClInclude Include="..\..\..\Win32-Extensions\scanner.h">
+      <Filter>Header Files</Filter>
+    </ClInclude>
+    <ClInclude Include="..\..\..\Win32-Extensions\TcApi.h">
```
guyharris: Again, the TurboCap library isn't part of libpcap/WinPcap/Npcap, it's a separate library provided by Riverbed, so it should be looking for that header externally.

Win32/Prj/libpcap.vcxproj
```diff
@@ -109,26 +113,41 @@
...
+    <ClCompile Include="..\..\..\Win32-Extensions\scanner.c" />
+    <ClCompile Include="..\..\sf-pcap-ng.c" />
+    <ClCompile Include="..\..\sf-pcap.c" />
   </ItemGroup>
   <ItemGroup>
+    <ClInclude Include="..\..\..\Win32-Extensions\bitset.h" />
+    <ClInclude Include="..\..\..\Win32-Extensions\grammar.h" />
+    <ClInclude Include="..\..\..\Win32-Extensions\scanner.h" />
+    <ClInclude Include="..\..\..\Win32-Extensions\TcApi.h" />
+    <ClInclude Include="..\..\..\Win32-Extensions\tokdefs.h" />
```
guyharris: Again, the build process should run Flex and Bison, which should put those source files in the main source directory, so that change shouldn't be necessary.

Win32/Prj/libpcap.vcxproj.filters
```diff
@@ -59,5 +71,29 @@
...
+      <Filter>Header Files</Filter>
+    </ClInclude>
+    <ClInclude Include="..\..\..\Win32-Extensions\tokdefs.h">
```
guyharris: Again, the build process should run Flex and Bison, which should put those source files in the main source directory, so that change shouldn't be necessary.

guyharris:
I've checked in some changes; the other ones have problems - I've made comments to the diffs.

pcap/pcap.h
hsluoyz: Hi @guyharris ,

In https://github.com/the-tcpdump-group/libpcap/blob/master/pcap-int.h#L412-L422,

``pcap_snprintf`` is also defined as ``snprintf`` if ``HAVE_SNPRINTF`` is defined (which is actually defined in the projecr file). And my VS 2013 doesn't define ``snprintf``, so we still need to face this issue. How do you suggest to solve it then?

guyharris: > How do you suggest to solve it then?

Define `HAVE_SNPRINTF` only if you actually have `snprintf()`.

Maybe it's impossible to do except by using CMake; if so, then the answer is "use CMake to generate MSVC project files, don't supply your own project files".

Win32/Prj/libpcap.vcxproj
hsluoyz: OK.

Win32/Prj/libpcap.vcxproj
hsluoyz: OK.

Win32/Prj/libpcap.vcxproj
hsluoyz: OK. I will do this later. For now, I will just copy ``grammar.c`` and ``scanner.c`` to the main source directory to achieve the same effect.

Win32/Prj/libpcap.vcxproj.filters
hsluoyz: OK.

Win32/Prj/libpcap.vcxproj.filters
hsluoyz: OK.

Win32/Prj/libpcap.vcxproj.filters
hsluoyz: OK.

Win32/Prj/libpcap.vcxproj.filters
hsluoyz: OK.

Win32/Prj/libpcap.vcxproj.filters
hsluoyz: OK.

PR #500, opened 2016-06-27
Fix some issues about libpcap Windows building.

hsluoyz:


pcap-int.h
```diff
@@ -409,6 +409,11 @@ int	pcap_offline_read(pcap_t *, int, pcap_handler, u_char *);
  * #defining them to be snprintf or vsnprintf, respectively, or by
  * defining our own versions and exporting them.
  */
+
```
guyharris: The only time `snprintf()` will be referred to is if `HAVE_SNPRINTF` is defined.

If `HAVE_SNPRINTF` is defined for versions of MSVC that don't provide `snprintf()`, that's a bug in whatever's defining `HAVE_SNPRINTF`, and the way to fix it is not to define `HAVE_SNPRINTF`.

I.e., never ever ever ever ever ever define `HAVE_SNPRINTF` in an MSVC project file unless the project file somehow checks the version of MSVC and defines it only in versions of MSVC that have it.  If an MSVC project file can't do that, then you need to do something such as

    #if defined(_MSC_VER)
        #if _MSC_VER >= 1900
            #define HAVE_SNPRINTF
        #endif
    #endif

(don't assume that, on Windows, this is being compiled by MSVC, it could be compiled by MinGW's GCC).

Note, however, that `pcap-stdinc.h` is already doing checks of this sort, so that might be the place to add that.

gvanem:
I don't understand how removing `fad-win32.c` from the project can work; `pcap_findalldevs_interfaces()` is called from `pcap.c`.

guyharris:
> `pcap_findalldevs_interfaces()` is called from `pcap.c`.

Not as of 881a4063fb3c129e07e17e1ad96f2c0dec622014 it isn't.

hsluoyz:
OK. Can you pass the other two commits?

guyharris:
If you could make them two separate pull requests, I can pull them separately.

hsluoyz:
OK. I have submitted a new PR.

PR #504, opened 2016-06-28
Add back WinPcap specific functions to libpcap, no compile errors now.

hsluoyz:
Wireshark x64 can capture packets normally under this version ``wpcap.dll`` built from libpcap source. I didn't test other use cases.

For now, only 4 functions are still missing compared to WinPcap:
1. endservent()
2. getservent()
3. wsockinit()
4. pcap_read()

I think 1. and 2. are needed to be added back later. 3. and 4. are some functions maybe we can ignore.

### DIFF
--- a/nametoaddr.c
+++ b/nametoaddr.c
@@ -317,7 +317,7 @@ struct eproto {
  * don't officially export it by declaring it in a header file.
  * (Programs *should* do this themselves, as tcpdump now does.)
  */
-PCAP_API_DEF struct eproto eproto_db[] = {
+struct eproto eproto_db[] = {
 	{ "pup", ETHERTYPE_PUP },
 	{ "xns", ETHERTYPE_NS },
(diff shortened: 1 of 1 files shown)


pcap-int.h
```diff
@@ -107,6 +107,107 @@ extern "C" {
  */
 #define MAXIMUM_SNAPLEN		262144
 
+/*
```
guyharris: Libpcap no longer has a `pcap_md` structure; instead, a `pcap_t` has a `void *` pointer to a private data structure, and each `pcap-XXX.c` file has its own private data structure, defined internally in the `pcap-XXX.c` file, to which that points.  We will not reintroduce the `pcap_md` structure.

hsluoyz: OK, is there any other issues about this PR? I mean I can fix them all together:) So the next PR can be merged with one try.

PR #505, opened 2016-06-30
Do not export eproto_db any more based on this post: http://seclists.org/tcpdump/2016/q2/60

hsluoyz:


guyharris:
One thing the comment says is

     * tcpdump used to import this, and it's declared as an export on
     * Debian, at least, so make it a public symbol, even though we
     * don't officially export it by declaring it in a header file.

so if we don't export it, Debian (and its derivatives, such as Ubuntu) might have to deem this an ABI change, which could be disruptive.

In addition, OSes that might have closed-source programs linking with libpcap, such as OS X, might also have a problem with this, as they can't necessarily say "go rewrite your program and make a new version to have it work on a new version of the OS".

So, at least on UN\*X, we should continue to export it, even if we don't put a declaration of it into any of our header files.

And, on Windows, I think WinDump may import `eproto_db` from WinPcap; if so, other programs might be doing that, in which case Npcap would break binary compatibility for those programs.

PR #513, opened 2016-07-28
Fix some compile warnings reported by MSVC.

hsluoyz:
Mainly for Windows:
``strncpy, strcpy`` -> ``strlcpy`` -> ``strncpy_s``
``strncat, strcat`` -> ``strlcat`` -> ``strncat_s``
``strtok`` -> ``strltok`` -> ``strtok_s``
``scanf`` -> ``scanf_s``
``sprintf`` -> ``pcap_snprintf``
Add ``(u_int)`` to some comparisons

### DIFF
--- a/inet.c
+++ b/inet.c
@@ -233,7 +233,10 @@ pcap_lookupdev(errbuf)
 	DWORD dwWindowsMajorVersion;
 	char our_errbuf[PCAP_ERRBUF_SIZE+1];
 
+#pragma warning (push)
+#pragma warning (disable: 4996) /* disable MSVC's GetVersion() deprecated warning here */
 	dwVersion = GetVersion();	/* get the OS version */
+#pragma warning (pop)
 	dwWindowsMajorVersion = (DWORD)(LOBYTE(LOWORD(dwVersion)));
 
 	if (dwVersion >= 0x80000000 && dwWindowsMajorVersion >= 4) {
--- a/pcap-new.c
+++ b/pcap-new.c
@@ -166,7 +166,7 @@ int pcap_findalldevs_ex(char *source, struct pcap_rmtauth *auth, pcap_if_t **all
 			}
 
 			/* Copy the new device identifier into the correct memory location */
-			strncpy(dev->name, tmpstring, strlen(tmpstring) + 1);
+			strlcpy(dev->name, tmpstring, strlen(tmpstring) + 1);
 
 
 			/* Create the new device description */
@@ -189,7 +189,7 @@ int pcap_findalldevs_ex(char *source, struct pcap_rmtauth *auth, pcap_if_t **all
 			}
 
 			/* Copy the new device description into the correct memory location */
-			strncpy(dev->description, tmpstring, strlen(tmpstring) + 1);
+			strlcpy(dev->description, tmpstring, strlen(tmpstring) + 1);
 
 			dev = dev->next;
 		}
@@ -312,7 +312,7 @@ int pcap_findalldevs_ex(char *source, struct pcap_rmtauth *auth, pcap_if_t **all
 					return -1;
 				}
 
-				strncpy(dev->name, tmpstring, stringlen);
+				strlcpy(dev->name, tmpstring, stringlen);
 
 				dev->name[stringlen] = 0;
 
@@ -331,7 +331,7 @@ int pcap_findalldevs_ex(char *source, struct pcap_rmtauth *auth, pcap_if_t **all
 				}
 
 				/* Copy the new device description into the correct memory location */
-				strncpy(dev->description, tmpstring, stringlen + 1);
+				strlcpy(dev->description, tmpstring, stringlen + 1);
 
 				pcap_close(fp);
 			}
@@ -514,7 +514,7 @@ int pcap_findalldevs_ex(char *source, struct pcap_rmtauth *auth, pcap_if_t **all
 			}
 
 			/* Copy the new device name into the correct memory location */
-			strncpy(dev->name, tmpstring2, stringlen + 1);
+			strlcpy(dev->name, tmpstring2, stringlen + 1);
 		}
 
 		if (findalldevs_if.desclen)
@@ -546,7 +546,7 @@ int pcap_findalldevs_ex(char *source, struct pcap_rmtauth *auth, pcap_if_t **all
 			}
 
 			/* Copy the new device description into the correct memory location */
-			strncpy(dev->description, tmpstring2, stringlen + 1);
+			strlcpy(dev->description, tmpstring2, stringlen + 1);
 		}
 
 		dev->flags = ntohl(findalldevs_if.flags);
@@ -668,10 +668,10 @@ int pcap_createsrcstr(char *source, int type, const char *host, const char *port
 	{
 	case PCAP_SRC_FILE:
 	{
-		strncpy(source, PCAP_SRC_FILE_STRING, PCAP_BUF_SIZE);
+		strlcpy(source, PCAP_SRC_FILE_STRING, PCAP_BUF_SIZE);
 		if ((name) && (*name))
 		{
-			strncat(source, name, PCAP_BUF_SIZE);
+			strlcat(source, name, PCAP_BUF_SIZE);
 			return 0;
 		}
 		else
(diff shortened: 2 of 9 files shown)


pcap-new.c
```diff
@@ -982,7 +982,7 @@ pcap_t *pcap_open(const char *source, int snaplen, int flags, int read_timeout,
 		break;
 
 	default:
-		strcpy(errbuf, "Source type not supported");
+		pcap_snprintf(errbuf, PCAP_ERRBUF_SIZE, "Source type not supported");
 		return NULL;
```
guyharris: This should probably just be

    strlcpy(errbuf, PCAP_ERRBUF_SIZE, "Source type not supported");

because it's just a copy - no string formatting is being done.

sockutils.c
```diff
@@ -65,8 +62,12 @@ int sockcount = 0;					/* Variable that allows calling the WSAStartup() only one
 
 /* Some minor differences between UNIX and Win32 */
 #ifdef WIN32
-#define SHUT_WR SD_SEND			/* The control code for shutdown() is different in Win32 */
-#define snprintf _snprintf		/* The snprintf is called _snprintf() in Win32 */
+#define SHUT_WR SD_SEND				/* The control code for shutdown() is different in Win32 */
+#define snprintf(x, y, z, ...) \
```
guyharris: This shouldn't be necessary as of 905467b72b83897216b172f440f38d14ba00953b - everything should be using `pcap_snprintf()`, which is guaranteed to always null-terminate the string.

pcap-new.c
hsluoyz: OK. I will change it to ``strlcpy``.

sockutils.c
```diff
@@ -899,9 +900,9 @@ int sock_check_hostlist(char *hostlist, const char *sep, struct sockaddr_storage
 		 * The problem is that strtok modifies the original variable by putting '0' at the end of each token
 		 * So, we have to create a new temporary string in which the original content is kept
 		 */
-		strcpy(temphostlist, hostlist);
+		strlcpy(temphostlist, hostlist, strlen(hostlist) + 1);
```
guyharris: ce15901cc90eef4b82c67d26e203c2a5a6cce1a2 should eliminate the need for this change.

hsluoyz: OK.

sockutils.c
hsluoyz: OK.

nametoaddr.c
```diff
@@ -126,7 +126,14 @@ pcap_nametoaddr(const char *name)
 	bpf_u_int32 **p;
 	struct hostent *hp;
 
+#ifdef _WIN32
+#pragma warning (push)
+#pragma warning (disable: 4996) /* disable MSVC's gethostbyname() warning here */
+	if ((hp = gethostbyname(name)) != NULL) {
+#pragma warning (pop)
```
gvanem: Wouldn't it be better to add a `#define _WINSOCK_DEPRECATED_NO_WARNINGS` (or in `CLFAGS`) instead of "*polluting*" the code with these pragmas? Besides these pragmas should be inside an `#ifdef _MSC_VER`.

Similarily for `GetVersion()` where one can define `BUILD_WINDOWS.

hsluoyz: I'm OK with replacing ``disable: 4996`` with ``_WINSOCK_DEPRECATED_NO_WARNINGS``.

But I think ``#pragma warning (push)`` and ``#pragma warning (pop)`` are necessary because only this one statement needs to disable the warning. If we put the disabling code in a header, the whole following code will disable the warning, which is not what we need.

And why we need an ``#ifdef _MSC_VER``? The warning disabling doesn't work on all MSVCs?

gvanem: > If we put the disabling code in a header, the whole following code will disable the warning, which is not what
> we need.

Ok, but how about:
```
+#if defined(BUILDING_LIBPCAP)
+  #define  _WINSOCK_DEPRECATED_NO_WARNINGS
+#endif
+
#include <winsock2.h>
```
in `pcap-stdinc.h`.

> And why we need an #ifdef _MSC_VER? 

Since the pragma is effective for MSVC only. And since the warning is *not* really a warning from the compiler. But a warning caused by *recent* changes in the MS-SDK. Have you tried building with e.g. MinGW and checked the warnings from these unknown pragmas?

hsluoyz: > Ok, but how about:

I think many source files have included ``pcap-stdinc.h``. Disabling this warning in this manner will still cause all warnings of ``_WINSOCK_DEPRECATED_NO_WARNINGS`` type to be disabled in those source files when building libpcap. This is also not good.

> Since the pragma is effective for MSVC only. And since the warning is not really a warning from the compiler. But a warning caused by recent changes in the MS-SDK.

I'm using MSVC 2013 for building libpcap. MSVC 2013 is also the official compiler for official WinPcap, Wireshark and Nmap. So I think MSVC 2013 can be represented as MSVC here.

> Have you tried building with e.g. MinGW and checked the warnings from these unknown pragmas?

Np. I didn't try to build on ``MinGW``. I don't know how it works on ``MinGW``. We can change it to ``#if defined(_WIN32) && defined (_MSC_VER)``.

pcap-remote.c
```diff
@@ -2126,7 +2128,7 @@ int rpcap_remoteact_getsock(const char *host, char *errbuf)
 		while (ai_next)
 		{
 			if (sock_cmpaddr(&temp->host, (struct sockaddr_storage *) ai_next->ai_addr) == 0)
-				return (temp->sockctrl);
+				return (int) (temp->sockctrl);
```
guyharris: I cleaned up the code in 8a36b5e79defb6a917d257d81cb986ad350ba881, so this shouldn't be necessary.

sockutils.c
```diff
@@ -51,10 +51,7 @@
 #include <stdio.h>	/* for the stderr file */
 #include <stdlib.h>	/* for malloc() and free() */
 
-
-
-
-
+#include "pcap-int.h"
```
guyharris: That shouldn't be necessary after change .

pcap-int.h
```diff
@@ -375,6 +379,14 @@ int	pcap_offline_read(pcap_t *, int, pcap_handler, u_char *);
    */
   #define strlcpy(x, y, z) \
 	strncpy_s((x), (z), (y), _TRUNCATE)
+  #define strlcat(x, y, z) \
```
guyharris: These should go in the new `portability.h` header.

nametoaddr.c
hsluoyz: How about this

```
#if defined(_WIN32) && defined(_MSC_VER) && undefined(_WINSOCK_DEPRECATED_NO_WARNINGS)
#define _WINSOCK_DEPRECATED_NO_WARNINGS
if ((hp = gethostbyname(name)) != NULL) {
#undef _WINSOCK_DEPRECATED_NO_WARNINGS
#endif
```

pcap-remote.c
hsluoyz: OK. I will remove this part of code.

sockutils.c
hsluoyz: OK.

pcap-int.h
hsluoyz: OK.

pcap-remote.c
```diff
@@ -513,12 +513,14 @@ static int pcap_stats_remote(pcap_t *p, struct pcap_stat *ps)
...
  * This function is just a void cointainer, since the work is done by the rpcap_stats_remote().
- * See that funcion for more details.
+ * See that function for more details.
  *
  * Parameters and return values are exactly the same of the pcap_stats_ex().
  */
-static struct pcap_stat *pcap_stats_ex_remote(pcap_t *p)
+static struct pcap_stat *pcap_stats_ex_remote(pcap_t *p, int *pcap_stat_size)
 {
+	*pcap_stat_size = sizeof(p->stat);
```
guyharris: I checked in a fix for that particular issue in eeee5694eb0620ec38423700d7a9e7bd26c87dd2.

inet.c
```diff
@@ -296,7 +299,7 @@ pcap_lookupdev(errbuf)
 		while(NAdapts--)
 		{
 			char* tmp = (char*)tUstr;
-			strcpy(tmp, tAstr);
+			strlcpy(tmp, tAstr, 8192);
```
guyharris: This code wasn't doing any bounds checking; I rewrote it to do bounds checking on the buffer, so it should now be OK without that change (it now uses `memcpy()` to do the copy - it's already determined the size of the string to copy, so that it can check if there's enough room in the buffer for the copy, so it can just copy all the bytes of the string, including the terminating NUL).

hsluoyz: OK.

nametoaddr.c
guyharris: Or maybe we should use `getaddrinfo()` if we have it, even for IPv4 addresses.

hsluoyz: I was thinking of that before. But I found the returned ``h_addr_list`` is hard to handle. I even doubt that is there a memory leak here? Because ``h_addr_list``'s memory is never freed after the function call. So i'm stuck here.

pcap-remote.c
hsluoyz: No. MSVC only reports a warning. If that function is directly called, then it should be an error.

bpf/net/bpf_filter.c
```diff
@@ -579,7 +579,12 @@ bpf_filter_with_aux_data(pc, p, wirelen, buflen, aux_data)
 			continue;
 
 		case BPF_ALU|BPF_NEG:
+#ifdef _WIN32
+#pragma warning (push)
+#pragma warning (disable: 4146) /* disable MSVC's unary minus operator warning here */
```
guyharris: 2fc98f9a355a97a69f8701cd7822ed5d6537c579 should get rid of that warning.

hsluoyz: OK. I will remove this part of code.

PR #514, opened 2016-07-29
Made pcap_set_timeout() function work even after pcap device is activated on Windows.

hsluoyz:
This commit is to fix this issue: https://github.com/nmap/nmap/issues/426

I will paste some of comments from that issue here:

--------------
Now Npcap has moved to latest libpcap 1.8.0. So it seems that we can use the new **immediate mode** @guyharris mentioned now. I think all we need to do is to call ``pcap_set_immediate_mode`` before ``pcap_activate``, then remove the ``PacketSetReadTimeout`` call, right?

But according to @dmiller-nmap , our goal is ``setting varying timeouts on a single pcap descriptor``. So it seems that we need to change this timeout when the adapter is opened. Not just setting the mode to immediate before activating the adapter. Do libpcap have a way to do this?

### DIFF
--- a/pcap.c
+++ b/pcap.c
@@ -655,7 +655,15 @@ int
 pcap_set_timeout(pcap_t *p, int timeout_ms)
 {
 	if (pcap_check_activated(p))
+#ifdef _WIN32
+		if (p->opt.timeout != timeout_ms)
+			if (PacketSetReadTimeout(p->adapter, p->opt.timeout))
+				return 0;
+			else
+				return PCAP_ERROR;
+#else
 		return (PCAP_ERROR_ACTIVATED);
+#endif
 	p->opt.timeout = timeout_ms;
 	return (0);
 }


guyharris:
I suspect the *ultimate* goal is to make `readip_pcap()` in Nmap do what it's supposed to do; see my long discussion(s) in nmap/nmap#426.  Whether making `pcap_set_timeout()` work on an activated device is necessary for that has yet to be determined, so I would wait until that's determined before merging this.

Changing the behavior of `pcap_set_timeout()` runs the risk of causing people to write code for Windows that assumes that it works and then try to port that code to various UN\*Xes under the assumption that it works there as well.  It *won't* work, for example, on Linux when TPACKET_V3 is used; once you've selected the timeout, the only way to change it is to destroy the ring buffer and re-create it, which is a somewhat expensive operation *and* which throws away all the packets in the ring buffer.

So if this is necessary on Windows, it should be provided by a *separate* API, and that API should *only* be provided on Windows, so that if you try to compile the program for Linux, the compile or link will fail, indicating that It Won't Work.

In addition, that separate API should *not* assume that the `pcap_t` is necessarily a `pcap-win32.c` device - it might, for example, be a TurboCap device or a remote capture handle.

So there should be a `set_timeout_op` in the `pcap_t` structure, set by the activate routine, and called by the new API.

But, again, that assumes that we should do this at all, rather than having the Nmap people re-examine what they're trying to do, taking into account all the details I gave on how timeouts work and how `select()` on UN\*X and SomethingWaitForMultipleSomething calls on Windows work with `pcap_t`s.

hsluoyz:
Since the issue this PR tries to solve has been fixed in https://github.com/nmap/nmap/issues/426, I will close this PR.

Issue #518, opened 2016-08-12
Has anywhere in libpcap defined the version string (like "1.8.0")?


hsluoyz:
I have [a C header](https://github.com/nmap/npcap/blob/master/wpcap/version.h#L7) in Npcap with content:
```c
//{{NO_DEPENDENCIES}}
// Microsoft Developer Studio generated include file.
//

#include "..\version.h"

static const char pcap_version_string[] = "libpcap version " WINPCAP_LIBPCAP_VERSION;
```

And I defined the macro ``WINPCAP_LIBPCAP_VERSION`` as "1.8.0", so this will be returned in ``pcap_lib_version()`` and shown in softwares like Wireshark's About window like ``with Npcap version 0.08, based on libpcap version 1.8.0``. However, I think it's better if this string is defined in libpcap instead of Npcap. Is there any string like this? Thanks.

infrastation:
Have you checked in the source code?

hsluoyz:
OK. I saw this file: https://github.com/the-tcpdump-group/libpcap/blob/master/VERSION. You bumped this file together with the new git tag.

I think I need "include" this file in my C header. But since it only contains a version string, I don't know how to include it in a C syntax..

fxlb:
You can see `version.c` building in `Makefile.in`.

fxlb:
and http://www.tcpdump.org/manpages/pcap_lib_version.3pcap.html.

guyharris:
As @fxlb noted, it's generated in `Makefile.in` from `VERSION` using various UN\*X commands; `version.h` is also generated in a similar fashion.

I don't know whether there's a way, using only what's available from within a Visual Studio project and tools available on Windows, to turn a file containing a version string (as text, not a C string) into a C source or header file that defines or declares a string based on that version string.  If there is, that's what you'd need to do.

guyharris:
So this is now fixed, and this can be closed?

hsluoyz:
I found a bug in in my code: https://github.com/the-tcpdump-group/libpcap/pull/520

PR #519, opened 2016-08-12
Libpcap version can be automatically generated based on the source.

hsluoyz:
This PR is to fix this issue: https://github.com/the-tcpdump-group/libpcap/issues/518
The corresponding change for Npcap is here: https://github.com/nmap/npcap/commit/db8b65da23a77c46624ea6278c15ade077a9e386

### DIFF
--- a/Win32/Prj/wpcap.vcxproj
+++ b/Win32/Prj/wpcap.vcxproj
@@ -113,7 +113,8 @@
       <AdditionalDependencies>ws2_32.lib;..\..\..\..\packetWin7\Dll\Project\Release No NetMon and AirPcap\Packet.lib;%(AdditionalDependencies)</AdditionalDependencies>
     </Link>
     <PreBuildEvent>
-      <Command>win_flex -Ppcap_ -7 --outfile=..\..\scanner.c --header-file=..\..\scanner.h ..\..\scanner.l
+      <Command>..\..\GenVersion.bat
+win_flex -Ppcap_ -7 --outfile=..\..\scanner.c --header-file=..\..\scanner.h ..\..\scanner.l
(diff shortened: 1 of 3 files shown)


Win32/Prj/wpcap.vcxproj
```diff
@@ -113,7 +113,8 @@
       <AdditionalDependencies>ws2_32.lib;..\..\..\..\packetWin7\Dll\Project\Release No NetMon and AirPcap\Packet.lib;%(AdditionalDependencies)</AdditionalDependencies>
     </Link>
     <PreBuildEvent>
-      <Command>win_flex -Ppcap_ -7 --outfile=..\..\scanner.c --header-file=..\..\scanner.h ..\..\scanner.l
+      <Command>call ..\..\..\GenVersion.bat
```
guyharris: You need to add `GenVersion.bat` to the commit; `git add` it and then do `git commit --amend` and push the results up to GitHub (or whatever the equivalent is if you're not using the Git command-line tools).

hsluoyz: OK. So I will move all the files into libpcap including the generated ``version.h''.

guyharris: What do you mean by "move all the files into libpcap"?  The generated `version.h` shouldn't be checked into the Git repository, and the generated `version.c` shouldn't be, either.

All you need to do is add `GenVersion.bat` as a file in the top-level directory, along with this change.  You can update this commit to include `GenVersion.bat`, as well as the changes to the project file, the way I described.

hsluoyz: By "all the files" I mean ``GenVersion.bat`` and ``version.h.in``. Currently they are both in Npcap.  The generated ``version.h`` is not checked in certainly.

PR #531, opened 2016-10-06
Rename configure.in to configure.ac

doughdemon:


infrastation:
Could you review the build failure and come up with a working set of changes and some rationale? Thank you.

doughdemon:
Where can I see the build failure? (Locally it compiles fine.)

Anyway, I've now replaced configure.in by configure.ac everywhere, in
particular in the Makefile. Likely the remaining references in the
Makefile have caused the build failure. The commit now also has a more
detailed commit message.

infrastation:
The updated version does pass the build test, you can see it on the pull request page.

PR #533, opened 2016-10-06
Don't need to open the interface any more in order to retrieve the loopback flag on Windows.

hsluoyz:
In order to accommodate this pull request, Npcap's Packet API is also changed here: https://github.com/nmap/npcap/commit/73edca1b863cb9fe90a4b92a7a4996a2421b2701 and will take effect in next Npcap 0.10 r7 release.

And BTW this this ``PCAP_IF_LOOPBACK`` flag value still in use? This code is even not enabled for some time, so the flag is not set. However, the loopback interface works just fine with Wireshark or Nmap.


guyharris:
> And BTW this this PCAP_IF_LOOPBACK flag value still in use?

Yes.  Tcpdump uses it, and Wireshark uses it.  It's part of the libpcap API; it's not an internal flag.

PR #541, opened 2016-11-17
Fix compile when INET6 is not defined

fbertux:
These two commits fix compile when INET6 is disable.


guyharris:
Fixed in 470df104c6f55f6d6f390df7448d8eb65c7642b9.

PR #551, opened 2017-01-27
Add Endace DAG large stream support

sfd:
Support capture from DAG capture streams with buffer size configured >2GB.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -852,7 +852,7 @@ if test "$want_dag" != no; then
 		# included if there's a found-action (arg 3).
 		saved_ldflags=$LDFLAGS
 		LDFLAGS="-L$dag_lib_dir"
-		AC_CHECK_LIB([dag], [dag_attach_stream], [dag_streams="1"], [dag_streams="0"])
+		AC_CHECK_LIB([dag], [dag_attach_stream64], [dag_large_streams="1"], [dag_large_streams="0"])
 		AC_CHECK_LIB([dag],[dag_get_erf_types], [
 			AC_DEFINE(HAVE_DAG_GET_ERF_TYPES, 1, [define if you have dag_get_erf_types()])])
 		AC_CHECK_LIB([dag],[dag_get_stream_erf_types], [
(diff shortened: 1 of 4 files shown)


guyharris:
Note that we'll be regenerating the configure script, so the changes due to using a different version of autoconf won't be preserved.

I presume you're intentionally dropping support for DAG libraries that don't have stream API support.

sfd:
I wasn't sure if supplying the generated files was necessary, so I put them in a separate commit.

Dropping support for very old DAG API versions is deliberate, these are nearly 10 years old and should not be in use. Dropping them simplifies the code, improving maintainability,

configure.ac
```diff
@@ -852,16 +852,16 @@ if test "$want_dag" != no; then
 		# included if there's a found-action (arg 3).
 		saved_ldflags=$LDFLAGS
 		LDFLAGS="-L$dag_lib_dir"
-		AC_CHECK_LIB([dag], [dag_attach_stream], [dag_streams="1"], [dag_streams="0"])
```
guyharris: Removing this means that the configure script won't fail if the library doesn't have `dag_attach_stream()`.  I added a change in 283a5c415ea214fe4b2b53a29a620f73ad949057 to do so.

configure.ac
```diff
@@ -852,16 +852,16 @@ if test "$want_dag" != no; then
...
 		AC_CHECK_LIB([dag],[dag_get_stream_erf_types], [
 			AC_DEFINE(HAVE_DAG_GET_STREAM_ERF_TYPES, 1, [define if you have dag_get_stream_erf_types()])])
 
 		LDFLAGS=$saved_ldflags
 
-		if test "$dag_streams" = 1; then
-			AC_DEFINE(HAVE_DAG_STREAMS_API, 1, [define if you have streams capable DAG API])
+		if test "$dag_large_streams" = 1; then
+			AC_DEFINE(HAVE_DAG_LARGE_STREAMS_API, 1, [define if you have large streams capable DAG API])
 			LIBS="$LIBS -ldag"
```
guyharris: Was it formerly intended to link with `-ldag` only if you had streams support, and now intended that you only link with `-ldag` if you have large streams support?  I suspect not, so I moved the `LIBS=` and `LDFLAGS=` lines before the check for `$dag_large_streams` in 283a5c415ea214fe4b2b53a29a620f73ad949057.

PR #552, opened 2017-02-05
since the chunks are unconditionally torn down (L760), need to uncond…

bch:
…itionally set them up (or rejig the teardown); nuke the short-circuit, initialize.


bch:
W/o this change, if one tries to apply a filter to a non-activated handle, freechunks() may fault attempting to free uninitialized pointers.

guyharris:
Fixed differently in ada95c9321eab3793d485ca026cda54afca9d541.

Issue #196, opened 2013-04-16
[Patch] s390 ctc patch


guyharris:
Converted from [SourceForge issue 573111](http://sourceforge.net/support/tracker.php?aid=573111), submitted by saturn_de

Patch for s390 ctc devices

guyharris:
> What ARPHRD value does the s390 ctc interface use?

`ARPHRD_SLIP`, if [`ctcm_main.c` in the CTC driver](https://github.com/torvalds/linux/blob/master/drivers/s390/net/ctcm_main.c) is to be believed; see `ctcm_dev_setup()`.

But where, in there, is an *Ethernet* header put at the beginning of the packet?  Are they put there, at the front of each network packet, by the transmitting host?  [The main header file for the CTCM driver](https://github.com/torvalds/linux/blob/master/drivers/s390/net/ctcm_main.h) defines a 6-byte link-layer header with 2 bytes of packet length, 2 bytes of packet type, and 2 unused bytes, but the only things that has in common with an Ethernet header are 1) it has a 2-byte field that can contain an Ethertype and 2) it has a length that's 2 bytes less than a multiple of 4.

If the packets don't begin with an Ethernet header - 6 bytes of destination MAC address, 6 bytes of source MAC address, and 2 bytes of length or Ethertype - `DLT_EN10MB` is the *wrong* DLT_ value for the interface.

PR #556, opened 2017-02-15
Patches that apply in openSUSE

pmonrealgonzalez:
In openSUSE we apply the following three patches in the case of PPP and the error handling are trivial. For the S390 there was an open issue #196 closed as wontfix but we still need to set the type to be ethernet there, thus providing patch that was kindly written by RH maintainers. Could you please review this issue?

### DIFF
--- a/pcap/bpf.h
+++ b/pcap/bpf.h
@@ -15,11 +15,7 @@
  * 2. Redistributions in binary form must reproduce the above copyright
  *    notice, this list of conditions and the following disclaimer in the
  *    documentation and/or other materials provided with the distribution.
- * 3. All advertising materials mentioning features or use of this software
- *    must display the following acknowledgement:
- *      This product includes software developed by the University of
- *      California, Berkeley and its contributors.
- * 4. Neither the name of the University nor the names of its contributors
+ * 3. Neither the name of the University nor the names of its contributors
  *    may be used to endorse or promote products derived from this software
  *    without specific prior written permission.
  *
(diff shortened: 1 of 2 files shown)


infrastation:
I am not competent to comment on the other points, but with regard to the `strncmp ("ctc"...` change it would seem reasonable to guard the hack with an `#ifdef` to compile it on S/390 only. Linux network interfaces can and do have surprising names, it is not impossible that some Linux sysadmins may use the `ctc` prefix not knowing it has a special meaning on S/390. In such case it would be wrong to apply the DLT override.

pmonrealgonzalez:
I just added the #ifdef to compile on S/390 as you suggested.

guyharris:
The only way in which these three changes are related to each other is that they're all in openSUSE.

Could you please submit them as separate changes?

guyharris:
> For the S390 there was an open issue #196 closed as wontfix but we still need to set the type to be ethernet there

*Why* do you need the type to be set to Ethernet?  Where, in [`ctcm_main.c` in the CTC driver](https://github.com/torvalds/linux/blob/master/drivers/s390/net/ctcm_main.c), is an *Ethernet* header put at the beginning of the packet?  Are they put there, at the front of each network packet, by the transmitting host?  [The main header file for the CTCM driver](https://github.com/torvalds/linux/blob/master/drivers/s390/net/ctcm_main.h) defines a 6-byte link-layer header with 2 bytes of packet length, 2 bytes of packet type, and 2 unused bytes, but the only things that has in common with an Ethernet header are 1) it has a 2-byte field that can contain an Ethertype and 2) it has a length that's 2 bytes less than a multiple of 4.

If the packets don't begin with an Ethernet header - 6 bytes of destination MAC address, 6 bytes of source MAC address, and 2 bytes of length or Ethertype - `DLT_EN10MB` is the *wrong* DLT_ value for the interface.

PR #557, opened 2017-02-16
Added DLT_PPP in function gen_inbound()

pmonrealgonzalez:
Patch applied to add DLT for PPP in function gen_inbound(). We have added the #ifdef as suggested in #556.

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -7549,6 +7549,7 @@ gen_inbound(compiler_state_t *cstate, int dir)
 	 */
 	switch (cstate->linktype) {
 	case DLT_SLIP:
+	case DLT_PPP:
 		b0 = gen_relation(cstate, BPF_JEQ,
 			  gen_load(cstate, Q_LINK, gen_loadi(cstate, 0), 1),
 			  gen_loadi(cstate, 0),


guyharris:
The first byte of a PPP packet isn't a direction indicator, it's 0xFF if PPP is in [RFC 1662 HDLC-like](https://tools.ietf.org/html/rfc1662) framing or the first octet of the protocol field if there's no HDLC-like framing.  Why would this test the direction of the packet?

pmonrealgonzalez:
Not sure about that. Could that be related to [1] and/or [2]?

[1] https://tools.ietf.org/pdf/rfc1144.pdf (Appendix B, page 43)
[2] http://www.tcpdump.org/linktypes.html says "PPP, as per RFC 1661 and RFC 1662; if the first 2 bytes are 0xff and 0x03, it's PPP in HDLC-like framing, with the PPP header following those two bytes, otherwise it's PPP without framing, and the packet begins with the PPP header. The data in the frame is not octet-stuffed or bit-stuffed."

guyharris:
Neither of those references say that there's a direction bit.

On Linux, libpcap does not ever supply "raw PPP" headers; it maps `ARPHDR_PPP` to `DLT_LINUX_SLL`, because, to quote the comment:

		/*
		 * Some PPP code in the kernel supplies no link-layer
		 * header whatsoever to PF_PACKET sockets; other PPP
		 * code supplies PPP link-layer headers ("syncppp.c");
		 * some PPP code might supply random link-layer
		 * headers (PPP over ISDN - there's code in Ethereal,
		 * for example, to cope with PPP-over-ISDN captures
		 * with which the Ethereal developers have had to cope,
		 * heuristically trying to determine which of the
		 * oddball link-layer headers particular packets have).
		 *
		 * As such, we just punt, and run all PPP interfaces
		 * in cooked mode, if we can; otherwise, we just treat
		 * it as DLT_RAW, for now - if somebody needs to capture,
		 * on a 2.0[.x] kernel, on PPP devices that supply a
		 * link-layer header, they'll have to add code here to
		 * map to the appropriate DLT_ type (possibly adding a
		 * new DLT_ type, if necessary).
		 */

so, unless SuSE have changed libpcap to provide raw PPP headers - with or without a direction indicator - programs capturing with libpcap won't provide `DLT_PPP` packets.

Perhaps there's some source of packets *other* than libpcap that writes out pcap or pcapng files with packets that have raw PPP headers and a direction indicator added to the raw PPP header.  If so, that source of packets should perhaps use `DLT_PPP_PPPD` rather than `DLT_PPP`, because we *already support testing the direction for `DLT_PPP_PPPD` - *no* code changes to libpcap would be required for that.

And, in fact, `DLT_PPP_PPPD` was requested by... Karsten Keil of SuSE.

The comment describing it is

/*
 * Another PPP variant as per request from Karsten Keil <kkeil@suse.de>.
 *
 * This is used in some OSes to allow a kernel socket filter to distinguish
 * between incoming and outgoing packets, on a socket intended to
 * supply pppd with outgoing packets so it can do dial-on-demand and
 * hangup-on-lack-of-demand; incoming packets are filtered out so they
 * don't cause pppd to hold the connection up (you don't want random
 * input packets such as port scans, packets from old lost connections,
 * etc. to force the connection to stay up).
 *
 * The first byte of the PPP header (0xff03) is modified to accomodate
 * the direction - 0x00 = IN, 0x01 = OUT.
 */

So, if something is generating packets in that format, and it's using `DLT_PPP`, please fix it to use `DLT_PPP_PPPD`.

PR #558, opened 2017-02-17
Apparently ctc interfaces on s390 has ethernet DLT

pmonrealgonzalez:
There was an open issue #196 closed as won't fix but we still need to set the type to be ethernet there, thus providing this patch in function activate_new() that was kindly written by RH maintainer @msekletar. Could you please review this issue?

This issue is a continuation of #556.

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -3510,6 +3510,12 @@ activate_new(pcap_t *handle)
 				handle->linktype = DLT_LINUX_SLL;
 		}
 
+#if defined(__s390__)
+		/* Hack to make things work on s390 ctc interfaces */
+		if (strncmp("ctc", device, 3) == 0)
+		    handle->linktype = DLT_EN10MB;
+#endif
+
 		handlep->ifindex = iface_get_id(sock_fd, device,
 		    handle->errbuf);
 		if (handlep->ifindex == -1) {


guyharris:
So what is the ARPHRD_ value that libpcap gets for the ctc interfaces?  It's presumably not ARPHRD_ETHER, otherwise it *would* return DLT_EN10MB.

guyharris:
> So what is the ARPHRD_ value that libpcap gets for the ctc interfaces?

`ARPHRD_SLIP`, if [`ctcm_main.c` in the CTC driver](https://github.com/torvalds/linux/blob/master/drivers/s390/net/ctcm_main.c) is to be believed; see `ctcm_dev_setup()`.

But where, in there, is an *Ethernet* header put at the beginning of the packet?  Are they put there, at the front of each network packet, by the transmitting host?  [The main header file for the CTCM driver](https://github.com/torvalds/linux/blob/master/drivers/s390/net/ctcm_main.h) defines a 6-byte link-layer header with 2 bytes of packet length, 2 bytes of packet type, and 2 unused bytes, but the only things that has in common with an Ethernet header are 1) it has a 2-byte field that can contain an Ethertype and 2) it has a length that's 2 bytes less than a multiple of 4.

If the packets don't begin with an Ethernet header - 6 bytes of destination MAC address, 6 bytes of source MAC address, and 2 bytes of length or Ethertype - `DLT_EN10MB` is the *wrong* DLT_ value for the interface.

mcr:
Still looking for confirmation of what the actual encapsulation is.  Please rebase and re-open.

PR #559, opened 2017-02-17
Error handling in function pcap_activate_linux()

pmonrealgonzalez:
Modification of the error handling in function pcap_activate_linux(). This follows from issue #556.

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -1548,16 +1548,10 @@ pcap_activate_linux(pcap_t *handle)
 		}
 	}
 	else if (ret == 0) {
-		/* Non-fatal error; try old way */
-		if ((ret = activate_old(handle)) != 1) {
-			/*
-			 * Both methods to open the packet socket failed.
-			 * Tidy up and report our failure (handle->errbuf
-			 * is expected to be set by the functions above).
-			 */
-			status = ret;
-			goto fail;
-		}
+	        snprintf(handle->errbuf, PCAP_ERRBUF_SIZE,
+			 "af_packet module unavailable, missing a reboot to new kernel?");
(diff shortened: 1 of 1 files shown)


guyharris:
So why is it better to completely remove support for PF_INET/SOCK_PACKET sockets than to try to use them if we get EINVAL or EAFNOSUPPORT from an attempt to create a PF_PACKET socket or if, in `if_bind()`, `getsockopt(fd, SOL_SOCKET, SO_ERROR, &err, &errlen)` fails or if `err` is an errno other than ENETDOWN?

pmonrealgonzalez:
This is to avoid fallback to the obsolete SOCK_PACKET (kernel < 2.2) interface. This still happens, for example, when you update the kernel, run iftop and the old kernel af_packet module wasn't loaded; the program fails and the kernel prints "iftop uses obsolete (PF_INET,SOCK_PACKET)".

guyharris:
> This is to avoid fallback to the obsolete SOCK_PACKET (kernel < 2.2) interface.

There's "fallback to `SOCK_PACKET` because the kernel is new enough to support `PF_PACKET` but the module hasn't been installed" and there's "fallback to `SOCK_PACKET` because the kernel isn't new enough to support `PF_PACKET`"; I see no need to gratuitously avoid the latter.

pmonrealgonzalez:
Thank you for your quick answer. According to man packet(7), `SOCK_PACKET` is deprecated and should not be used. IMHO, if you want to keep support for `SOCK_PACKET` it would be great to at least throw a warning.

mcr:
Would the original poster like to find a way to distinguish between the two cases that guy commented on?  If so, please re-open and rebase.

PR #563, opened 2017-02-28
libpcap: added raw filters

chemag:
The goal of supporting raw filters* is to provide libpcap/tcpdump support
for generic BPF insns, including those that are not-supported by libpcap
(e.g., the BPF_MOD/BPF_XOR ops in Linux, or any of the multiple ancillary
loads in linux). It also allows testing new kernel extensions to the
BPF ISA without having to modify libpcap/tcpdump.

We provide support by modifying pcap_compile() so that it first checks
for raw filters. This works for expressions appended in the command line,
and for expressions read from a file ("-F" option). Filters starting
with an integer and a valid separator (',' or '\n') are considered raw.
All other filters are considered (traditional) expressions.

We also make sure that filters compiled from raw filters are left for the
kernel to validate (added "skip_validate" to pcap_t).

*Raw filters are those generated by tcpdump -ddd. i.e.

$ ./tcpdump -ddd -i eth0 icmp
6
40 0 0 12
21 0 3 2048
48 0 0 23
21 0 1 1
6 0 0 65535
6 0 0 0

We also support replacing the new line characters with commas, as to
make possible to have inline filters.

$ ./tcpdump -ddd -i eth0 icmp |tr '\n' ','
6,40 0 0 12,21 0 3 2048,48 0 0 23,21 0 1 1,6 0 0 65535,6 0 0 0,

Some examples:
$ ./tcpdump -nn -i eth0 "6,40 0 0 12,21 0 3 2048,48 0 0 23,21 0 1 1,6 0 0 65535,6 0 0 0,"
$ ./tcpdump -nn -i eth0 -F ~/bpf/icmp.2.bpfraw


guyharris:
That's a bit of a big Git comment.  Should we discard everything starting with "*Raw filters are those generated by tcpdump -ddd. i.e.", and change the first paragraph to read

> The goal of supporting raw filters (those generated by tcpdump -ddd) ...

chemag:
I like large comments (it's a doc, after all), but please feel free to remove what you decide

gencode.c
```diff
@@ -50,6 +50,10 @@
 
 #endif /* _WIN32 */
 
+#if defined(BSD) || (defined (__APPLE__) && defined(__MACH__))
+#include <net/bpf.h>
```
guyharris: So why does this need to be included?  We're dealing with *raw* filters, i.e. raw BPF machine code, so there shouldn't be any need to look at the definitions for the kernel's BPF interpreter.

guyharris:
Is there any reason to run raw filters through the optimizer?  Presumably if a user specifies a raw filter, they want exactly that chunk of BPF machine code to be used.

guyharris:
Documentation for the user would belong in a man page; documentation for libpcap developers would belong in a comment in the code.

chemag:
> Documentation for the user would belong in a man page; documentation for libpcap developers would  belong in a comment in the code.

I removed all the test notes.

chemag:
> Is there any reason to run raw filters through the optimizer? Presumably if a user specifies a raw filter, they want exactly that chunk of BPF machine code to be used.

AFAICT, we don't run the optimizer when we call pcap_compile_raw(). In that case, pcap_compile() returns before calling the optimizer

gencode.c
chemag: (Leftover from the previous patch). 

Done

infrastation:
Please note this pull request failed to build on MacOS.

gencode.c
```diff
@@ -661,6 +661,103 @@ syntax(compiler_state_t *cstate)
...
+	char sp, separator, separator1 = ',', separator2 = '\n';
+	unsigned short bpf_len, i = 0, line = 0;
+	size_t size;
+	struct bpf_insn tmp;
+
+	program->bf_len = 0;
+
+	if (sscanf(bpf_string, "%hu%c", &bpf_len, &sp) != 2 ||
+			(sp != separator1 && sp != separator2) ||
+			bpf_len > BPF_MAXINSNS || bpf_len == 0) {
```
guyharris: That's not a syntax error.

In addition, it requires that `BPF_MAXINSNS` be available in a header file, which it might or might not be, and the header file depends on the OS.  Remember, this has to compile on, at minimum, various versions of:

-  various Linux distributions
- various BSDs, including Darwin
- Solaris
- AIX
- HP-UX

and the header file on the system on which it's compiled might, or might not, match the system on which it's running.

If we're going to impose a limit on the number of instructions (the main reason for which would be to keep from allocating a huge blob of memory and eating up most of the address space or backing store), if we're going to use an `unsigned short`, we might as well just impose a limit of 65535 instructions and let the kernel complain if it has a lower limit.

chemag: Your suggestion is ''s/BPF_MAXINSNS/65535/', right?

chemag: Note that my kernel (3.13.0) defines BPF_MAXINSNS to be 4096. I doubt there is an issue with moving from 4096 to 64k (raw filters are only used from tcpdump clients).

Another question: do you prefer me harcoding the 64k constant, or something like:

#ifndef BPF_MAXINSNS
#define BPF_MAXINSNS 65536
#endif

chemag: Did the ifndef code. Also used 4096 instead of 64k (that's what the kernel defines).

guyharris:
> Also used 4096 instead of 64k (that's what the kernel defines).

Actually, [the kernel defines it as 512](https://opensource.apple.com/source/xnu/xnu-3789.41.3/bsd/net/bpf.h.auto.html), as do [the kernel](https://svnweb.freebsd.org/base/head/sys/net/bpf.h?revision=314436&view=markup) and [the kernel](http://cvsweb.netbsd.org/bsdweb.cgi/src/sys/net/bpf.h?rev=1.69&content-type=text/x-cvsweb-markup&only_with_tag=MAIN) and [the kernel](http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/sys/net/bpf.h?rev=1.62&content-type=text/x-cvsweb-markup) and [the kernel](http://gitweb.dragonflybsd.org/dragonfly.git/blob_plain/HEAD:/sys/net/bpf.h).  [The kernel](https://github.com/illumos/illumos-gate/blob/master/usr/src/uts/common/io/bpf/net/bpf.hl) also appears to do so, although Oracle's version may have diverged from the last OpenSolaris version.

Unfortunately, [the kernel](http://www-03.ibm.com/systems/power/software/aix/) isn't open-source, so I can't post a URL.  [The kernel-mode driver](https://github.com/wireshark/winpcap/blob/master/packetNtx/driver/win_bpf.h) also defines it as 512.

Oh, and [the kernel](http://www8.hp.com/us/en/products/servers/hp-ux.html) doesn't have BPF, so it doesn't define it as anything.

Translation: there's no such thing as "the kernel" in the context of libpcap; there are a *number* of kernels it deals with - that's the whole *point* of libpcap; it hides, as best it can, the variety of packet capture mechanisms, so that code can be written to run on *several different OSes* with a minimum of platform-specific #ifdefs.

Note also that any one of those kernels might change the value in the future.

So *my* inclination is not to pay attention to what any particular kernel happens to choose, and not to make an effort to try to find the appropriate header on various different platforms.  Just pick an arbitrary maximum, and give it a name other than `BPF_MAXINSNS`, to emphasize that it's a *libpcap* limit rather than any particular OS's limit.

chemag:
> Actually, the kernel defines it as 512, as do the kernel and the kernel and the kernel and the kernel ...

My bad. There's only [one kernel](https://www.kernel.org/) for me :)

> So my inclination is not to pay attention to what any particular kernel happens to choose, and not to make an effort to try to find the appropriate header on various different platforms. Just pick an arbitrary maximum, and give it a name other than BPF_MAXINSNS, to emphasize that it's a libpcap limit rather than any particular OS's limit.

Done

chemag:
Ping

chemag:
Ping

chemag:
One more ping...

mcr:
if there is still interest, please rebase, thank you!

chemag:
> if there is still interest, please rebase, thank you!

Done. 

Tested again, and added a "Tested:" section to the patch comment.

Thanks!

chemag:
Also, per a previous comment from Guy, I moved a chunk of the comment to the `pcap_compile` man page

gencode.c
```diff
@@ -717,6 +722,103 @@ syntax(compiler_state_t *cstate)
...
+	unsigned short bpf_len, i = 0, line = 0;
+	size_t size;
+	struct bpf_insn tmp;
+
+	program->bf_len = 0;
+
+	if (sscanf(bpf_string, "%hu%c", &bpf_len, &sp) != 2 ||
+			(sp != separator1 && sp != separator2) ||
+			bpf_len > PCAP_MAXINSNS || bpf_len == 0) {
+		snprintf(p->errbuf, PCAP_ERRBUF_SIZE,
```
guyharris: Use `pcap_snprintf()` rather than `snprintf()`.

gencode.c
```diff
@@ -717,6 +722,103 @@ syntax(compiler_state_t *cstate)
...
+		return (-1);
+	}
+	separator = sp;
+
+	/* allocate the fcode buffer */
+	program->bf_len = bpf_len;
+	size = sizeof(struct bpf_insn) * program->bf_len;
+	program->bf_insns = (struct bpf_insn *)malloc(size);
+	if (program->bf_insns == NULL) {
+		snprintf(p->errbuf, PCAP_ERRBUF_SIZE,
```
guyharris: Use `pcap_snprintf()` rather than `snprintf()`.

gencode.c
```diff
@@ -717,6 +722,103 @@ syntax(compiler_state_t *cstate)
...
+			++token;
+		/* skip empty lines and comments */
+		if (*token == separator || *token == '\n' || strlen(token) < 1 ||
+				token[0] == '#')
+			continue;
+
+		if (i >= bpf_len) {
+			program->bf_len = 0;
+			free(program->bf_insns);
+			snprintf(p->errbuf, PCAP_ERRBUF_SIZE,
```
guyharris: Use `pcap_snprintf()` rather than `snprintf()`.

gencode.c
```diff
@@ -717,6 +722,103 @@ syntax(compiler_state_t *cstate)
...
+					"program exceeds encoded length!");
+			return (-1);
+		}
+
+		if (sscanf(token, "%hu %hhu %hhu %u%c",
+				&tmp.code, &tmp.jt, &tmp.jf, &tmp.k, &sp) != 5) {
+			program->bf_len = 0;
+			free(program->bf_insns);
+			snprintf(p->errbuf, PCAP_ERRBUF_SIZE,
```
guyharris: Use `pcap_snprintf()` rather than `snprintf()`.

gencode.c
```diff
@@ -717,6 +722,103 @@ syntax(compiler_state_t *cstate)
...
+		program->bf_insns[i].jf = tmp.jf;
+		program->bf_insns[i].k = tmp.k;
+
+		i++;
+	}
+
+	if (i != bpf_len) {
+		program->bf_len = 0;
+		free(program->bf_insns);
+		snprintf(p->errbuf, PCAP_ERRBUF_SIZE,
```
guyharris: Use `pcap_snprintf()` rather than `snprintf()`.

gencode.c
chemag: Done (`s/snprintf/pcap_snprintf/g`)

infrastation:
On one hand, merging these changes into libpcap would improve consistency in tcpdump: since it can print the compiled bytecode with `-ddd`, it would be reasonable if it could parse and use this compiled bytecode (as `iptables -I INPUT -m bpf --bytecode` does), and for that libpcap (but not the programs that use libpcap, as Guy explained) would have to recognize it.

On the other hand, the "ddd" format has a pitfall, in that it does not specify for which DLT the expression was compiled. This, for example, creates the space for things quietly going wrong when the end user compiles a filter using the popular `DLT_EN10MB` type, whereas in the iptables context above the bytecode is applied to what effectively is `DLT_RAW`.

So perhaps this change and cBPF savefile address related, but distinct use cases, and the only coordination between the two should be that the complete API in the end makes as much sense as possible.

PR #565, opened 2017-03-08
savefile errorbuff processing

bch:
Also set the user-supplied errorbuffer when failing buffering on a savefile.


guyharris:
The changes in 7d7f405d0200b583b6094be793c61c2b38d5c88d are already all in the master branch.

The change in 3046bd2ea35583b1dfd455808d1d3ed785d6b693 is unnecessary in the master branch; the "setnonblock" functions aren't passed a pointer to the user buffer, so they all fill in the `errbuf` member of the `pcap_t` passed to them, and the `pcap_setnonblock()` function copies `p->errbuf` to the `errbuf` argument, so the user-supplied error buffer is *always* set on an error.

bch:
re: 3046bd2ea35583b1dfd455808d1d3ed785d6b693  -- I see now that I was patching against a release, different than MASTER. My fault.

Cheers,

-bch

PR #567, opened 2017-03-13
Add .appveyor.yml for MSVC automated builds

gvanem:
After umpteen trials and errors, the AppVeyor finally produces a `wpcap.dll` and `wpcap.lib` (I assume that is an import-lib for the former). The build-log is [here](https://ci.appveyor.com/project/gvanem/libpcap).

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,21 @@
+branches:
+  only:
+  - master
+
+clone_depth: 5
+
+matrix:
+  fast_finish: true
+
+install:
+  - cinst winflexbison
+  - win_flex --version
+  - win_bison --version
+  - appveyor DownloadFile http://www.winpcap.org/install/bin/WpdPack_4_1_2.zip
+  - 7z x .\WpdPack_4_1_2.zip -oc:\projects\libpcap\Win32
+
+build_script:
+  - md build
+  - cd build
+  - cmake -DPACKET_DLL_DIR=c:\projects\libpcap\Win32\WpdPack -G"Visual Studio 12 2013" ..
+  - msbuild -nologo -p:Configuration=Release pcap.sln
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@ cmake_minimum_required( VERSION 2.8.8 )
 if (POLICY CMP0042)
     cmake_policy(SET CMP0042 OLD)
 endif()
- 
+
 project( pcap )
 #
 # Call the library "wpcap" on Windows, for backwards compatibility.
@@ -54,6 +54,9 @@ include_directories(
 )
 
 if( WIN32 )
+    #
+    # The 'Win32/WpdPack/' directory branch is for an AppVeyor build.
+    #
     if( NOT "${PACKET_DLL_DIR}" STREQUAL "" )
         include_directories("${PACKET_DLL_DIR}/Include")
         if( CMAKE_CL_64 )
@@ -65,6 +68,7 @@ if( WIN32 )
     include_directories(
         ../Common/
         Win32/Include
+        Win32/WpdPack/Include
     )
 endif( WIN32)
 
--- a/pcap.c
+++ b/pcap.c
@@ -1610,7 +1610,7 @@ pcap_open_live(const char *device, int snaplen, int promisc, int to_ms, char *er
 		    NULL, errbuf));
 	}
 	if (srctype == PCAP_SRC_FILE) {
-		snprintf(errbuf, PCAP_ERRBUF_SIZE, "unknown URL scheme \"file\"");
+		pcap_snprintf(errbuf, PCAP_ERRBUF_SIZE, "unknown URL scheme \"file\"");
 		return (NULL);
 	}
 	if (srctype == PCAP_SRC_IFLOCAL) {


infrastation:
Thank you very much! I am enabling the check for the project.

infrastation:
You can see 2 checks done for the latest commit.

gvanem:
Yes, fine. But i'd like to clean up the 
```
include_directories(
          ../Common/
          Win32/Include
          Win32/WpdPack/Include
```
part. But I gathered that for a non-AppVeyor build, it wouldn't hurt to have a non-existing 
`Win32/WpdPack/Include` directory in the search-path. But like I wrote earlier, I don't really speak cmake. I find it's syntax rather confusing. 

But, I'll also try to get `tests/CMakeLists.txt` working for AppVeyor and MSVC.

infrastation:
I don't use cmake, on this occasion I just tried it because it seemed to be available in the VM and it worked. Actually, the AppVeyor build could have two jobs in the matrix: one using cmake and another using msbuild. After all, if the input files for those systems are in the repository, they both are supposed to work.

gvanem:
Just to clarify; the `pcap.sln` + `pcap.vcxproj` used in AppVeyor, *are* built by cmake and needs the *WinPcap developer pack*. Or is cmake + NPcap possible?
The MSbuild files in `Win32/Prj`, are hand-made (updated) by Yang Luo. These needs the *NPcap sources*. [This](https://github.com/nmap/npcap/archive/v0.83.zip) is the latest.

It would be nice to get NPcap into the AppVeyor matrix too, Yang, are you listening?

infrastation:
@hsluoyz, please see above.

hsluoyz:
The current ``wpcap.sln + wpcap.vcxproj`` are hand-made and don't support CMake yet. The build process on Windows is mostly different from Linux systems. So I think using CMake won't gain much.

The build of ``wpcap.vcxproj`` relies on Npcap, for example the path to ``Packet.lib``. Although ``Packet.lib`` is also available in Npcap SDK. I didn't ever attempt to build libpcap with Npcap SDK. Building with Npcap source code would be better.

You can git clone Npcap sources in Appveyor then put libpcap source into Npcap source's ``wpcap\libpcap`` folder. It's a little strange here because that path already contains a copy of Libpcap source (as a submodule). You need to delete that first, except that you can choose to git clone Npcap without cloning the submodules. Npcap only has one submodule Libpcap. So this operation will just leave the Libpcap slot open for the source going to be tested.

gvanem:
> Although Packet.lib is also available in Npcap SDK.

Where is this SDK? Does it work for MinGW too?

guyharris:
Should there be a split between the kernel-mode part of {Win,N}Pcap, and its user-mode `Packet.dll` library, and libpcap, so that there's a software project that provides a driver+`Packet.dll`, with a `Packet.dll` SDK, and a separate libpcap project that, on Windows, uses that SDK?

The only problem there is that the `Packet.dll` interface has been advertised as *not* being a stable API, and making it a separate project might run the risk of making people treat it as an API and limiting what parts of it can be changed in the future.

hsluoyz:
> Where is this SDK? Does it work for MinGW too?

https://nmap.org/npcap/

> The only problem there is that the Packet.dll interface has been advertised as not being a stable API

This is correct. And there's no need to separate them to different projects. Because all of them are for Windows only, and irreplacable.  Many customized APIs have been modified or added to each layer. It should be viewed as a black box.

guyharris:
> And there's no need to separate them to different projects. Because all of them are for Windows only, and irreplacable.

To what does "them" refer?  It presumably includes the driver and `packet.dll`, but does it include the libpcap code?  If it includes the libpcap code, that code is *not* for Windows only; the only difference between libpcap-on-{macOS,FreeBSD,NetBSD,OpenBSD,DragonFly BSD,Solaris,HP-UX,AIX,Linux,etc.} and libpcap-on-Windows is that, on Windows, it can't just use a capture mechanism that's part of the OS, it needs the extra driver and library.

hsluoyz:
My mistake. Not all parts are for Windows only.

guyharris:
> Not all parts are for Windows only.

Exactly.  The problem is that, currently, NPcap has a copy of libpcap source, which may or may not be up to date; libpcap has had recently had significant changes to get the remote capture code working, and I have a number of additional changes I'm working on.  People might want a version of libpcap for Windows that incorporates those changes; to get that libpcap version based on NPcap, currently, I presume that somebody would have to pick up the NPcap source, replace the libpcap source in it with the current libpcap source, fix whatever project files need to be fixed, and build it

Having a project with the driver and `packet.dll`, which installed header files and `.lib` files as part of an SDK, would make it possible to build libpcap with those versions of the driver and `packet.dll`; however, it would, as I noted, mean that `packet.dll`'s interface would become something that behaves like an API, making it more difficult to change.  I'm not convinced that making it a stable interface like that would necessarily be a good idea.

hsluoyz:
> People might want a version of libpcap for Windows that incorporates those change

Npcap updates Libpcap source in a release manner. So you can release 1.8.2 when the rpcap is ready, and it will be integrated by Npcap.

> I presume that somebody would have to pick up the NPcap source, replace the libpcap source in it with the current libpcap source, fix whatever project files need to be fixed, and build it.

It's actually very easy for the developer. There can be no delete and copy. Just clone Npcap's source with Libpcap submodule. Then switch to the submodule as a Git repo and add the official remote to it. Use the submodule repo as the development environment. The umbrella Npcap repo can be used to update Npcap's source.

For Appveyor, the above process seems to be tedious. Not as fast as just copying the Libpcap source.

> I'm not convinced that making it a stable interface like that would necessarily be a good idea.

Another reason is that I didn't look after the Npcap SDK so much. Maybe sometimes the repo's code is changed, but the SDK is forgot to be updated.

PR #568, opened 2017-03-16
MSVC's strdup() in _DEBUG mode

gvanem:
In MSVC debug-mode (`-MDd` or `-MTd`), there is an important warning on
the redefinition of `strdup`:
```
  portability.h(104): warning C4005: 'strdup': macro redefinition
  f:\ProgramFiler-x86\WindowsKits\Include\10.0.10586.0\ucrt\crtdbg.h(319):
    note: see previous definition of 'strdup'
```
It's definition is:
```
  #define strdup(s)  _strdup_dbg(s, _NORMAL_BLOCK, __FILE__, __LINE__)
```
Undoing that in `"portability.h"` would defeat the purpose of `_DEBUG` mode.

### DIFF
--- a/pcap-win32.c
+++ b/pcap-win32.c
@@ -826,25 +826,15 @@ pcap_read_win32_dag(pcap_t *p, int cnt, pcap_handler callback, u_char *user)
 
 /* Send a packet to the network */
 static int
-pcap_inject_win32(pcap_t *p, const void *buf, size_t size){
-	LPPACKET PacketToSend;
+pcap_inject_win32(pcap_t *p, const void *buf, size_t size)
+{
+	PACKET pkt;
 
-	PacketToSend=PacketAllocatePacket();
-
-	if (PacketToSend == NULL)
-	{
-		pcap_snprintf(p->errbuf, PCAP_ERRBUF_SIZE, "send error: PacketAllocatePacket failed");
-		return (-1);
-	}
-
-	PacketInitPacket(PacketToSend, (PVOID)buf, (UINT)size);
-	if(PacketSendPacket(p->adapter,PacketToSend,TRUE) == FALSE){
+	PacketInitPacket(&pkt, buf, size);
+	if(PacketSendPacket(p->adapter,&pkt,TRUE) == FALSE) {
 		pcap_snprintf(p->errbuf, PCAP_ERRBUF_SIZE, "send error: PacketSendPacket failed");
-		PacketFreePacket(PacketToSend);
 		return (-1);
 	}
-
-	PacketFreePacket(PacketToSend);
 
 	/*
 	 * We assume it all got sent if "PacketSendPacket()" succeeded.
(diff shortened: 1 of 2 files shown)


infrastation:
This is how GitHub works: 1 pull request = 1 branch, even if you add more commits to the branch later. As a side note, instead of the merge commit you could have rebased on the master branch to make the history look a little bit simpler (not critical).

gvanem:
@infrastation rebased on the master branch ..

You mean, I should run `git rebase` locally here? Not sure what that does.

I'm really tempted to just delete my `gvanem/libpcap` repo after every PR I get merged into `the-tcpdump-group/libpcap`. Thus making git loose any *hidden* state it's keeping.

infrastation:
`git rebase master` will put your local changes in the current branch on top of your local master branch. The other thing is, the local master branch will need to have the changes that the-tcpdump-group has, please see the example commands in tcpdump's `CONTRIBUTING` file if you want to try it.

Alternatively, deleting the local repository and starting from scratch will achieve the same effect, I did this for quite a long time until I found the right documentation and learned how it works. It does not really matter which commands you use, it matters what gets recorded in the commit history.

Ah, and since recently there's a "rebase" pull request merge option on GitHub, so we can probably just try it and see if it does the job right.

guyharris:
> It's definition is

By "its definition" do you mean "*Microsoft's* definition"?

If any version of MSVC is defining `strdup`, and it's a definition that makes it work like UN\*X's `strdup()`, e.g.

    #define strdup _strdup

then, if we're compiling with that version of MSVC, we shouldn't be defining `strdup` *at all*, debug mode or no.

Microsoft's *documentation* of `strdup()`, for [VS 2005](https://msdn.microsoft.com/en-us/library/ms235454(v=vs.80).aspx) through [VS 2015](https://msdn.microsoft.com/en-us/library/ms235454(v=vs.140).aspx), says those functions are deprecated and that " the ISO C++ conformant _strdup ..." should be used instead.

Presumably they're deprecated to avoid namespace pollution, i.e. somebody might have written a program for C on Windows and had their own `strdup()` function, which might not have the same signature, and that should be allowed.  It's probably allowed on UN\*X if you set the right flags to keep non-ISO C functions out of the namespace.

But the question here is whether there are any versions of Visual Studio where `strdup()` is deprecated but is present in the namespace by default; if so, we should avoid defining `strdup` on those versions of VS.

gvanem:
> But the question here is whether there are any versions of Visual Studio where strdup() is 
> deprecated but is present in the namespace by default; if so, we should avoid defining strdup 
> on those versions of VS.

That would only add to the `ifdef`-mess. But it's not really important for me to enable debug-mode in Wpcap. My intention with it was to find memory leaks. But I failed to find any! 

And `strdup` etc. exists in *all* versions of MSVC (in `oldnames.lib`). The annoying POSIX warning (since MSVC 2012?), can be disabled using `-D _CRT_NONSTDC_NO_WARNINGS --D_CRT_OBSOLETE_NO_WARNINGS`. No big deal.

So that leaves the patch for `PacketSetDumpLimits()` as more important IMHO... before I update the `msdos/*` stuff.

PR #576, opened 2017-05-09
bound snaplen for linux tpacket_v2 to ~64k

altoor:
The aggregated ethernet packets produced by the linux kernel
can't exceed 64K (plus eth overhead), so we can bound accordingly
the snaplen for tpacket_v2 on such devices.

This allows a more efficient usage of the receive buffer;

tcpdump has increased the default snaplen up to 256K in recent
versions, to cope better with USB data, and that leads to
tpacket_v2 rx ring being configured by default with tp_frame_nr
equal to 7, which is very bad for most use-case.

With this change the default tpacket_v2 layout has tp_frame_nr
equal to 32, which is still quite low, but works much better
and is consistent with the previous tcpdump behavior.

Signed-off-by: Paolo Abeni <pabeni@redhat.com>

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -3999,6 +3999,8 @@ prepare_tpacket_socket(pcap_t *handle)
 	return 1;
 }
 
+#define MAX(a,b) ((a)>(b)?(a):(b))
+
 /*
  * Attempt to set up memory-mapped access.
  *
@@ -4042,10 +4044,10 @@ create_ring(pcap_t *handle, int *status)
 #ifdef HAVE_TPACKET2
 	case TPACKET_V2:
 #endif
-		/* Note that with large snapshot length (say 64K, which is
-		 * the default for recent versions of tcpdump, the value that
-		 * "-s 0" has given for a long time with tcpdump, and the
-		 * default in Wireshark/TShark/dumpcap), if we use the snapshot
+		/* Note that with large snapshot length (say 256K, which is
+		 * the default for recent versions of tcpdump, Wireshark,
+		 * TShark, dumpcap or 64K, the value that "-s 0" has given for
+		 * a long time with tcpdump), if we use the snapshot
 		 * length to calculate the frame length, only a few frames
 		 * will be available in the ring even with pretty
 		 * large ring size (and a lot of memory will be unused).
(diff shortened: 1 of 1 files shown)


altoor:
Hi,

On Wed, May 10, 2017 at 3:01 AM, Guy Harris <notifications@github.com> wrote:
> @guyharris commented on this pull request.
>
> Is there a kernel API header that defines this limit? If so, we should probably use that.

No, there is no define for that. The limit is enforced for gro/gso in
skb_gro_receive(), file net/core/skbuff.c:

        if (unlikely(p->len + len >= 65536))
                return -E2BIG;

and is also due to the fact that gro/lro packets must be valid ip
packets, with total_len fitting the header field limit.

> ________________________________
>
> In pcap-linux.c:
>
> >   offload = iface_get_offload(handle);
>   if (offload == -1) {
>   *status = PCAP_ERROR;
>   return -1;
>   }
> - if (!offload) {
> - mtu = iface_get_mtu(handle->fd, handle->opt.device,
> -    handle->errbuf);
> - if (mtu == -1) {
> - *status = PCAP_ERROR;
> - return -1;
> - }
> - if (frame_size > (unsigned int)mtu + 18)
> - frame_size = (unsigned int)mtu + 18;
> - }
> + if (!offload)
> + max_frame_len = mtu;
> + else
> + max_frame_len = MAX(mtu, 65535);
>
> "64K" is 65536, not 65535. Do you really mean 65535 here? Note: "we used to use 65535 for "-s 0"" doesn't mean we should use it here; there's no good reason why 65535 rather than 65536 should have been used for "-s 0".

That was inaccuracy on my side, the real bound is 65535 (included), as
explained above.

I'll fix the commend and send a new pull request if you don't have objections.

Cheers,

Paolo

pcap-linux.c
```diff
@@ -4042,9 +4044,9 @@ create_ring(pcap_t *handle, int *status)
 #ifdef HAVE_TPACKET2
 	case TPACKET_V2:
 #endif
-		/* Note that with large snapshot length (say 64K, which is
-		 * the default for recent versions of tcpdump, the value that
-		 * "-s 0" has given for a long time with tcpdump, and the
+		/* Note that with large snapshot length (say 256K, which is
```
guyharris: 256K is also the current default for Wireshark/TShark/dumpcap, so the comment should reflect that.

pcap-linux.c
```diff
@@ -4069,29 +4071,34 @@ create_ring(pcap_t *handle, int *status)
...
-				mtu = iface_get_mtu(handle->fd, handle->opt.device,
-				    handle->errbuf);
-				if (mtu == -1) {
-					*status = PCAP_ERROR;
-					return -1;
-				}
-				if (frame_size > (unsigned int)mtu + 18)
-					frame_size = (unsigned int)mtu + 18;
-			}
+			if (!offload)
```
guyharris: It might read better as

	 if (offload)
	 	 max_frame_len = MAX(mtu, 65535);
	 else
	 	 max_frame_len = mtu;

altoor:
Hi Guy,

did you had any time to have a look at the later version of the patch ? I think/hope I've integrated all your suggestions.

Thank you,

Paolo

guyharris:
> did you had any time to have a look at the later version of the patch ?

I didn't even know there *was* a later version; GitHub didn't bother to send me an email when you updated the patch.

msekletar:
> I didn't even know there was a later version; GitHub didn't bother to send me an email when you updated the patch.

Yeah, Github doesn't do that. I figure this should be probably mentioned somewhere in our README that people should always comment on the PR whenever they force push their feature branch.

PR #578, opened 2017-05-10
Create plugin API for read/write file I/O, e.g. to allow compression

raybellis:
This patch adds automatic on-the-fly decompression of files with names ending '.gz' on supported systems when those filenames are passed to `pcap_open_offline()` family functions that take a filename.

It uses `fopencookie` on glibc systems, and `funopen` on BSD.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1869,13 +1869,8 @@ if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/.devel OR EXISTS ${CMAKE_BINARY_DIR}/.deve
         #
         # We do *not* care about every single place the compiler would
         # have inserted Spectre mitigation if only we had told it to
-        # do so with /Qspectre.  I guess the theory is that it's seeing
-        # bounds checks that would prevent out-of-bounds loads and that
-        # those out-of-bounds loads could be done speculatively and that
-        # the Spectre attack could detect the value of the out-of-bounds
-        # data *if* it's within our address space, but unless I'm
-        # missing something I don't see that as being any form of
-        # security hole.
+        # do so with /Qspectre.  Maybe it's worth it, as that's in
+        # Bison-generated code that we don't control.
         #
         # XXX - add /Qspectre if that is really worth doing.
         #
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1866,6 +1866,20 @@ if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/.devel OR EXISTS ${CMAKE_BINARY_DIR}/.deve
         # structure members.
         #
         check_and_add_compiler_option(-wd4820)
+        #
+        # We do *not* care about every single place the compiler would
+        # have inserted Spectre mitigation if only we had told it to
+        # do so with /Qspectre.  I guess the theory is that it's seeing
+        # bounds checks that would prevent out-of-bounds loads and that
+        # those out-of-bounds loads could be done speculatively and that
+        # the Spectre attack could detect the value of the out-of-bounds
+        # data *if* it's within our address space, but unless I'm
+        # missing something I don't see that as being any form of
+        # security hole.
+        #
+        # XXX - add /Qspectre if that is really worth doing.
+        #
+        check_and_add_compiler_option(-wd5045)
     else()
         #
         # Other compilers, including MSVC with a Clang front end and
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,7 +87,7 @@ endif()
 # We assume that UTF-8 source is OK with other compilers.
 #
 if(MSVC AND NOT ${CMAKE_C_COMPILER} MATCHES "clang*")
-    set(C_ADDITIONAL_FLAGS "${C_ADDITIONAL_FLAGS} -utf-8")
+    set(C_ADDITIONAL_FLAGS "${C_ADDITIONAL_FLAGS} /utf-8")
 endif(MSVC AND NOT ${CMAKE_C_COMPILER} MATCHES "clang*")
 
 #
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -955,7 +955,7 @@ set(PROJECT_SOURCE_LIST_C
     optimize.c
     pcap-common.c
     pcap.c
-	pcap-ioplugin.c
+    pcap-ioplugin.c
     savefile.c
     sf-pcapng.c
     sf-pcap.c
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -955,6 +955,7 @@ set(PROJECT_SOURCE_LIST_C
     optimize.c
     pcap-common.c
     pcap.c
+	pcap-ioplugin.c
     savefile.c
     sf-pcapng.c
     sf-pcap.c
--- a/pcap-ioplugin.c
+++ b/pcap-ioplugin.c
@@ -78,7 +78,8 @@
  * fopen's safe version on Windows.
  */
 #ifdef _MSC_VER
-FILE *fopen_safe(const char *filename, const char* mode)
+FILE *
+fopen_safe(const char *filename, const char* mode)
 {
 	FILE *fp = NULL;
 	errno_t errno;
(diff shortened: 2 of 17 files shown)


raybellis:
Support for .gz handling on writes to follow, since those don't work nicely in conjunction with file rotation.    Having gzip invoked with the post-rotation processing option is very disk I/O intensive compared with just writing out data in gzip format from the outset.

mcr:
Ray Bellis <notifications@github.com> wrote:
    > Support for .gz handling on writes to follow, since those don't work
    > nicely in conjunction with file rotation. Having gzip invoked with the
    > post-rotation processing option is very disk I/O intensive compared
    > with just writing out data in gzip format from the outset.

I have some concern with introducing a new libz.so or something dependancy to
libpcap.   I wonder if a popen("|gzip >") will work?

--

raybellis:
On 11/05/2017 15:32, Michael Richardson wrote:

> I have some concern with introducing a new libz.so or something
> dependancy to
> libpcap. I wonder if a popen("|gzip >") will work?

If I've done it right (although I might not have) it'll just fail to not
supporting it if libz.so isn't available at compile time and it'll be
treated as an unrecognized file format.

In any event, if libz.so isn't available it's pretty unlikely that the
gzip binary will be either (unless it's a bizarre system with a static
only binary)

mcr:
Ray Bellis <notifications@github.com> wrote:
    > If I've done it right (although I might not have) it'll just fail to
    > not
    > supporting it if libz.so isn't available at compile time and it'll be
    > treated as an unrecognized file format.

    > In any event, if libz.so isn't available it's pretty unlikely that the
    > gzip binary will be either (unless it's a bizarre system with a static
    > only binary)

It's an issue that distro maintainers will need to think about.
Adding the dependancy pulls in that library regardless of whether or not
it will be used.   libz.so is probably pretty ubiquitous, so probably I'm
just grasping at straws, but already have a bunch of annoying dependancies
for libpcap...

--

raybellis:
Another factor is that the output side compression I plan to work on
would probably be *way* harder using `popen` compared to using the gzip
library, particular where file rotation is in play.

raybellis:
I've changed the patch so that on compatible UNIX systems it'll dynamically load libz.so and use it, rather than being linked at compile time.

The only part I'm not sure about is whether it's completely safe to just pass the function pointer to `gzread` directly to `fopencookie` since the type signature for the third parameter isn't necessarily identical (`int` vs `size_t`)

raybellis:
I'm closing this because I'm working on a completely revised version that uses pluggable .so's to support arbitrary read (and later, write) handlers, rather than being gzip specific.

This would allow any compression method with suitable hook functions to be used, without creating a packaging-time dependency within libpcap itself.

raybellis:
Reopening now that I've done the major implementation (albeit lacking `pdump_open_append` support)

raybellis:
sample plugin (for gzip) at https://github.com/raybellis/libpcap-gzip/

mcr:
can you rebase?

raybellis:
Sure, if you're happy with the approach! :)

raybellis:
I've rebased, although the windows CI build is failing in places (mostly in the fuzz tests) because somewhere I've failed to include a reference to the new `pcap-ioplugin.c` file.

I didn't yet attempt to build my separate gzip plugin against this revised branch.

raybellis:
`pcap-ioplugin.c` is now recognised by the Windows build tools, but unfortunately the `sys/queue.h` header doesn't exist.   Investigation continues.

mcr:
Ray Bellis <notifications@github.com> wrote:
    > unfortunately the sys/queue.h header doesn't exist. Investigation
    > continues.

sys/queue.h is not POSIX, often does not exist on windows builds.
Bring your own copy.

raybellis:
I'm already working on it! :P

CMakeLists.txt
```diff
@@ -955,6 +955,7 @@ set(PROJECT_SOURCE_LIST_C
     optimize.c
     pcap-common.c
     pcap.c
+	pcap-ioplugin.c
```
guyharris: Wrong indentation.

pcap-ioplugin.c
```diff
@@ -0,0 +1,236 @@
...
+#endif
+#include <sys/types.h>
+#endif /* _WIN32 */
+
+#include <errno.h>
+#include <memory.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/queue.h>
```
guyharris: Not all systems on which libpcap works have `sys/queue.h` - that's why the build is failing on Windows, for example.

raybellis: Yeah, I know that now.   I'm working on replacing `<sys/queue.h>` with a simple in-place singly-linked list implementation.

raybellis: resolved in 6ca1b7f0f076fb5eb6a48e6e0806793c3fbb6c8d

pcap-ioplugin.c
```diff
@@ -0,0 +1,261 @@
...
+};
+
+static struct {
+	struct file_entry		*head;
+} file_list = { NULL };
+
+static int registered_atexit = 0;
+static int running_atexit = 0;
+
+static void pcap_ioplugin_closeall()
```
guyharris: `static void pcap_ioplugin_closeall(void)` - C is still keeping around backwards compatibility with pre-prototype versions, so you have to put `void` to explicitly indicate that it takes no arguments.

raybellis: gah, too used to forgiving compilers! :P .  Resolved in 841b1b3, along with some other function declaration style changes to closer match the project style.

raybellis:
No changes - just a rebase on current master.

fxlb:
"This branch cannot be rebased due to conflicts
Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch. "

fxlb:
(github message)
github1

fxlb:
Trying to rebase your PR gives:
```
$ git rebase master 
First, rewinding head to replay your work on top of it...
Applying: auto ungzip files named *.gz
Applying: simplified gzip reading to allow auto-detection and on stdin
Applying: use dlopen to load zlib at runtime
Applying: fix for when dlopen isn't available
Applying: ensure gzip functions actually loaded
Applying: moved compress to I/O plugin architecture
Applying: moved plugin and stdio function to own file
Applying: added I/O plugin file save support
Applying: clean up written files on exit
Applying: removed debug dlerror call
Applying: add pcap-ioplugin to win32 project
Applying: add comments re: binary mode from master
Applying: remove include missed from rebase
Applying: correct the order of macros in config.h.in from rebase
Applying: add pcap-ioplugin.c to CMakeLists
Applying: fix indentation
Applying: removed use of <sys/queue.h>
Applying: style and void prototype
Applying: remove fopen_safe() - no longer present in master
Applying: auto ungzip files named *.gz
Using index info to reconstruct a base tree...
M       config.h.in
M       configure
M       configure.ac
M       savefile.c
Falling back to patching base and 3-way merge...
Auto-merging savefile.c
CONFLICT (content): Merge conflict in savefile.c
Auto-merging configure.ac
CONFLICT (content): Merge conflict in configure.ac
Auto-merging configure
CONFLICT (content): Merge conflict in configure
Auto-merging config.h.in
error: Failed to merge in the changes.
Patch failed at 0020 auto ungzip files named *.gz
hint: Use 'git am --show-current-patch' to see the failed patch

Resolve all conflicts manually, mark them as resolved with
"git add/rm <conflicted_files>", then run "git rebase --continue".
You can instead skip this commit: run "git rebase --skip".
To abort and get back to the state before "git rebase", run "git rebase --abort".
```

raybellis:
looks like I've messed up, although I'm not sure how given that when I updated the PR it reported no conflicts?

fxlb:
In your rebase you should not have commit like ` Merge branch 'raybellis-gzip' of github.com:raybellis/libpcap into raybellis-gzip' `.

raybellis:
Yeah, I saw that, but thought I had cleaned it up.   I think the best option is for me to a new clean branch (and PR) based off current master?

raybellis:
See #823 instead.

Issue #581, opened 2017-05-16
optimizer discards setting values that are used


mkubecek:
Note: this problem was encountered with BPF program generated by patched version of libpcap so that it's possible the BPF program is malformed in some way I missed. If this is the case, I would appreciate some pointer to what I did wrong.

While trying to improve the `vlan` filter to correctly handle both packets with vlan info passed in metadata and packets with vlan info passed in packet itself, I ran into an issue with libpcap optimizer. This generated BPF program (for filter `vlan and vlan`)

```
 -- (000) ld       #0x0
 -- (001) st       M[0]
 -- (002) st       M[1]
 -- (003) ldb      [-4048]
[00](004) jeq      #0x1             jt 17       jf 5
 -- (005) ld       M[0]
 -- (006) add      #4
 -- (007) st       M[0]
 -- (008) ld       M[1]
 -- (009) add      #4
 -- (010) st       M[1]
 -- (011) ldh      [12]
[00](012) jeq      #0x8100          jt 17       jf 13
 -- (013) ldh      [12]
[00](014) jeq      #0x88a8          jt 17       jf 15
 -- (015) ldh      [12]
[00](016) jeq      #0x9100          jt 17       jf 27
 -- (017) ldx      M[1]
 -- (018) ldh      [x + 12]
[00](019) jeq      #0x8100          jt 26       jf 20
 -- (020) ldx      M[1]
 -- (021) ldh      [x + 12]
[00](022) jeq      #0x88a8          jt 26       jf 23
 -- (023) ldx      M[1]
 -- (024) ldh      [x + 12]
[00](025) jeq      #0x9100          jt 26       jf 27
[00](026) ret      #262144
[00](027) ret      #0
```

is optimized to

```
 -- (000) ldb      [-4048]
[00](001) jeq      #0x1             jt 6        jf 2
 -- (002) ldh      [12]
[06](003) jeq      #0x8100          jt 6        jf 4
[07](004) jeq      #0x88a8          jt 6        jf 5
[08](005) jeq      #0x9100          jt 6        jf 11
 -- (006) ldh      [x + 12]
[01](007) jeq      #0x8100          jt 10       jf 8
[03](008) jeq      #0x88a8          jt 10       jf 9
[04](009) jeq      #0x9100          jt 10       jf 11
[02](010) ret      #262144
[05](011) ret      #0
```

[...]

PR #582, opened 2017-05-17
two optimizer fixes

mkubecek:
Two optimizer fixes:

- fix dot graph debugging dump to show graph edges
- fix incorrect optimization (issue #581)


guyharris:
Could you submit this as two separate pull requests?  The only connection between them is that they both modify optimizer code.

mkubecek:
OK, created PR #613 and #614

PR #583, opened 2017-05-17
Make VLAN filter handle both metadata and inline tags

mkubecek:
Note: for this to work correctly with BPF optimization enabled, the fix for issue #581 (included in pull request #582) is needed.

Commit 04660eb1e561 ("Use BPF extensions in compiled filters") taught
libpcap to use VLAN BPF extensions to implement "vlan" and "vlan <id>"
filters. While this fixed filtering in the case when VLAN information
is in packet metadata (which is more frequent on Linux with recent
kernels), it broke the case when outermost VLAN tag is still in packet
data. Unfortunately, even if kernel supports BPF VLAN extensions, we have
to expect both types of packets.

This means that if the initial SK_AD_VLAN_TAG_PRESENT test is negative, we
need to also check for inline VLAN tag. As this test is done only for some
packets (determined at run time), we need to update variable part of link
type and link payload offsets, not constant part.

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -358,6 +358,11 @@ struct _compiler_state {
 	int is_geneve;
 
 	/*
+	 * TRUE if we need variable length part of VLAN offset
+	 */
+	int is_vlan_vloffset;
+
+	/*
 	 * These are offsets for the ATM pseudo-header.
 	 */
 	u_int off_vpi;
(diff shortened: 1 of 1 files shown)


mkubecek:
Hm... looks like MSVC doesn't like the empty initializer.

I'll wait for comments and then decide between fixing this by a follow-up commit and replacing the whole pull request.

infrastation:
If you want you can force-push to existing branch and GitHub will replace the contents of the pull request.

mkubecek:
Done, thanks for the tip.

guyharris:
> Done, thanks for the tip.

So presumably "done" includes removing or changing the empty initializer?  I didn't see any empty initializer in the change?

mkubecek:
Yes, I replaced

```
    struct s = { };
```

in `gen_vlan_no_bpf_extensions()` by

```
    struct s;
    s.next = NULL;
```

PR #591, opened 2017-06-13
updated VERSION

viclew1:


guyharris:
The master branch isn't 1.8.1, it's a non-yet-finalized-and-released 1.9.0, so the change isn't correct.

PR #595, opened 2017-07-07
Add pcap_set_protocol() to allow specifying a specific capture protocol.

buytenh:
This adds pcap_set_protocol(), which allows you to specify a specific
protocol to be captured.  On Linux, a protocol value specified with
this function will be used instead of ETH_P_ALL when creating and
binding a packet socket.

For ethernet packets, a similar result can be achieved by explicitly
creating a BPF filter that filters on ethertype and attaching that
filter to the packet socket, however, filtering using the packet socket
protocol field is slightly more efficient.

Also, in some cases, the packet socket protocol field and the
ethertype field have different values, and in those cases it is
impossible to achieve the same result by using a BPF filter.  An
example of this is on Arista platforms, where sFlow-sampled packets
are delivered, unmodified, to virtual interfaces on the management
CPU with their packet socket protocol value set to the magic value
0x002d, independently of their actual ethertype.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -231,6 +231,7 @@ MAN3PCAP_NOEXPAND = \
 	pcap_set_datalink.3pcap \
 	pcap_set_immediate_mode.3pcap \
 	pcap_set_promisc.3pcap \
+	pcap_set_protocol.3pcap \
 	pcap_set_rfmon.3pcap \
 	pcap_set_snaplen.3pcap \
 	pcap_set_timeout.3pcap \
--- a/pcap-int.h
+++ b/pcap-int.h
@@ -112,6 +112,9 @@ struct pcap_opt {
 	int	timeout;	/* timeout for buffering */
 	u_int	buffer_size;
 	int	promisc;
+#ifdef __linux__
+	int	protocol;
+#endif
 	int	rfmon;		/* monitor mode */
 	int	immediate;	/* immediate mode - deliver packets as soon as they arrive */
 	int	nonblock;	/* non-blocking mode - don't wait for packets to be delivered, return "no packets available" */
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -425,7 +425,7 @@ static int	iface_get_id(int fd, const char *device, char *ebuf);
 static int	iface_get_mtu(int fd, const char *device, char *ebuf);
 static int 	iface_get_arptype(int fd, const char *device, char *ebuf);
 #ifdef HAVE_PF_PACKET_SOCKETS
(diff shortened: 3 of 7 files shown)


guyharris:
> On Linux, a protocol value specified with this function will be used instead of ETH_P_ALL when creating and binding a packet socket.

And on everything else?

buytenh:
> And on everything else?

Good point.  I guess we would either have to fail the call to pcap_set_protocol(), or just not provide the function at all.  Any ideas/preferences as to what would be best?

guyharris:
Not provide the function at all, so that cross-platform programs that use it (tcpdump, Wireshark, etc.) will either have to avoid using it or have to avoid using it on anything other than Linux (so that the author will have to think carefully about how to use it).

buytenh:
Thanks for the feedback, and sorry for the delay.  I updated the commit to only provide the function on Linux.  How does this look?

infrastation:
Looks like now it builds only on Linux (see the log file in Travis build).

guyharris:
> Also, in some cases, the packet socket protocol field and the ethertype field have different values, and in those cases it is impossible to achieve the same result by using a BPF filter. An example of this is on Arista platforms, where sFlow-sampled packets are delivered, unmodified, to virtual interfaces on the management CPU with their packet socket protocol value set to the magic value 0x002d, independently of their actual ethertype.

I.e., if you want to see only those sFlow-sampled packets, you can't do it with a BPF filter, because they're not marked in the packet data as such?

buytenh:
> Looks like now it builds only on Linux (see the log file in Travis build).

Ooops, a stupid thinko.  Sorry about that, this is now fixed.

buytenh:
> I.e., if you want to see only those sFlow-sampled packets, you can't do it with a BPF filter, because they're not marked in the packet data as such?

Indeed! :)

infrastation:
Having looked through the comments -- but not the code -- something like `ether sockproto XXXX` (in addition to the existing `ether proto XXXX`) came to my mind, not sure if that would be a better approach. On one hand with a BPF primitive one can filter after the capture if the DLT saves the socket protocol value for the packet. On the other would the additional burden be worth the gain?

buytenh:
There isn't actually a way to retrieve the `ether sockproto` value for a captured packet, sadly -- you can tell the kernel to only give you packets with `sockproto X`, by passing `X` as the third argument to your `socket(PF_PACKET, ...` call, but once you have received a packet from your packet socket, there's no way to tell what `X` was.

infrastation:
Thank you for the explanation, I take it that this is not like packet direction, where there is `tcpdump -Q in|out|inout` and `inbound/outbound` primitives.

PR #598, opened 2017-08-18
Libpcap 1.8 fix HAVE_REMOTE on Unix

feld:
The libpcap 1.8 branch doesn't ensure HAVE_REMOTE is only honored by WIN32 systems. Currently software that tries to support HAVE_REMOTE for WIN32 systems fail to build on Unix because the remote-ext.h header is missing.


guyharris:
This is not necessary on the master branch and doesn't work on the master branch.

I've checked it into the 1.8 branch, the official home of which is on the bpf.tcpdump.org repository.

guyharris:
If you want rpcap support on UN*X, the master branch works a *lot* better.

PR #603, opened 2017-09-03
Fix a couple of build issues under Windows

blazeable:
Fix two types of warnings when building with Visual Studio 2015 (debug x64):
- strdup() undefined
- stdint.h macros redefined in scanner.c


CMakeLists.txt
```diff
@@ -200,6 +198,7 @@ set(PROJECT_SOURCE_LIST_C
 )
 
 if( WIN32 )
+    set(PROJECT_SOURCE_LIST_C ${PROJECT_SOURCE_LIST_C} pcap-win32.c )
```
guyharris: That's what the

    set(PROJECT_SOURCE_LIST_C ${PROJECT_SOURCE_LIST_C} pcap-${PCAP_TYPE}.c)

does after the code to determine PCAP_TYPE.  Why is this necessary?

guyharris:
So what is the point of renaming pcap-win32.c to pcap-npf.c?

nametoaddr.c
```diff
@@ -102,6 +102,10 @@ struct rtentry;		/* declarations in <net/if.h> */
 #include <pcap/namedb.h>
 #include "nametoaddr.h"
 
+#if defined(_MSC_VER) && !defined(strdup)
```
guyharris: `portability.h` should take care of this.  Why is this necessary?

pcap-new.c
```diff
@@ -44,6 +44,10 @@
 #include <stdlib.h>		// for malloc(), free(), ...
 #include <string.h>		// for strstr, etc
 
+#if defined(_MSC_VER) && !defined(strdup)
```
guyharris: Again, `portability.h` should handle this; why is this necessary?

pcap-rpcap.c
```diff
@@ -45,6 +45,10 @@
 #include "pcap-rpcap.h"
 #include "pcap-rpcap-int.h"
 
+#if defined(_MSC_VER) && !defined(strdup)
```
guyharris: And, again, `portability.h` should handle this.

pcap.c
```diff
@@ -78,6 +78,10 @@ struct rtentry;		/* declarations in <net/if.h> */
 #define INT_MAX		2147483647
 #endif
 
+#if defined(_MSC_VER) && !defined(strdup)
```
guyharris: And again...

sockutils.c
```diff
@@ -65,6 +65,10 @@
 #include "portability.h"
 #include "sockutils.h"
 
+#if defined(_MSC_VER) && !defined(strdup)
```
guyharris: And again - and `portability.h` is directly included *right above this*.

blazeable:
Currently the pcap-win32.c file contains the Packet32 driver implementation as well as some functions which should be generic/driver-independent for the Windows platform (DllMain entry point, WinSock library initialization). The idea is to keep the Win32-generic code in pcap-win32.c, and split the Packet32 driver implementation into pcap-npf.c.

nametoaddr.c
blazeable: It should, but the change in commit 9268e8661ffa4fe25bb742699127c711cb6e9e23 leaves a function undefined warning in debug mode (ie. when _DEBUG is defined), which this change corrects. Without it:
`C:\Users\blazeable\GitHub\libpcap\nametoaddr.c(265): warning C4013: 'strdup' undefined; assuming extern returning int`

pcap-new.c
blazeable: Same reason: corrects the warning.

pcap-rpcap.c
blazeable: Corrects the warning.

pcap.c
blazeable: Corrects the warning.

sockutils.c
blazeable: Yes, and yet, without it:
`4>C:\Users\blazeable\GitHub\libpcap\sockutils.c(889): warning C4013: 'strdup' undefined; assuming extern returning int`

CMakeLists.txt
blazeable: See comment added to the pull request itself. pcap-win32.c would contain the DllMain entry point, and so should then be compiled systematically on Windows, whereas capture driver implementations would be compiled depending on which Cmake option is enabled or which library was detected.

nametoaddr.c
guyharris: So that seems to imply that either:

1. `crtdbg.h` is only being included in some of the files that include `portability.h` and use `strdup()`, so, with `_DEBUG` defined, some files aren't getting `strdup` defined by `crtdbg.h`;

1. some versions of MSVC don't have the definition of `strdup(s)` as a call to `_strdup_dbg()`.

In the first case, presumably they should be including `crtdbg.h`, so that the debugging version of `strdup()` is used everywhere.

In the second case, presumably we should do something such as

    #ifndef strdup
        #define strdup  _strdup
    #endif

inside `#ifdef _MSC_VER`.  @gvanem, what do you think here?

pcap-new.c
guyharris: Again, why must this be handled in several places in the source rather than by a change to `portability.h`?

pcap-rpcap.c
guyharris: Again, why must this be handled in several places in the source rather than by a change to `portability.h`?

pcap.c
guyharris: Again, why must this be handled in several places in the source rather than by a change to `portability.h`?

sockutils.c
guyharris: Again, why must this be handled in several places in the source rather than by a change to `portability.h`?

guyharris:
> Currently the pcap-win32.c file contains the Packet32 driver implementation as well as some functions which should be generic/driver-independent for the Windows platform (DllMain entry point, WinSock library initialization). The idea is to keep the Win32-generic code in pcap-win32.c, and split the Packet32 driver implementation into pcap-npf.c.

So this is to support building libpcap without Packet32, as a library to read and write capture files without doing live capture?

guyharris:
> Fix stdint.h macro redefinition warnings on Windows

What are those warnings?  Is the issue that `stdint.h` is getting included by the generated lexical analyzer, but, on Windows, it's getting included after something's already defined some of those macros if they're not already defined, so an early include is needed to ensure that `stdint.h`'s definitions win?

nametoaddr.c
guyharris: I've changed `portability.h` to define `strdup` on MSVC iff it's not already defined, regardless of whether `_DEBUG` is defined or not.  That way, *if* `crtdbg.h` has been included before `portability.h`, and *if* it's already defined `strdup`, it won't redefine it, otherwise it *will* define it.

This requires that `crtdbg.h` be included before `portability.h`; I've checked in another change to move the includes of `sockutils.h`, which may include `crtdbg.h`, before includes of `portability.h` or of files that include `portability.h`.

On my Windows 7 VM, with VS 2015, that appears to eliminate build issues with `strdup`, both with Debug, Release, and RelWithDebugInfo builds.

@gvanem, is that an OK fix?  If not, what's better?

This should remove the need to define `strdup` in any `.c` file, or in any `.h` file other than `portability.h`.

guyharris:
> Is the issue that `stdint.h` is getting included by the generated lexical analyzer, but, on Windows, it's getting included after something's already defined some of those macros if they're not already defined, so an early include is needed to ensure that `stdint.h`'s definitions win?

It's because the output of Flex assumes that if the compiler isn't claiming, via the value of `__STDC_VERSION__`, to be full C99, it can't include `stdint.h` to get various sized integer types, and their maximum values, defined, and defines them itself, but `scanner.l` was including `stdint.h`, after those definitions, if we have a reason to believe it's available, even if it's an MSVC implementation that doesn't claim to be full C99 but is sufficiently recent to have `stdint.h`.  This caused multiple definitions.

I moved a bunch of the `#if` stuff that checked what to include on UN\*X in various source files, as well as the `#if` stuff that checked what to include on Windows in `pcap-stdinc.h`, into a `pcap-types.h` file and, in `scanner.l`, included that file at the beginning of the generated scanner.  This fixes those warnings.

So now, the only stuff from this pull request that's of any interest is the stuff to allow building without Packet32; you should merge with the current master, leaving behind only the building-without-Packet32 changes, and change the commit message to indicate that it's allowing building without Packet32, rather than that it's fixing "build issues" such as the now-fixed warnings.

nametoaddr.c
gvanem: With the recent fixes, I get no warnings with `_DEBUG`.

But `<crtdbg.h>` included in `sockutils.h` should perhaps be moved to `portability.h`?  Besides, `pcap-stdinc.h` has now become so small, it could be moved into `pcap-types.h`.

gvanem:
What is the point of adding `HAVE_PACKET32`. In which Wpcap case is using `Packet.dll` not wanted or possible? Wpcap w/o `Packet*.c` code is AFAICS just futile; even with `NPcap`.

blazeable:
> So this is to support building libpcap without Packet32, as a library to read and write capture files without doing live capture?

Yes, and also to break the dependency with Packet32.

> What is the point of adding HAVE_PACKET32. In which Wpcap case is using Packet.dll not wanted or possible? Wpcap w/o Packet*.c code is AFAICS just futile; even with NPcap.

In case you want to develop your own NDIS 6.0 filter driver for packet capture which is not bound by the legacy of Packet32 and you want to support it in libpcap :) Breaking the dependency on Packet32 is something I need to do anyway, and I figured it could be beneficial for the project as a whole, since there is little flexibility when building on Windows at the moment.

pcap.c
blazeable: I figured it was a more reliable way to correct the issue, even if it is uglier I agree. The potential problem with making the change in portability.h is that something included after it may try to redefine it (and this may be a standard header file, apparently). Doing this reliably in portability.h would then require ensuring the file is included as close to the source code as possible, which is more intrusive and probably not a good idea. Defining the macro directly in the source file after all headers have been included seemed like the safest, most conservative option.

blazeable:
> So now, the only stuff from this pull request that's of any interest is the stuff to allow building without Packet32; you should merge with the current master, leaving behind only the building-without-Packet32 changes, and change the commit message to indicate that it's allowing building without Packet32, rather than that it's fixing "build issues" such as the now-fixed warnings.

I created pull request before pushing those changes, wasn't expecting them to show up in the existing pull request. I'll do this.

pcap.c
guyharris: Making the change in source files runs the risk of new code being written, using `strdup()`, but not adding the #define (lots of new stuff is written on UN\*X, where the #define isn't necessary).

nametoaddr.c
guyharris: I've moved the include of `winsock2.h` into `pcap-types.h`, as it's needed to define `u_int`; `pcap-types.h` already includes `sys/types.h` to define `u_int` on UN\*X and MS-DOS.  (So does `pcap/pcap.h`, so perhaps `pcap-types.h` neither needs to include `winsock2.h` nor `sys/types.h`.)

I also moved the necessary other includes to the individual files, as needed; not all files need all of those includes.  That's also similar to what we do on UN\*X.

That renders `pcap-stdinc.h` unnecessary, so I got rid of it.

PR #604, opened 2017-09-04
Enable building libpcap without Packet32 (WinPcap) on Windows

blazeable:
Currently the pcap-win32.c file contains the Packet32 driver implementation as well as some functions which should be generic/driver-independent for the Windows platform (DllMain entry point, WinSock library initialization). The idea is to keep the Win32-generic code in pcap-win32.c, and split the Packet32 driver implementation into pcap-npf.c.

This change:
- Breaks the dependency on Packet32, giving a bit more flexibility when building on Windows
- Makes it easier to implement support for non-Packet32 capture drivers on Windows


gvanem:
> ... and split the Packet32 driver implementation into pcap-npf.c.

And yet your new `pcap-npf.c` contains lots of non-NPF code. Like remote-capture and DAG-API.

> ... In case you want to develop your own NDIS 6.0 filter driver for packet capture which is not bound
> by the legacy of Packet32 

Can you show me an example of this without a huge rewrite?

guyharris:
> And yet your new `pcap-npf.c` contains lots of non-NPF code. Like remote-capture and DAG-API.

The DAG support was  implemented in WinPcap before the "modularization" of libpcap, so there wasn't a convenient place into which to plug those mechanisms.  It'd be good if the DAG support for Windows were redone to plug into the module mechanism (as is the case for DAG support for UN\*X).

The current `pcap-win32.c` has `HAVE_REMOTE` #ifdefs, but those aren't for remote capture, those are to support some WinPcap extensions; that should be cleaned up.

blazeable:
> And yet your new pcap-npf.c contains lots of non-NPF code. Like remote-capture and DAG-API.

It's not really "new". I renamed pcap-win32.c to pcap-npf.c, then recreated pcap-win32.c containing only DllMain and pcap_wsockinit. This is, I think, the most limited set of changes needed to break the Packet32 dependency, and easily allow adding new pcap-xxx.c capture driver support by adding more cases in CMakeLists.txt.

At the moment, DAG API is only supported through WinPcap. I have never used the API, don't have the software, don't really know what it is and am not really interested by it, so I'm not in a position to refactor it into pcap-dag.c.

I hadn't considered the remote capture code in it, but what's currently in pcap-win32.c that isn't in pcap-rpcap.c or rpcapd?

blazeable:
> Can you show me an example of this without a huge rewrite?

Not at the moment, no. libpcap support for my driver is not currently in a working state, as I'm still trying to come up with a good solution for integrating my Event Tracing for Windows-based driver with the file descriptor-based model libpcap requires.

guyharris:
> The DAG support was implemented in WinPcap before the "modularization" of libpcap, so there wasn't a convenient place into which to plug those mechanisms. It'd be good if the DAG support for Windows were redone to plug into the module mechanism (as is the case for DAG support for UN*X).

This is particularly bad because part of the DAG support is in Packet32.dll, as is almost all of the AirPcap support; that was, again, done before the modularization of libpcap, so it's done that way for historical reasons.  I may ask the Endace people if they could either 1) add Windows support to `packet-dag.c` or do a clean `packet-dag-win32.c`, allowing all the DAG code to be removed from `packet-win32.c` (and perhaps from `Packet32.dll`, although that might break compatibility if anybody directly used `Packet32.dll` APIs in code that captured from DAG adapters - as might avoiding using `Packet32.dll` for DAG adapters in libpcap).

guyharris:
> the file descriptor-based model libpcap requires

libpcap's only APIs that deal with file descriptors are UN\*X-only; `pcap_get_selectable_fd()` is not available on Windows or MS-DOS.  `pcap_fileno()` is offered on Windows, but people shouldn't use it unless they know *for certain* that they're using an NPF device rather than a DAG card or AirPcap adapter.  Feel free to allow it to fail.

pcap-win32.c
```diff
@@ -32,7 +32,7 @@
  *
  */
 
-#include "pcap-stdinc.h"
+#include "pcap-types.h"
```
guyharris: This change doesn't apply to the current master branch.

guyharris:
I've made changes that should make it possible to build on Windows without NPF.

If you have any changes that you need to make to the code in the master branch, you should make a new clone of the master branch, make *that* your master branch, and make those changes, and only those changes, there; do not try to do anything more with this repository.

sgeto:
> I'm still trying to come up with a good solution for integrating my Event Tracing for Windows-based driver with the file descriptor-based model libpcap requires.

@blazeable
 Very exciting. Any news on that? How far are you?

PR #608, opened 2017-09-13
Mingw compatibility via cmake + some general win32 adjustments

sgeto:
Hello

In my other pull request I failed to realize that not the MinGW compiler (or linker) were unable to deal with *.lib files, but libtool, which I had been using almost exclusively at that time. libpcap doesn't use libtool. And cmake wasn't really my thing back then.

In the light of this, this new pull request now attempts to add MinGW compatibility to libpcap using cmake.

Along the way, I also took the audacity to fix a few MSVC/cmake related issues I encountered. Hope that's fine.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -351,7 +351,7 @@ if(PCAP_TYPE STREQUAL "npf")
     #
     # Link with packet.dll before WinSock2.
     #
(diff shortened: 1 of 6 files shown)


guyharris:
> Along the way, I also took the audacity to fix a few MSVC/cmake related issues I encountered. Hope that's fine.

It's fine, but I'd prefer that the individual changes be separate pull requests.

PR #609, opened 2017-09-14
Created a FindPacket.cmake module to deal with Packet.lib ins and outs

sgeto:
I changed things up a little.
I think the most notable benefit of *this* approach is that developers can point to Packet.lib and its headers in numerous other ways now, or just simply have them in their compilers' search path. That wasn't possible before.

So setting the PACKET_DLL_DIR option is optional now. ${HAVE_PACKET32} becomes true if cmake can find the library (and headers) in any others way.
Of course the regular way (-DPACKET_DLL_DIR=/path/to/packet/folder) still works.
In addition, PACKET_DLL_DIR can be set as an environment variable too.

The PacketIsLoopbackAdapter function check now works too. The location of the library had to be provided prior of running the check if it isn't in a system path.

### DIFF
--- a/cmake/FindPacket.cmake
+++ b/cmake/FindPacket.cmake
@@ -1,0 +1,88 @@
+#
+# Copyright (C) 2017 Ali Abdulkadir <autostart.ini@gmail.com>.
+#
+# Permission is hereby granted, free of charge, to any person
+# obtaining a copy of this software and associated documentation files
+# (the "Software"), to deal in the Software without restriction,
+# including without limitation the rights to use, copy, modify, merge,
+# publish, distribute, sub-license, and/or sell copies of the Software,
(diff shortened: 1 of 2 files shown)


cmake/FindPacket.cmake
```diff
@@ -0,0 +1,88 @@
...
+
+# Find the header
+find_path(PACKET_INCLUDE_DIR Packet32.h
+  HINTS
+  "${PACKET_DLL_DIR}"
+  PATH_SUFFIXES include Include
+)
+
+if(PACKET_INCLUDE_DIR)
+message(STATUS "Found Packet32.h: ${PACKET_INCLUDE_DIR}")
```
guyharris: This line should be indented.

PR #611, opened 2017-09-14
Fixed naming collision in MSVC builds (cmake)

sgeto:
The dll's import library as well as the static library are named wpcap.lib.
The static library is created last and deletes the import library.

There's no agreed naming convention for static libraries on MSVC.
A popular conventions is to add the _static suffix to the static library.

Microsoft sometimes adds the lib prefix to their static libraries (libvcruntime.lib, libucrt.lib). This is also how [boost ](http://www.boost.org/doc/libs/1_42_0/more/getting_started/windows.html#library-naming) decided to name their static libraries on MSVC.
I like that, but this may not be obvious to everyone.

The new name for the static is of course up to you.
Another way to fix this is would be to change the output directory. But both libraries are meant to be eventually installed into lib.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -750,7 +750,12 @@ add_library(${LIBRARY_NAME}_static STATIC
     ${PROJECT_SOURCE_LIST_H}
 )
 add_dependencies(${LIBRARY_NAME}_static SerializeTarget)
-set_target_properties(${LIBRARY_NAME}_static PROPERTIES OUTPUT_NAME "${LIBRARY_NAME}")
+
+if(MSVC)
+    set_target_properties(${LIBRARY_NAME}_static PROPERTIES OUTPUT_NAME "${LIBRARY_NAME}_static")
+else(MSVC)
+    set_target_properties(${LIBRARY_NAME}_static PROPERTIES OUTPUT_NAME "${LIBRARY_NAME}")
+endif(MSVC)
 
 target_link_libraries(${LIBRARY_NAME} ${PCAP_LINK_LIBRARIES})
 


guyharris:
> Microsoft sometimes adds the lib prefix to their static libraries (libvcruntime.lib, libucrt.lib). This is also how boost decided to name their static libraries on MSVC.
> I like that, but this may not be obvious to everyone.

It should be obvious to anybody who's been near a UN\*X system (although UN\*X systems also use a "libXXX" name for their *dynamic* libraries).

We'll go with this for now; I'll look at what WinPcap does when I fire up my Windows 7 VM.

sgeto:
> It should be obvious to anybody who's been near a UN*X system
Makes sense actually

> I'll look at what WinPcap does when I fire up my Windows 7 VM.
To my knowledge, winpcap never had static libraries.

guyharris:
> To my knowledge, winpcap never had static libraries.

I.e., it may have had an import library named `wpcap.dll`, but it didn't have a static library named `{something}wpcap{something}.dll`?

sgeto:
Yes. The only wpcap.lib they had was an import library.
Their repo also doesn't have VS solution/project for a static library.

sgeto:
lol their repo uses this one as a submodule... which never had one.

guyharris:
Their repository is a relatively recent creation, and I think it *originally* may have had their version of the libpcap source, but they then switched to using the libpcap repository.

But they haven't made any releases *from* that repository.

sgeto:
Looks like they had a draft for a NMake Makefile that builds a static solution way back in 2005.
But they were planning to name it wpcap.lib as well.
commit 20f8c8f63f03555ca8f718fdabe618c3f4a18280

sgeto:
https://code.wireshark.org/review/gitweb?p=winpcap.git;a=commit;h=20f8c8f63f03555ca8f718fdabe618c3f4a18280

gvanem:
Should the `Win32/Prj/wpcap.*` be updated also? 
Or are we supposed to generate a Solution from Cmake from now on?

sgeto:
I'd say yes, but nmap uses the the static VS 2013 project file to build their fork of winpcap.
Things will most likely break on their side if changes are made there,
See [this](https://github.com/nmap/nmap/issues/593) issue.

PR #613, opened 2017-09-17
optimizer: replacing unknown value with unknown value is not a no-op

mkubecek:
Function vstore() checks if new and old value are the same and if they are,
it replaces the instruction with NOP. We must not do this if both old and
new value are 0 which means "unknown" as in this case, the actual values
(at runtime) can differ so that the instruction cannot be omitted.

An example of a program where this leads to incorrect optimization is shown
in issue #581.

### DIFF
--- a/optimize.c
+++ b/optimize.c
@@ -578,7 +578,7 @@ F(opt_state_t *opt_state, int code, int v0, int v1)
 static inline void
 vstore(struct stmt *s, int *valp, int newval, int alter)
 {
-	if (alter && *valp == newval)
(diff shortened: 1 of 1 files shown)


guyharris:
I think part of the problem is that, traditionally, *all* tests done by the generated code were part of the "match or don't match?" decision tree, with no tests done *solely* to set some memory location's value and ultimately ending up at the same instruction, and the optimizer didn't have to deal with cases where two code paths, in each of which a given location had a known value, merged back into a single code, with the two values being different, and thus with the location's value being unknown at the beginning of that code path.

PR #614, opened 2017-09-17
optimizer: show edges in dot graph

mkubecek:
Resetting ic->cur_mark to 0 in dot_dump() results in no edges in dot graph
exported for pcap_optimizer_debug > 3 as all blocks are marked with 1 after
running dot_dump_node(); resetting ic->cur_mark to 0 and incrementing it
results in all blocks "marked" when we enter dot_dump_edge().

Fixes: fb2069644165 ("Don't use global state for the BPF compiler.")


mcr:
I'm thinking that we should have some test cases where we simply export the
resulting BPF program and do a diff against expected results.
As I understand it, in this case we are going to export dot results rather
than BPF data.  Is it worth doing regression on this part?

--

infrastation:
That should be useful if works as expected.

PR #616, opened 2017-09-18
Are we a submodule?

sgeto:
I had this idea yesterday, but I am not convinced of it that much anymore.

The plan was to add additional include and lib directories as hints (places to look for packet.lib and co.) if they exist.

What do you think?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,9 +76,7 @@ if(WIN32)
     endif(PACKET_FOUND)
 
     include_directories(
-        ../Common/
         Win32/Include
-        Win32/WpdPack/Include
     )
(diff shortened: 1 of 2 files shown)


guyharris:
The "Fixed "else" block" commit looks like a fix independent of the "Are we a submodule?" change; is that the case?  If so, it should be submitted as a separate pull request.

sgeto:
It's related. I initially planned to used two variables `SUBMOD_LIBDIR_DBG `and `SUBMOD_LIBDIR_REL ` then decided to just `SUBMOD_LIB_DIR` instead. I forgot to reflect this change there.

sgeto:
I may come back to this...

PR #617, opened 2017-09-18
Fix for BUILD_SHARED_LIBS option

sgeto:
The fact that there are two add_library() targets (one explicitly set to static) causes the BUILD_SHARED_LIBS global flag to not do what it was created for.
Manually conditionalizing it is!

BTW If you could bump up the minimum required cmake version (to at least 2.8.8) things could be done a bit more elegantly.
See [Object Libraries](https://cmake.org/cmake/help/v3.2/command/add_library.html#object-libraries)
This would allow using the same object files to build both static and dynamic libraries.
Doing so manually is not recommended.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -787,13 +787,15 @@ add_custom_target(SerializeTarget
     ${CMAKE_CURRENT_BINARY_DIR}/version.c
 )
 
-add_library(${LIBRARY_NAME} SHARED
-    ${PROJECT_SOURCE_LIST_C}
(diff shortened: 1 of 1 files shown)


guyharris:
> BTW If you could bump up the minimum required cmake version (to at least 2.8.8) things could be done a bit more elegantly.

OK, in some number of years Solaris may finally be gone, and we won't have to worry about having to use the cmake that Oracle's package system provides, which is 2.8.6.  Or maybe OpenCSW provides a newer one.

sgeto:
I see

PR #618, opened 2017-09-18
Use batch files only when building wpcap in cmd.exe

sgeto:
These 5 commits go together.

BTW Why is the C++ compiler employed? The configuration phase could be speed up (by a few seconds) without it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -703,7 +703,7 @@ set_source_files_properties(${CMAKE_CURRENT_BINARY_DIR}/grammar.c PROPERTIES
 #
 #set(PROJECT_SOURCE_LIST_C ${PROJECT_SOURCE_LIST_C} ${CMAKE_CURRENT_BINARY_DIR}/grammar.c)
 
-if(WIN32)
+if(WIN32 AND NOT MSYS AND NOT ${CMAKE_CROSSCOMPILING})
     #
     # CMake does not love Windows.  Convert its UN*X-style paths to
     # Windows-style paths to hand to Windows programs.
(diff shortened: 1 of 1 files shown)


guyharris:
> These 5 commits go together.

They look largely independent, e.g. changing the capitalization of `wspiapi.h` doesn't seem to depend on the other changes, nor does it seem to be a dependency for any of the other changes.

I'd prefer it if you submitted all the changes that neither depend on other changes nor are dependencies of other changes as individual changes, and if there *are* dependencies, but there's no cycle in the dependency graph, submitting them as independent changes in dependency order.

guyharris:
> BTW Why is the C++ compiler employed?

"Employed" in that we're setting some CXX variables or "employed" in the sense that we actually compile C++ code (or compile C code with a C++ compiler)?

sgeto:
not setting the language your project uses make cmake assume C++.
this makes it perform a few tests, but in the end it only uses the c compiler

sgeto:
I really just thought that most MinGW compilers build the project only if these five changes are in place

sgeto:
reduced the PR to one commit

CMakeLists.txt
```diff
@@ -703,7 +703,7 @@ set_source_files_properties(${CMAKE_CURRENT_BINARY_DIR}/grammar.c PROPERTIES
 #
 #set(PROJECT_SOURCE_LIST_C ${PROJECT_SOURCE_LIST_C} ${CMAKE_CURRENT_BINARY_DIR}/grammar.c)
 
-if(WIN32)
+if(WIN32 AND NOT MSYS AND NOT ${CMAKE_CROSSCOMPILING})
```
guyharris: So that's "cross compiling" as in "building for Windows on some flavor of UN\*X", which means "you have UN\*X tools to build `version.c` and `pcap_version.h`"?

So with MinGW on Windows *without* UN\*X tools, this will use the batch file?

sgeto: Yes. In addition, there is a cmake generator that uses UNIX tools that have been ported to windows to build the project (including a POSIX/bourne compatible shell and of course the gcc windows port). It is know as Msys. cmake stays clear of the batch file if this generator is chosen.

PR #619, opened 2017-09-18
Added if block around set_target_properties

sgeto:
added by commit 3c4120 (my PR)


guyharris:
Why not generate complete debugging information for a static library?  Does it not work with static libraries?

sgeto:
> Why not generate complete debugging information for a static library? Does it not work with static libraries?

It works for static libraries. This is the default. I made this change because:

- users would need to remember to keep the resulting pdb file and put it somewhere in their path. remembering to do the same with the pdb file for the dll is easier since the dll itself must be reachable at runtime.

- adding the pdb file as install target is rather complex. meaning there will be few extra lines that could be just a single one. see 
https://cmake.org/cmake/help/v3.1/prop_tgt/COMPILE_PDB_OUTPUT_DIRECTORY.html
https://cmake.org/cmake/help/v3.1/prop_tgt/PDB_OUTPUT_DIRECTORY.html

anyways, I can implement that if you prefer it. in fact, I already have.
maybe you could still merge this one? the BUILD_SHARED_LIBS option currently fails without it.

guyharris:
> the BUILD_SHARED_LIBS option currently fails without it.

So how does doing something only if the BUILD_SHARED_LIBS option is set fix the BUILD_SHARED_LIBS build?  Or did you mean that building *static* libraries fails without it - i.e., *not* setting the BUILD_SHARED_LIBS option fails without the change?

sgeto:
I mean that that I missed to add `if(BUILD_SHARED_LIBS)` around this one line in my BUILD_SHARED_LIBS pull request. Building the static library only by setting BUILD_SHARED_LIBS to off now doesn't work because of it.

This is for the most part unrelated to debugging symbol discussion we are having. It's something I forgot earlier.

PR #624, opened 2017-09-21
Use strtok_r as strtok_s under MinGW as well

sgeto:
With cmake, strtok_r isn't tested for on windows and assumed not to exist.
Newer version of MinGW have implemented it. Older version (e. g. the one provided by Debian) however seem to only have it under strtok_s.
It's probably a ok to stick with strtok_s for all of windows for now.


guyharris:
> With cmake, strtok_r isn't tested for on windows and assumed not to exist.

Eh?  What about

    check_function_exists(strtok_r HAVE_STRTOK_R)

under "Functions." in `CMakeLists.txt`?  Isn't that checking for `strtok_r()` on *all* platforms?

PR #625, opened 2017-09-21
 Define ssize_t as int only when _SSIZE_T_DEFINED is not defined

sgeto:
it's already defined.


gvanem:
IMHO, it's better to just say:
```c
#if defined(_WIN32) && !defined(_SSIZE_T_DEFINED)
  typedef int ssize_t;
#endif
```

sgeto:
You're right. It's better this way.
I'll change it to this

PR #629, opened 2017-09-24
replaced all tabs in CMakeLists.txt with (four) spaces

sgeto:
for consistency.

I also think that 2 spaces should be used for indentation.
That how most projects do it.
Can I go for it?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -143,7 +143,7 @@ if(HAVE_NET_PFVAR_H)
 int
 main(void)
 {
-	return PF_NAT+PF_NONAT+PF_BINAT+PF_NOBINAT+PF_RDR+PF_NORDR;
+    return PF_NAT+PF_NONAT+PF_BINAT+PF_NOBINAT+PF_RDR+PF_NORDR;
 }
 "
         HAVE_PF_NAT_THROUGH_PF_NORDR)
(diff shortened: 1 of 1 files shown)


guyharris:
Tab stops are set every 8 spaces in all sane environments. :-)

Therefore, for consistent 4-space indentation, tabs should be replaced by 8 spaces; I've done that in 91140544fb35aedb530be9a96f651aba75e22652.  Otherwise, lines that were indented with 8 spaces before the change, and lines that were indented with a tab before the change, which are *intended* to be indented the same, end up not being indented the same.

> I also think that 2 spaces should be used for indentation.
> That how most projects do it.

Most projects where?  Wireshark's top-level `CMakeLists.txt` uses tab/8-space indentation.  CMake itself isn't completely consistent, but the current version in Git appears to be using 2-space indentation for most files (at least one seems to have 3-space indentation).

> Can I go for it?

As long as it involves turning 4N spaces at the beginning of a line into 2N spaces - i.e., as long as it preserves the structure of the indentation - yes.

sgeto:
Sorry I mean most projects I've seen...
True, not even cmake is exactly consistent.

Actually, I have a few more local changes. I think I should get them out of the way before changing indentation. I hate merge conflicts...

PR #631, opened 2017-09-24
Link with -static-libgcc on MinGW when USE_STATIC_RT is enabled

sgeto:
To avoid introducing a hard-to-satisfy dependency on a specific flavor of libgcc_s_*-1.dll, MinGW builds should by default link statically with libgcc. This can be changed by setting USE_STATIC_RT to off.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,9 +26,9 @@ endif()
 ###################################################################
 
 option(INET6 "Enable IPv6" ON)
-if(MSVC)
+if(WIN32)
     option(USE_STATIC_RT "Use static Runtime" ON)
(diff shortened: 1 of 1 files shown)


guyharris:
@gvanem?

gvanem:
I don't do much MinGW anymore. But dropping dependency on `libgcc_s*.dll` is a good thing IMHO. Besides, I don't understand much Cmake either. So just merge it.

sgeto:
The various libgcc*.dll are often incompatible. Meaning that each one will need to ship with the binaries that have been build with them,

BTW your recent changes fried the solution in Win32/Prj.
cmake to the rescue!

PR #638, opened 2017-10-01
Build and install the remote packet capture daemon (cmake)

sgeto:


sgeto:
winflexbison, you're weird...

rpcapd/daemon.c
```diff
@@ -35,6 +35,10 @@
 
 #include "ftmacros.h"
 
+#if defined(_WIN32) && !defined(_SSIZE_T_DEFINED)
```
guyharris: And daemon.c no longer uses `ssize_t`, so this change is no longer necessary.

rpcapd/daemon.c
```diff
@@ -59,6 +55,10 @@ typedef int ssize_t;
 #include <shadow.h>		// for password management
 #endif
 
+#if defined(_WIN32) && !defined(_SSIZE_T_DEFINED)
```
guyharris: Again, no longer necessary, as `daemon.c` no longer uses `ssize_t`.

pcap-dll.rc
```diff
@@ -1,12 +1,23 @@
 #include "config.h"
```
guyharris: Why?  Why not have separate .rc files for the library and the daemon, given that they're different pieces of software?

CMakeLists.txt
```diff
@@ -478,11 +478,6 @@ else()
     endif(NOT HAVE_STRTOK_R)
 endif(WIN32)
 
-if(HAVE_REMOTE)
```
guyharris: Please submit this as a separate pull request.

rpcapd/win32-svc.c
```diff
@@ -33,7 +33,6 @@
 
 #include "rpcapd.h"
 #include <signal.h>
-#define	_WINSOCKAPI_
```
guyharris: Please submit this as a separate pull request.

rpcapd/daemon.c
```diff
@@ -97,13 +97,13 @@ static void daemon_seraddr(struct sockaddr_storage *sockaddrin, struct rpcap_soc
 static void *daemon_thrdatamain(void *ptr);
```
guyharris: Please submit this, plus "One more typo", as a separate pull request.

rpcapd/win32-svc.c
```diff
@@ -95,7 +95,7 @@ void WINAPI svc_control_handler(DWORD Opcode)
 				This can be seen by placing a breakpoint at the end of svc_main(), in which you will 
 				see that is never reached. However, as soon as you set the service status to "stopped",
 				the StartServiceCtrlDispatcher() returns and the main thread ends. Then, Win32 has a good
-				authomatic cleanup, so that all the threads which are still running are stopped
```
guyharris: Please submit this, plus "Typos!", as a separate pull request.

pcap-dll.rc
sgeto: Well, this was just my first thought since they go hand in hand.
Two separate files it is!

rpcapd/daemon.c
sgeto: > Again, no longer necessary, as daemon.c no longer uses ssize_t.

Since when? It seemed necessary when pushed the PR and
I can't really check right now.

guyharris: Since de7aab10f11dd9b64d55afffaeb40e264370ce8d.

rpcapd/win32-svc.c
sgeto: This goes along with its replacement set [here](https://github.com/sgeto/libpcap/blob/rpcapd/rpcapd/CMakeLists.txt#L26)

If submitted separately, there will be a nasty warning when building rpcapd.
I personally think this shouldn't be separated.

rpcapd/win32-svc.c
sgeto: okay

rpcapd/rpcapd.c
```diff
@@ -91,30 +91,34 @@ static void main_cleanup_childs(int sign);
 */
 static void printusage(void)
 {
+	(void)fprintf(stderr, "RPCAPD, a remote packet capture daemon.\n"
+	"Compiled with %s\n\n", pcap_lib_version());
 	char *usagetext =
-	"USAGE:\n"
+	"USAGE:"
 	" "  PROGRAM_NAME " [-b <address>] [-p <port>] [-6] [-l <host_list>] [-a <host,port>]\n"
```
sgeto: Not to self: there is no -6 option. Or is there?

rpcapd/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+  endif(PTHREADS_FOUND)
+endif(NOT MSVC)
+
+if(CMAKE_USE_PTHREADS_INIT OR PTHREADS_FOUND)
+
+set(BUILDING_RPCAPD 1)
+
+if(WIN32)
+set(RPCAPD_EXTRA_SOURCES win32-svc.c ${PCAP_DLL_RC})
+add_definitions(-DWIN32_LEAN_AND_MEAN -DBUILDING_RPCAPD)
```
guyharris: Where is `BUILDING_RPCAPD` used?  If it's only used to allow both the DLL and rpcapd to share a single .rc file, that's not necessary if they don't share a single .rc file.

rpcapd/rpcapd.c
guyharris: No, there's a `-4` option.  Fixed in 8f26928e66da7072957b0c3c309f4f89005aec3b.

sgeto: 👍

rpcapd/win32-svc.c
guyharris: If the right way to prevent `winsock.h` being included by `windows.h` is to define `WIN32_LEAN_AND_MEAN`, then:

- if autotools are used to build rpcapd on Windows, it also need to be changed to define `WIN32_LEAN_AND_MEAN` on Windows, and that change can include the removal of the `#define`;

- if autotools *aren't* used to build rpcapd on Windows, then the change to remove the `#define` of `_WINSOCKAPI_` can be done, as a separate change, before submitting changes to build it with CMake, as submitting it won't break any build.

sgeto: 1. it's "just" a warning. There won't be any build errors.
2. autotools can't build build rpcapd. not even on unix. it errors because it is not setup to link with threads (and crypt). 
on windows, it errors even earlier (during configure). mingw can't pass the ANSI compiler check.
3. before you fix that, I have already done quite a few things (locally) on/for the autotools build system. I will push them (one by one :-D) after my cmake changes. If that's okay.

guyharris: > autotools can't build rpcapd. not even on unix.

The OS running on my development machine is a trademark-licensed Unix, and rpcapd builds fine on it, because that OS's pthreads APIs are part of the system library (`libSystem.dylib`).  It happens to have an `-lpthreads` library, but that might just be something Apple added to help build processes that *assume* an `-lpthreads` library.  It may be that other UN\*Xes don't have an `-lpthreads`.

sgeto: I see I rushed into conclusion there. It doesn't build on my linux box

guyharris: The best way to handle that is probably to first check whether, for example, `pcap_create()` is available with the standard libraries and, if it's not, check whether it's available in libpthreads and, if it is, add that to the list of libraries with which to link (and if it isn't, fail).

sgeto:
Any decision on this?
I'm a bit stalled here.

CMakeLists.txt
guyharris: So why is the remote capture stuff moved that late in the file? I view the remote capture client part, at least, as being just another capture type, so it belongs with them.

In a511023dd7ebacff23ab220aa8315ccff2597a3b I've moved it right after the TurboCap tests in both autoconf and CMake.

rpcapd/rpcapd.c
```diff
@@ -91,36 +91,39 @@ static void main_cleanup_childs(int sign);
 */
 static void printusage(void)
 {
```
guyharris: This is a separate change that just cleans up the code and the usage message, and thus applies regardless of the build system or the target platform; please submit it as a separate pull request.

guyharris:
> Any decision on this?

1. Please submit the usage-message change as a separate pull request.

1. Why move the test for `ENABLE_REMOTE`?

CMakeLists.txt
sgeto: mostly to get all needed libraries into PCAP_LINK_LIBRARIES, which this early doesn't include any.
On windows, those would be winsock 2 and packet.

sgeto: update: the new place is fine. I will revert this change.

CMakeLists.txt
```diff
@@ -1352,6 +1352,16 @@ if(WIN32)
 endif(WIN32)
 
 ######################################
+# Remote Packet Capture Support
+######################################
+
```
guyharris: Shouldn't you just add `add_subdirectory(rpcapd)` to the existing code that tests `ENABLE_REMOTE`?  (Note that I changed the option to `ENABLE_REMOTE` from `HAVE_REMOTE`; the new name better describes what what the option actually means.)

rpcapd/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+    include_directories(${PTHREADS_INCLUDE_DIR})
+  else(PTHREADS_FOUND)
+    message(WARNING "Pthreads not found. Won't be able to build rpcapd.")
+  endif(PTHREADS_FOUND)
+endif(NOT MSVC)
+
+if(CMAKE_USE_PTHREADS_INIT OR PTHREADS_FOUND)
+
+if(WIN32)
+set(RPCAPD_EXTRA_SOURCES win32-svc.c ${pcap_SOURCE_DIR}/rpcapd.rc)
```
guyharris: Shouldn't the two lines within `if(WIN32)` be indented?

CMakeLists.txt
sgeto: yes. That confused me a little. I just fixed that.

rpcapd/CMakeLists.txt
sgeto: yes they should be.

rpcapd/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+  if(PTHREADS_FOUND)
+    set(PCAP_LINK_LIBRARIES ${PCAP_LINK_LIBRARIES} ${PTHREADS_LIBRARY})
+    include_directories(${PTHREADS_INCLUDE_DIR})
+  else(PTHREADS_FOUND)
+    message(WARNING "Pthreads not found. Won't be able to build rpcapd.")
+  endif(PTHREADS_FOUND)
+endif(NOT MSVC)
+
+if(CMAKE_USE_PTHREADS_INIT OR PTHREADS_FOUND)
+
```
guyharris: Also, shouldn't all the stuff inside the pthreads test be indented?

sgeto: yep...

guyharris:
Thanks for deprecating OpenSSL, Apple.

guyharris:
And thanks, GitHub, for not letting me merge this anyway, because I'm on my iPhone.  I'll do it when I get home. (And then figure out what to do about libcrypt on macOS - it's an issue for tcpdump as well.)

sgeto:
👍 👍

PR #640, opened 2017-10-02
excluded unavailable members of pcap_t on windows

sgeto:
dbus isn't exactly at home on windows and I am sure no dbus guru.
Any suggestions on how to see whether it runs as expected?

### DIFF
--- a/pcap-dbus.c
+++ b/pcap-dbus.c
@@ -208,12 +208,16 @@ dbus_activate(pcap_t *handle)
 	handle->setfilter_op = install_bpf_program; /* XXX, later add support for dbus_bus_add_match() */
 	handle->setdirection_op = NULL;
 	handle->set_datalink_op = NULL;      /* can't change data link type */
+#ifndef _WIN32
 	handle->getnonblock_op = pcap_getnonblock_fd;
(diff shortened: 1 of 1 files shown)


guyharris:
> dbus isn't exactly at home on windows

Neither is UN\*X-style non-blocking mode or `select()` as a general event-checking mechanism rather than a hack for WinSock to help porting UN\*X programs that do `select()` on network connections.

The lines being `#ifndef`fed out are just doing standard UN\*X things, so just eliminating them will make this work as well as possible (which isn't very well, even on UN\*X, as we don't have a descriptor for D-Bus on which to *do* a `select()`).  If you have a program on UN\*X that does capturing and that  does event checking on, among other event sources, `pcap_t`s, your program won't work if capturing on D-Bus.

So this is the right way to handle this.

guyharris:
Actually, those members *are* available, but the *functions* in question aren't available.

And they won't work on UN\*X, as they try to set non-blocking mode on `p->fd`, but `p->fd` is -1, because there *is* nothing that should be treated as "the file descriptor".

Doing non-blocking mode probably involves some additional work, so the get and set routines should just return an error.  I've done that.

PR #642, opened 2017-10-02
Build and install pcap-config on MinGW as well

sgeto:


CMakeLists.txt
```diff
@@ -1715,7 +1715,7 @@ install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/pcap/ DESTINATION include/pcap)
 install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/pcap.h DESTINATION include)
 
 # On UN*X, generate pcap-config and arrange that it be installed.
```
guyharris: The comment should be updated to "On UN*X, and on Windows when not using MSVC,"

PR #645, opened 2017-10-04
Use (native) m4 macro to get version string from file

sgeto:


configure.ac
```diff
@@ -16,21 +16,7 @@ dnl
...
-# So we put a place holder here.  The alternative would be to eliminate
-# the VERSION file, directly set the version here, CMakeLists.txt,
-# and msdos/makefile.dj, have pcap_version.h and version.c not be
-# expanded from .in files but instead use PACKAGE_VERSION, use
-# @PACKAGE_VERSION@ in Makefile.in for the full version string, and do
-# what's necessary to get the version in .travis-coverity-scan-build.sh.
-# That would involve updating three files when the version changes.
-#
-AC_INIT(pcap, place-holder-version)
+AC_INIT(pcap, m4_esyscmd_s([cat VERSION]))
```
guyharris: That requires autoconf 2.64 or later, so you'd need to update the `AC_PREREQ` call to say 2.64.

sgeto: Not sure about the minimum requirements.
Tested this on 2.69.

But, would bumping that up be a problem?

guyharris: > Not sure about the minimum requirements.

I just checked the 2.63 and 2.64 source; it's in 2.64, but not in 2.63, so make the required minimum 2.64.

> But, would bumping that up be a problem?

Since we ship the generated configure script with release tarballs (as everybody using autotools is supposed to do), it'd only be a problem for people doing development *and* who don't have a sufficiently recent autoconf *and* who need to change the configure script.  I'm fine with cranking up the requirement, even if that causes them problems.

guyharris:
The Travis build failed because the CMake build was broken, but this doesn't affect CMake builds, and all the autotools builds succeeded.

PR #646, opened 2017-10-05
Prettified printusage() output

sgeto:
sample output:

```
RPCAPD, a remote packet capture daemon.                                                              
Compiled with libpcap version 1.9.0-PRE-GIT (packet.dll version 0.94)                                
                                                                                                     
USAGE: rpcapd [-b <address>] [-p <port>] [-4] [-l <host_list>] [-a <host,port>]                      
              [-n] [-v] [-d] [-s <file>] [-f <file>]                                                 
                                                                                                     
  -b<address>     the address to bind to (either numeric or literal).                                
                  Default: binds to all local IPv4 and IPv6 addresses                                
                                                                                                     
  -p<port>        the port to bind to.                                                               
                  Default: binds to port 2002                                                        
                                                                                                     
  -4              use only IPv4.                                                                     
                  Default: use both IPv4 and IPv6 waiting sockets                                    
                                                                                                     
  -l<host_list>   a file that contains a list of hosts that are allowed                              
                  to connect to this server (if more than one, list them one per line).              
                  We suggest to use literal names (instead of numeric ones) in                       
                  order to avoid problems with different address families.                           
                                                                                                     
  -n              permit NULL authentication (usually used with '-l')                                
                                                                                                     
  -a<host,port>   run in active mode when connecting to 'host' on port 'port'                        
                  In case 'port' is omitted, the default port (2003) is used                         
                                                                                                     
  -v              run in active mode only (default: if '-a' is specified, it accepts                 
                  passive connections as well                                                        
                                                                                                     
  -d              run in daemon mode (UNIX only) or as a service (Win32 only)                        
                  Warning (Win32): this switch is provided automatically when the service            
                  is started from the control panel                                                  
                                                                                                     
  -s<file>        save the current configuration to file                                             
                                                                                                     
  -f<file>        load the current configuration from file; all switches                             
                  specified from the command line are ignored                                        
                                                                                                     
  -h              print this help screen                                                             
                                                                                                     
```


rpcapd/rpcapd.c
```diff
@@ -91,30 +91,34 @@ static void main_cleanup_childs(int sign);
...
-	"      passive connections as well\n"
-	"  -d: run in daemon mode (UNIX only) or as a service (Win32 only)\n"
-	"      Warning (Win32): this switch is provided automatically when the service\n"
-	"      is started from the control panel\n"
-	"  -s <file>: save the current configuration to file\n"
- 	"  -f <file>: load the current configuration from file; all the switches\n"
-  	"      specified from the command line are ignored\n"
-    "  -h: print this help screen\n\n";
+	"              [-n] [-v] [-d] [-s <file>] [-f <file>]\n\n"
+	"  -b<address>     the address to bind to (either numeric or literal).\n"
```
guyharris: The general convention is to separate the flag, and its argument, with a space, e.g.:

    $ grep --help
    usage: grep [-abcDEFGHhIiJLlmnOoqRSsUVvwxZ] [-A num] [-B num] [-C[num]]
            [-e pattern] [-f file] [--binary-files=value] [--color=when]
            [--context[=num]] [--directories=action] [--label] [--line-buffered]
            [--null] [pattern] [file ...]

The angle brackets are OK - some programs use them, some programs don't.

sgeto: > The general convention is to separate the flag, and its argument, with a space

True. Most of these adjustments are based on how wget's help output looks like. except this...

PR #648, opened 2017-10-08
Process, install and symlink man pages on MinGW as well

sgeto:
That took a while for what it is.
There are way too many quote signs going on there.

Anyways, cmake's `create_symlink` command is only available on unix. So with MinGW we will search for and use plain `ln`, which BTW doesn't *really* create any symlinks, but simply duplicates and renames the files in question.

In retrospective, this could have been done right from the beginning using some kind of copy routine... great day.


sgeto:
Can you review this please? I really like to see this come through

guyharris:
Needs conflicts to be resolved.

PR #650, opened 2017-10-09
Fixed wrong quote usage in rc files

sgeto:
The usage of quotes rc files is not very straight forward.
PACKAGE_VERSION_DLL may not be put in quotes except when used for FileVersion. There, it must be in quotes or windres will complain.
PACKAGE_NAME and PACKAGE_VERSION are already carring quotes when arriving form config.h.
So no need to put them in quotes again.

Currently none of these three defines is properly handled by the preprocessor because of this. This PR fixes this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -195,7 +195,7 @@ if(WIN32)
 
     # Append NANO (used for Windows internal versioning) to PCAP_VERSION_PREDLL
     # 0 means unused.
-    set(PACKAGE_VERSION_DLL "${PACKAGE_VERSION_PREDLL},0")
+    set(PACKAGE_VERSION_DLL ${PACKAGE_VERSION_PREDLL},0)
 endif(WIN32)
 
 set(PACKAGE_NAME "${LIBRARY_NAME}")
--- a/pcap-dll.rc
+++ b/pcap-dll.rc
@@ -21,12 +21,12 @@
         VALUE "CompanyName",      "The TCPdump Group"
         VALUE "FileDescription",  "System-Independent Interface for User-Level Packet Capture"
         VALUE "FileVersion",      "PACKAGE_VERSION_DLL"
-        VALUE "InternalName",     "PACKAGE_NAME"
+        VALUE "InternalName",     PACKAGE_NAME
         VALUE "LegalCopyright",   "Copyright (c) The TCPdump Group"
         VALUE "LegalTrademarks",  ""
         VALUE "OriginalFilename", "wpcap.dll"
-        VALUE "ProductName",      "PACKAGE_NAME"
-        VALUE "ProductVersion",   "PACKAGE_VERSION"
+        VALUE "ProductName",      PACKAGE_NAME
+        VALUE "ProductVersion",   PACKAGE_VERSION
       END
     END
   BLOCK "VarFileInfo"
--- a/rpcapd/rpcapd.rc
+++ b/rpcapd/rpcapd.rc
@@ -24,12 +24,12 @@
         VALUE "CompanyName",      "The TCPdump Group"
         VALUE "FileDescription",  "Remote Packet Capture Daemon"
         VALUE "FileVersion",      "PACKAGE_VERSION_DLL"
-        VALUE "InternalName",     "PACKAGE_NAME"
+        VALUE "InternalName",     PACKAGE_NAME
         VALUE "LegalCopyright",   "Copyright (c) The TCPdump Group"
         VALUE "LegalTrademarks",  ""
         VALUE "OriginalFilename", "rpcapd.exe"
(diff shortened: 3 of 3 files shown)


sgeto:
This wpcap is set to pcap's version which is lower that the other much older wpcap.
I doubt that this is any good. If they coexist, the system may use the older one because of its higher version string number.
Okay, I'll say it: You should consider dropping binary compatibility with wpcap by naming this lib pcap on windows as well.

guyharris:
> You should consider dropping binary compatibility with wpcap by naming this lib pcap on windows as well.

I guess one question is who is the "you" here?  The Tcpdump Group doesn't make binary releases; for one thing, we don't maintain any of the kernel-mode code, or DLL wrapper for kernel-mode code, that `pcap-npf.c` uses.  There's the old WinPcap, but no releases of that have been made in ages, and there's [Npcap](http://www.npcap.org), which is actively being developed.

Their home page says

> ***WinPcap compatibility***: For applications that don't yet make use of Npcap's advanced features, Npcap can be installed in “WinPcap Compatible Mode.” This will replace any existing WinPcap installation. If compatibility mode is not selected, Npcap can coexist alongside WinPcap; applications which only know about WinPcap will continue using that, while other applications can choose to use the newer and faster Npcap driver instead.

so they're at least offering binary compatibility.  If it's *not* installed in WinPcap Compatible Mode, it might still not be called `pcap.dll` - it might be called `npcap.dll`.

And they have their own version number; the issue here, with both WinPcap and Npcap, is that what they're releasing are bundles of the libpcap DLL, the packet32 DLL, and the driver, and if the driver and packet32 DLL change, even without a change to libpcap, they'd put out a new release.

guyharris:
So perhaps we should build a `pcap.dll` library, and if anybody wants to build WinPcap or Npcap atop the current libpcap-for-Windows, they can tweak the build process and .rc file as they want.

sgeto:
I think the reason the windows port of this lib was renamed to wpcap in the first place was because it wasn't build directly from pcap's sources, but patched and modified, and also more limited then the original.
This isn't the case anymore (for the most part). The same holds true for windump.
Not many application ported to windows that I know of have renamed their binaries. Especially when they're build directly from the same sources that build the original. So this is left only as a courtesy which I think is slowly getting in the way.

Npcap's driver and lib have the same name as the original wpcap. They went for a name change, but I guess it didn't work out for them.
The text you quoted from their readme isn't very accurate anymore either. _Actual_ WinPcap compatibility has been dropped. It's just that name and where their installer places the DLLs.

> and there's Npcap, which is actively being developed.

Since Npcap is being actively developed and their wpcap.dll is build directly from this repo anyways, they could easily adopt to a name change.
And everybody else should sooner or later get the message as well... or simply stay with wpcap from npcap or older one.

sgeto:
> So perhaps we should build a pcap.dll library, and if anybody wants to build WinPcap or Npcap atop the current libpcap-for-Windows

atop as in "in addition to"?

guyharris:
"Atop" as in "building a WinPcap or Npcap release by combining the WinPcap/Npcap kernel driver and packet DLL with a wpcap/npcap.dll built from the libpcap source, including `pcap-npf.c`.

So, yes, "in addition to" in that sense, i.e. you add the kernel driver+packet DLL and libpcap together, compile them all, and get a binary package to distribute.

sgeto:
wow that's a lot.
so essentially merging most of WinPcap/Npcap's code into this repo?

guyharris:
> so essentially merging most of WinPcap/Npcap's code into this repo?

I'm not currently suggesting that, as that would not only mean merging the code, it'd mean maintaining that code here.

I'm saying that the way somebody could currently provide a libpcap-for-Windows capable of capturing traffic would be to have their own kernel-mode code (and possibly a user-mode library to talk to that code) and somehow combine it with libpcap, whether by copying the libpcap code into their own repository (as Npcap does and as WinPcap did) or by, for example, building and installing their own code, including headers and libraries used by libpcap, building libpcap with that, and then packaging the libpcap library binary with the binaries for their own code.

guyharris:
I made it default to calling the library `pcap.dll` on Windows in f029e5dfe9bd6575daef37c80853bd60dbaf6dcf.  That can be overridden with `-DLIBRARY_NAME`, e.g. `-DLIBRARY_NAME=wpcap`.

sgeto:
And so it begins... 👍 
Could you reflect this change in [pcap-dll.rc](https://github.com/the-tcpdump-group/libpcap/blob/master/pcap-dll.rc#L27) as well?

sgeto:
> by, for example, building and installing their own code, including headers and libraries used by libpcap, building libpcap with that, and then packaging the libpcap library binary with the binaries for their own code.

Hold that thought...

PR #653, opened 2017-10-15
Address CLang "undefined defined" warnings.

infrastation:
With this change in libpcap tcpdump compiles in 79 seconds instead of 145 seconds. Most importantly, it does not flood the terminal with warnings (https://travis-ci.org/the-tcpdump-group/tcpdump/jobs/287485992 is a good example, see commit message for additional details). I am not sure this is the best solution, though the best I could produce so far.

### DIFF
--- a/optimize.c
+++ b/optimize.c
@@ -60,7 +60,7 @@ int pcap_optimizer_debug;
  *
  * This is the same as the count of trailing zeroes in the word.
  */
-#if PCAP_IS_AT_LEAST_GNUC_VERSION(3, 4)
+#if PCAP_IS_AT_LEAST_GNUC_VERSION_3_4
   /*
    * GCC 3.4 and later; we have __builtin_ctz().
    */
--- a/pcap/compiler-tests.h
+++ b/pcap/compiler-tests.h
@@ -59,10 +59,36 @@
  * compiler that claims to be "just like GCC" of that version or a
  * later release.
  */
-#define PCAP_IS_AT_LEAST_GNUC_VERSION(major, minor) \
-	(defined(__GNUC__) && \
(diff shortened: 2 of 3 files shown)


guyharris:
Presumably it's complaining about fff

> Prior to evaluation, macro invocations in the list of preprocessing tokens that will become the controlling constant expression are replaced (except for those macro names modified by the defined unary operator), just as in normal text. If the token '''defined''' is generated as a result of this replacement process or use of the '''defined''' unary operator does not match one of the two specified forms prior to macro replacement, the behavior is undefined.

from the C90 spec's "6.8.1 Conditional inclusion" and the C99 spec's and C11 spec's "6.10.1 Conditional inclusion".

infrastation:
Hmmm, today I have realized there was another solution, which would allow the 2-ary macros:
```c
#if defined(__GNUC__)
#define PCAP_DEFINED_GNUC 1
#else
#define PCAP_DEFINED_GNUC 0
#endif

#define PCAP_IS_AT_LEAST_GNUC_VERSION(major, minor) \
PCAP_DEFINED_GNUC && \
(and so on)
```
Do you think it is worth making it this way?

guyharris:
From a purely stylistic point of view, it might be, although I might change the names to `PCAP_IS_GNUC` and so on.

If one of the goals is to reduce compilation time, we should probably see whether that approach slows compilation down and, if so, how much.

infrastation:
I think there's even a cleaner way, let me make the changes and test it.

infrastation:
OK, the new commit is in the master branch. Sorry for having made more noise than it was worth, the most difficult part was to start working on this issue.

PR #657, opened 2017-10-29
Use pthread_equal() to compare thread IDs

sgeto:
Some pthread implementations define pthread_t as a structure rather than a pointer or an integer.

Perfectly fine according to The Open Group Base Specifications.
What is not fine is for daemon_serviceloop() to assume it to always be an integer.

So to improve portability, pthread_equal() should be used to compare thread IDs.

From http://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_equal.html:

> RATIONALE
> Implementations may choose to define a thread ID as a structure. This allows additional flexibility and
> robustness over using an int. For example, a thread ID could include a sequence number that allows
> detection of "dangling IDs" (copies of a thread ID that has been detached). Since the C language does
> not support comparison on structure types, the pthread_equal() function is provided to compare thread
> IDs.

### DIFF
--- a/rpcapd/daemon.c
+++ b/rpcapd/daemon.c
@@ -2,30 +2,30 @@
  * Copyright (c) 2002 - 2003
  * NetGroup, Politecnico di Torino (Italy)
  * All rights reserved.
- * 
- * Redistribution and use in source and binary forms, with or without 
- * modification, are permitted provided that the following conditions 
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
  * are met:
- * 
- * 1. Redistributions of source code must retain the above copyright 
+ *
+ * 1. Redistributions of source code must retain the above copyright
  * notice, this list of conditions and the following disclaimer.
- * 2. Redistributions in binary form must reproduce the above copyright 
- * notice, this list of conditions and the following disclaimer in the 
- * documentation and/or other materials provided with the distribution. 
- * 3. Neither the name of the Politecnico di Torino nor the names of its 
- * contributors may be used to endorse or promote products derived from 
- * this software without specific prior written permission. 
- * 
- * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS 
- * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT 
- * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR 
- * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT 
- * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, 
- * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT 
- * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, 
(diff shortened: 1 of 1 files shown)


gvanem:
I tried compiling using *Pthreads-w32*  2.9.1 and can confirm:
```
daemon.c(424): error C2083: struct comparison illegal
daemon.c(427): error C2440: '=': cannot convert from 'int' to 'pthread_t'
```

But how do you manage to link `rpcapd.exe`? I'm getting a link error:
```
fileconf.obj : error LNK2019: unresolved external symbol _fopen_safe referenced in function _fileconf_read
```
I suppose `portability.h` should be patched into:
```diff
--- a/portability.h 2017-10-18 11:33:40
+++ b/portability.h 2017-10-29 11:00:40
@@ -100,13 +100,16 @@

   #include <stdio.h>

+  #if defined(BUILDING_PCAP)
   /*
    * Make fopen() call a wrapper that calls fopen_s(), to squelch
    * warnings.
+     * Do not call 'fopen_safe()' while e.g. compiling rpcapd.exe.
    */
   #define fopen(x, y) \
        fopen_safe((x), (y))
   FILE *fopen_safe(const char *filename, const char* mode);
+  #endif
 #endif

 #ifdef _MSC_VER
```

guyharris:
Please rebase this against the head of master; I've done separate commits to remove trailing white space (spaces *and* tabs) and to fix the two typos you fixed.

guyharris:
> I suspect that sockutils.c should be part of rpcapd's sources rather than libpcap's. Can't confirm that yet

I can firmly deny it, given that it's used by both the rpcap server ***AND*** client code - and it's intended to be available to any future remote-capture clients, for other protocols, in libpcap.

Perhaps there needs to be a built-but-not-installed non-shared library, containing code used both by libpcap and rpcapd, and:

- the libpcap shared library is linked with that non-shared library, linking its object files into that library;
- the libpcap non-shared library includes the object files that are also in that other non-shared library;
- rpcapd is linked with that non-shared library, because the common code isn't part of the libpcap API.

gvanem:
> By the way: Can you have a look at the -d option?

With `rpcapd.exe -d`, I'm getting `ERROR_FAILED_SERVICE_CONTROLLER_CONNECT` which [MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/ms686324(v=vs.85).aspx) explains with some incomprehensible mumble.

guyharris:
The check that's supposed to be done isn't "is this thread us?", it's "was a thread created?"  That check should be done with a separate variable that's initially "false", set to "true" when a thread is created, and set to "false" when a thread is destroyed (cancelled).

I've done that in a16ca2786d56bff6cab41dd1542a387c77bb7370.

sgeto:
Thanks. I would not have got to this any time this week...

guyharris:
> fileconf.obj : error LNK2019: unresolved external symbol _fopen_safe referenced in function _fileconf_read

That should be fixed in ee4e10fc6a5a9f97438284758c02e64131b03a01 (although it may get warnings from MSVC about using`fopen()`, which should be fixed by using `pcap_fopen()` instead).

sgeto:
> Perhaps there needs to be a built-but-not-installed non-shared library, containing code used both by libpcap and rpcapd

This or maybe remove `sockutils.c` and `pcap-rpcap.c` from pcap's and rpcapd's sources and make their header files (`sockutils.h` and pcap-rpcap.h) include them. So `#include <sockutils.c>` at the end of `sockutils.h`.

I just tried it and it works, but it will need some extensive guarding or somethings like `/FORCE:MULTIPLE` on MSVC.
A lot a work still so I'm not sure if your idea may be quicker to implement.

@gvanem 

> But shouldn't the rpcapd service be created first? (via CreateService() etc.).

I think you're right. That's missing

PR #661, opened 2017-11-13
Instruct static linking on Windows with PCAP_STATIC_LIB

sgeto:
When linking statically with pcap variables/functions can not be decorated/marked as imported (__declspec(dllimport). 
To prevent this a users can (and should) define PCAP_STATIC_LIB when building their applications statically.

### DIFF
--- a/pcap/funcattrs.h
+++ b/pcap/funcattrs.h
@@ -59,6 +59,11 @@
      * API.
      */
     #define PCAP_API_DEF	__declspec(dllexport)
+     /*
+     * Define PCAP_STATIC_LIB when linking your application statically.
+     */
+  #elif defined(PCAP_STATIC_LIB)
+    #define PCAP_API_DEF
   #else
     #define PCAP_API_DEF	__declspec(dllimport)
   #endif


guyharris:
> users can (and should) define PCAP_STATIC_LIB when building their applications statically.

Are you saying that somebody building an *executable* that uses libpcap must define `PCAP_STATIC_LIB` when building *the executable*?  That sounds like a requirement that should not be inflicted on users; do they *really* have to do that, on Windows, for *every* library with which they build, if the library is offered both as a static and a dynamic library?

guyharris:
No, the right way to do it is the way zlib does it; if you're using the library rather than building it, then, *if* `PCAP_DLL` is defined, the API functions are declared with `__declspec(dllimport)`, otherwise they're declared without `__declspec`.  That way, if you don't define `PCAP_DLL`, your build will work with a DLL or with a static library, with a per-call performance hit if you're using the DLL, and if you *do* define `PCAP_DLL`, your build will work with a DLL, without the per-call performance hit, but won't link with the static library.

I.e., the *default* build behavior works with either library; only if you explicitly say "I'm building with the export library and will be running with the DLL", by defining `PCAP_DLL`, do you end up with optimized-but-DLL-only code.

sgeto:
> the right way to do it is the way zlib does it

Well, yes. Totally agree there actually.
It never seemed necessary for pcap to not decorate symbols with __declspec(dllimport) on windows since winpcap was only build (and offered) as a shared library.

The possibility of offering static linkage now and by default (after so many years without it) can be tricky.
I think this should probably be eased in. Mostly to prevent winpcap aware builds from breaking.
And this was essentially what this PR was all about. To keep things as they were, but give application implementor a way to to link statically.

I understand the rejection. It looked a bit goofy. 
I would personally like to see pcap taking the same road on windows as most other high profile projects.

guyharris:
> It never seemed necessary for pcap to not decorate symbols with __declspec(dllimport) on windows since winpcap was only build (and offered) as a shared library.

WinPcap didn't decorate them.  The only reason *to* decorate function symbols is to optimize procedure calls to routines in the DLL.  You *have* to decorate data symbols, but that's a good reason not to export them - we don't officially export any in the header files, although there are some that are unofficially "exported" for use by older code that declared them in the code; anybody doing that is on their own trying to make it work on Windows.

> The possibility of offering static linkage now and by default (after so many years without it) can be tricky.

Presumably by "by default" you mean "*installing* a static library by default", not "*linking* with a static library by default"; we're not going to do that on *any* platform.

> I would personally like to see pcap taking the same road on windows as most other high profile projects.

So what do they do with DLLs and static libraries?  The only problem I see is that the import library and the static library would have the same name by default.

sgeto:
On windows, if BUILDING_PCAP was not true, function symbols were decorated/marked as imported (__declspec(dllimport), which was preventing static linkage.
So even if a user had/build a static lib of libpcap, they wouldn't be able to link their apps with it unless they patched the corresponding line in funcattrs.h to not mark symbols as imported.
Static linking was therefore never _really_ available.

RPCAPD and the tests/samples linked statically because BUILDING_PCAP was defined and because cmake is weird (in that it compensated for it somehow).

I just went through winpcap's repo and saw that funcattrs.h and the BUILDING_PCAP macro are missing. This must have been added after the project became inactive. So I guess static linkage was possible with winpcap, but their sdk never provided a static lib and the project files in the repo only allowed building a dll (by using a .def file and e..g defining _USRDLL and LIBPCAP_EXPORTS).
Again, static linking also not being _really_ available.

With today's changes function symbols are not decorated/marked as imported by default if BUILDING_PCAP is not true, allowing to link statically with libpcap out of the box (again, or for the first time, depending on how you look at it).

This is how most other projects do it with DLLs and static libraries. So it's nice. Definitely better than my proposal. I was just thinking the transition may be too fast, but... nah.

sgeto:
> The only problem I see is that the import library and the static library would have the same name by default.

We actually had a discussion about this a while ago.
MinGW uses a well established library naming convention:

```
[Ali:/mingw64/lib] $ ls -lAh libcrypto.*
-rw-r--r-- 1 Ali Ali 3.6M Jun  8 23:47 libcrypto.a
-rw-r--r-- 1 Ali Ali 2.3M Jun  8 23:46 libcrypto.dll.a
[Ali:/mingw64/lib] $ ls -lAh libpcap.*
-rw-r--r-- 1 Ali Ali 1.1M Nov 22 04:42 libpcap.a
-rw-r--r-- 1 Ali Ali  73K Nov 22 04:40 libpcap.dll.a
[Ali:/mingw64/lib] $
```
Import libraries use a .dll.a suffix, which gcc/ld/libtool ect recognise/honour/whatever.

with MSVC it's get a bit confusing. There is no agreed convention. Here's openssl again:
tree /A /F
|   4758cca.lib
|   aep.lib
|   atalla.lib
|   capi.lib
|   chil.lib
|   cswift.lib
|   gmp.lib
|   gost.lib
|   libeay32.lib
|   nuron.lib
|   padlock.lib
|   ssleay32.lib
|   sureware.lib
|   ubsec.lib
|
\---VC
    |   libeay32MD.lib
    |   libeay32MDd.lib
    |   libeay32MT.lib
    |   libeay32MTd.lib
    |   ssleay32MD.lib
    |   ssleay32MDd.lib
    |   ssleay32MT.lib
    |   ssleay32MTd.lib
    |
    \---static
            libeay32MD.lib
            libeay32MDd.lib
            libeay32MT.lib
            libeay32MTd.lib
            ssleay32MD.lib
            ssleay32MDd.lib
            ssleay32MT.lib
            ssleay32MTd.lib

Debug and release builds shouldn't be linked. And so shouldn't different runtimes. So some devs add suffixes to distinguish between them.
Openssl, boost, microsoft and few other also add the a lib prefix to their static lib.
libpcap should probably do the same. So the import library should be called pcap.lib (that's already the case) and the static lib should be renamed from pcap_static.lib to libpcap.lib. You know, to be able to add add them fancy suffixes. And because it looks better. :-D

guyharris:
`BUILDING_PCAP` no longer has anything to do with `__declspec`; it currently *ONLY* affects `pcap_fopen_offline_with_tstamp_precision()` and `pcap_fopen_offline()`, which are greasy hacks to work around Microsoft's C library changing in binary-incompatible forms from compiler release to compiler release (until the Universal CRT) and thus causing `FILE` structures to potentially be different between the version of MSVCRT with which libpcap is built and the version of MSVCRT with which a program using libpcap is built.  See the comment in `pcap/pcap.h`.  It's defined *only* when building a source file that's part of the libpcap library; it's not set when building rpcapd or the test programs, and nobody writing code that uses libpcap should ever define it.

What we do now in `pcap/funcattrs.h` with `PCAP_API_DEF` and `PCAP_API` is controlled by `pcap_EXPORTS`, which is defined by CMake if we're building a libpcap library source file *AND* we're building the file for a DLL (and should *not* be defined by anybody compiling code that uses libpcap), and by `PCAP_DLL`, which should be defined by a *user* of libpcap when building code that uses libpcap *if and only if* their code will *only* be dynamically linked with libpcap and they want the small performance improvement you get from declaring with `__declspec(dllimport)`.

guyharris:
So we'd need to build *multiple* versions of the library, for different versions of the Microsoft C runtime (threaded vs. unthreaded and debug vs. release)?  And would we need to build *another* one for MinGW?

sgeto:
> So we'd need to build multiple versions of the library

No. It should be fine the way it is. User can choose whether they need debug builds and/or the type of run-time when building libcap with cmake. Provided that they know what look out for.


What could be done to make things more obvious, is it to suffix the resulting libraries based on the users choice.
So, for example, if the used chooses to build with with a dynamic rt the result library could be named libpcapMT.lib and/or if they choose to get a debug version then also add -d.

No need for any of this for MinGW. Mixing release and debug build (and runtimes) is fine there.

guyharris:
> No. It should be fine the way it is. User can choose whether they need debug builds and/or the type of run-time when building libcap with cmake.

Most users shouldn't *be* building libpcap; if they have to build libpcap in order to use it in a threaded program or use it in an unthreaded program, we've failed.

sgeto:
I think the default when creating a new project with visual studio (or cmake) is to use the dynamic runtime (MD), whereas the default in libpcap is to use the static one (MT).
MSVC will complain if you would to link these two. Maybe making the dynamic runtime the default again is a good idea.

sgeto:
> Most users shouldn't be building libpcap; if they have to build libpcap in order to use it in a threaded program or use it in an unthreaded program, we've failed.

Devs (who provide binary releases build with MSVC) should definitely decide up front what path their program is taking. You, as the maintainer, should only make these options available. No worries.

guyharris:
> Devs (who provide binary releases build with MSVC) should definitely decide up front what path their program is taking. 

You say "program", so presumably this is referring to programs using libpcap, not libpcap-for-Windows-plus-any-necessary-driver packages such as WInPcap or Npcap.  Is that the case?

If so, those devs probably don't want to be bothered with building libpcap from source; they would probably just want to be able to either package with WinPcap/Npcap or have the end user download and install WinPcap/Npcap (if, for example, the Npcap license makes packaging with Npcap difficult).

> You, as the maintainer, should only make these options available.

Making the options available to application developers means making them available *in binary form*, or making it possible for *pcap developers to provide them in binary form.  They might want to provide a *pcap SDK that offers enough libraries to permit the application developer to build a threaded or a non-threaded program using *pcap, and to make both debug and release builds of the application, so that might mean that libpcap needs to make it possible to build multiple versions of pcap.dll, pcap.lib, and libpcap.lib.  Do we need to do that?

(And I'm not convinced yet that there are no UN\*Xes that also may require multiple versions of libpcap, given that some of them have compiler options, for the OS vendor's C compiler, that should be used for multi-threaded code.)

gvanem:
I have only 1 comment.<br>
The comparison to zlib is a bit odd. The `zlib1.dll`  is linked using a `win32/zlib.def`. But `wpcap.dll` doesn't. If it would, it will mean more work maintaining a `Win32/wpcap.def` file.

Then again, the slight decrease in performance (due to the thunks) as described [here](https://msdn.microsoft.com/en-us/library/zw3za17w.aspx) is neglible.

guyharris:
> The `zlib1.dll` is linked using a `win32/zlib.def`. But `wpcap.dll` doesn't.

What difference does that make?  We're using `PCAP_API`, which expands to `__declspec(dllexport)` when building libpcap, rather than a .def file, to control which routines are exported from the dynamically-linked library; zlib, for some unknown reason, appears to do *both*.

"The way zlib does it" refers to the way the headers work *when included by programs that use it*, rather than to the way they work *when building the library*, i.e. don't use `__declspec(dllimport)` unless the developers of the code *using* libpcap explicitly request it by defining `PCAP_DLL`.  WinPcap didn't use `__declspec(dllimport)` at all; I'd originally made libpcap *always* use it, but now I changed it to do so only if `PCAP_DLL` is defined.

sgeto:
> You say "program", so presumably this is referring to programs using libpcap

Yes.
I'd say that WInPcap or Npcap may not be suitable for every developer. libpcap came a long way and the old libraries that come with WInPcap don't have these fixes/features. 
Npcap (as their name and license suggests) is optimized to just have the things needed to get the windows port of nmap going. It is build directly from libpcap's sources, I get that, but only with the feature _they_ need, not with all things libpcap can do on windows. These are good enough for most developers/programs. Others prefer building libpcap directly from the source and only use WInPcap or Npcap's driver/dll.

> Making the options available to application developers means making them available in binary form, or making it possible for *pcap developers to provide them in binary form. They might want to provide a *pcap SDK that offers enough libraries to permit the application developer to build a threaded or a non-threaded program using *pcap, and to make both debug and release builds of the application, so that might mean that libpcap needs to make it possible to build multiple versions of pcap.dll, pcap.lib, and libpcap.lib. Do we need to do that?

Need may be a strong word but it would be nice if the build system alone could assist in achieving this. (Without the need of scripts, patches etc)

I was actually thinking on doing that. To provide unofficial (possibly bleeding-edge) binaries of libpcap (and tcpdump) for windows. Maybe on a website or via my nuget feed. I am thinking to build packet dll separately and then bundle it together with _this_ libpcap.
(The Npcap license doesn't mind that. It only forbids you from distributing copies of their signed packages/installer)

guyharris:
> Npcap (as their name and license suggests) is optimized to just have the things needed to get the windows port of nmap going.

[citation needed], as they have [a comparison](https://nmap.org/npcap/vs-winpcap.html) that seems to be intended to indicate why Npcap should be considered a replacement for WinPcap for everything, not just for map.

> It is build directly from libpcap's sources, I get that, but only with the feature they need, not with all things libpcap can do on windows.

Precisely what features are missing?  (Other than "they don't have a release with the bleeding-edge tip-of-the-master-branch libpcap", but that applies to most Linux distributions, the \*BSDs, macOS, etc., etc., etc..)

sgeto:
First I didn't say that Npcap can not be a replacement for WinPcap or that it's only meant be used with nmap.

Their installer places its files in %systemroot%\system32\Npcap, which no program intended to work with Winpcap is by default aware of. I think they call this "Npcap native mode". Applications written with Winpcap in mind would need to be patched to look there first.  This "Npcap native mode" doesn't even follow the standard dll search path. Npcap suggests to use SetDllDirectory() to alter the standard dll search path to look for Npcap first. Only a handful of apps I know of are set up to look there **first**. Heavy emphasis on that. If an app finds the old winpcap first (and it will since it's in dll search path) it will use it. And:

> In python, the DLLs located in %systemroot%\system32 are pre-loaded, and I don't think that there is a way to avoid this behavior. This means that SetDllDirectory is useless, as we cannot unload the already loaded DLL.

taken from https://github.com/secdev/scapy/pull/524

The installer offers a so-called "compatibility mode" (which installs things the same way winpcap did). 
The latest version of their installer now also tries to compensate for this design flaw by offering (or maybe forcing.. not sure) users to uninstall winpcap _and_ win10pcap _before_ installing Npcap (compatibility mode or otherwise).

So if I had Npcap (native mode) installed and would build libpcap from _this_ repo and link it with my app, or the rpcapd, or even the tests, and then try to run it it would fail because not even the official libpcap is set up look for Npcap's packet.dll in that exotic location.

Don't want to repeat myself, but:
Probably at some point, the winpcap project shipping a copy of wpcap.dll along with what it takes to make it perform its magic on windows (driver and dll aka _dependencies_) may have made sense, because it wasn't possible to build it directly from lippcap's sources.
Ncap still doing so is... well... nice I guess, 
That's not to confuse with "Npcap is it, and without it we're doomed". 
Without the packet dll and driver we may be, but I don't hear anyone complaining about those that come with the standard winpcap installer.

> Precisely what features are missing?

Not sure if we're talking about the development files (headers and libraries) of the stuff for users (dll, driver etc). And I think you mean Npcap so

The wpdpack/npcap sdk lacks everything that has anything to with MinGW and Cygwin. there are locale bpf definitions in [Packet32.h](https://github.com/nmap/npcap/blob/master/Common/Packet32.h#L150) that are conflicting with those in pcap/bpf, See https://github.com/nmap/nmap/issues/1032 and other things I don't remember right now...

Their wpcap,dll has no support dbus or netmap (anything using file descriptors really. (not because it can't be done, but because it's not needed), has no debugging symbols, no static library, it's build with the VS 2013 solution in libpcap\Win32\Prj\wpcap.sln (because, and I quote: "I use VS 2013 to build DLLs because Nmap is also built with VS 2013". Convenient so both can use the same runtime, rather than using the highest available).

guyharris:
> Their wpcap,dll has no support dbus or netmap

The underlying problem here is that libpcap is currently a monolithic library; if your libpcap isn't built with support for XXX, the XXX suppliers can't provide an add-on for the existing libpcap, they'd have to provide their own version of libpcap.

If it weren't for the fact that programs doing packet capture require elevated privileges on most platforms, I'd just do that with plugins, but if libpcap were to support plugins, it'd have to be *really* careful about what plugins it trusts.

(I also would like to do some form of privilege separation so that, for example, libpcap wouldn't need elevated privileges to open BPF devices/PF_PACKET sockets/etc., it'd talk to a privileged program that does the opening and hands the FD to libpcap over a UNIX-domain socket.  That way the only privilege that a program using libpcap has is that it has a handle that lets it capture traffic (and possibly inject traffic).  That'd probably then require some other way to limit who gets to talk to that privileged program, however.)

sgeto:
Plugins sound pretty cool actually

> If it weren't for the fact that programs doing packet capture require elevated privileges on most platforms, I'd just do that with plugins,

No sure I see the connection. Are you concerted about plugins with malicious intent?

guyharris:
> Are you concerted about plugins with malicious intent?

Yes.  Perhaps allowing plugins *only* to be loaded from the directory in which the library is installed (or a plugin subdirectory of that directory, protected the same way that the installation directory is protected) would be sufficient, in that if somebody could replace or otherwise hack the libpcap shared library itself, you already have a problem - letting them install a plugin just makes it easier to force libpcap to do whatever you want.

sgeto:
> somebody could replace or otherwise hack the libpcap shared library itself

Especially If libpcap is installed at system level and that somebody has/gained the privilege to replace it, they could do anything. Installing a bad plugin or not wouldn't make matters worse, because the whole system is already compromised.

PR #682, opened 2018-03-19
Tame -Wall on MSVC and improve Clang/C2 support

sgeto:
MSVC's -Wall has more in common with clang's -Weverthing than gcc's -Wall. The result is very overwhelming (3000+ lines of warnings). 
From now on, if building with MSVC *and* if a a .devel is found *and* if MSVC is not using clang as it's front end, increase the warning level to 4. Else do the regular  check_and_add_compiler_option() routine.

It is probably also worth noticing that cmake's compiler identification string for LLVM/Clang is *not* true when using the Clang/C2 toolset. Only something like `if(MSVC AND NOT ${CMAKE_C_COMPILER} MATCHES "clang*")` or `if(CMAKE_GENERATOR_TOOLSET STREQUAL "v140_clang_3_7")` can detect it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1557,15 +1557,19 @@ endmacro()
 set(C_ADDITIONAL_FLAGS "")
 if(EXISTS ${CMAKE_SOURCE_DIR}/.devel)
     #
-    # -W options, used by GCC and compilers that try to behave like it.
-    #
-    check_and_add_compiler_option(-Wall)
-    check_and_add_compiler_option(-Wsign-compare)
-    check_and_add_compiler_option(-Wmissing-prototypes)
-    check_and_add_compiler_option(-Wstrict-prototypes)
-    check_and_add_compiler_option(-Wshadow)
-    check_and_add_compiler_option(-Wdeclaration-after-statement)
-    check_and_add_compiler_option(-Wused-but-marked-unused)
+    # -W options, used by GCC and compilers that try to behave like it including
+    # the Clang/C2 toolsets (which is disguises as MSVC).
+    if(MSVC AND NOT ${CMAKE_C_COMPILER} MATCHES "clang*")
+        check_and_add_compiler_option(-W4)
+    else()
+        check_and_add_compiler_option(-Wall)
+        check_and_add_compiler_option(-Wsign-compare)
+        check_and_add_compiler_option(-Wmissing-prototypes)
+        check_and_add_compiler_option(-Wstrict-prototypes)
+        check_and_add_compiler_option(-Wshadow)
+        check_and_add_compiler_option(-Wdeclaration-after-statement)
+        check_and_add_compiler_option(-Wused-but-marked-unused)
+    endif()
 endif()
 
 #
--- a/optimize.c
+++ b/optimize.c
@@ -71,7 +71,10 @@ int pcap_optimizer_debug;
    * _BitScanForward().
    */
 #include <intrin.h>
+
+#ifndef __clang__
 #pragma intrinsic(_BitScanForward)
+#endif
 
 static __forceinline int
 lowest_set_bit(int mask)


CMakeLists.txt
```diff
@@ -1557,15 +1557,20 @@ endmacro()
...
-    check_and_add_compiler_option(-Wmissing-prototypes)
-    check_and_add_compiler_option(-Wstrict-prototypes)
-    check_and_add_compiler_option(-Wshadow)
-    check_and_add_compiler_option(-Wdeclaration-after-statement)
-    check_and_add_compiler_option(-Wused-but-marked-unused)
+    # -W options, used by GCC and compilers that try to behave like it including
+    # the Clang/C2 toolsets (which is disguises as MSVC).
+    if(MSVC AND NOT ${CMAKE_C_COMPILER} MATCHES "clang*")
+        check_and_add_compiler_option(-W4)
+        check_and_add_compiler_option(-WD4996)
```
sgeto: Woupsi...

guyharris:
> MSVC's -Wall

MSVC (as in "Microsoft's C compiler", as in "a compiler that does not use the clang front end or the LLVM back end") recognizes a `-Wall` option, so that `check_and_add_compiler_option()` adds it to the compiler option list?

guyharris:
The underlying problem with "MSVC/clang" (as in "the clang front end and the Microsoft C2 back end") is that

> Since Visual Studio 2015 Update 1, Microsoft added Clang as an official alternative frontend to its C2 backend code generator. However, this doesn't play well with CMake, yet, mainly since CMake doesn't recognize the alternative platform toolset v140_clang_3_7 correctly, and fails at defining some flags (exceptions and debugging format) and directories correctly.

as per [this note from the Boost developers](https://github.com/boostorg/hana/wiki/Setting-up-Clang-on-Windows).

So is the `-Wall` issue an issue with Real MSVC, with Microsoft's front end, or is it only an issue with "MSVC/clang"?

sgeto:
> MSVC (as in "Microsoft's C compiler", as in "a compiler that does not use the clang front end or the LLVM back end") recognizes a -Wall option, so that check_and_add_compiler_option() adds it to the compiler option list?

yes. cl.exe has /Wall

sgeto:
> So is the -Wall issue an issue with Real MSVC, with Microsoft's front end, or is it only an issue with "MSVC/clang"?

It's an issue with real MSVC. It's not really an issue. It just gives too many warnings to see the big picture.

sgeto:
I've been using the newer v141_clang_c2 toolchain for a few weeks now and I must disagree with the note.
cmake and/or Microsoft have come a long way since Visual Studio 2015 Update 1 and it actually plays very well with cmake. One just need to keep in mind that as far as cmake is concerned, Clang/C2 is MSVC, not Clang/LLVM

I can add a build matrix to appveyor if you're interested.

guyharris:
> One just need to keep in mind that as far as cmake is concerned, Clang/C2 is MSVC, not Clang/LLVM

But it's neither - it's Clang/C2, rather than {Microsoft's C front end}/L2; if CMake reports them both as "MSVC", then one might say CMake doesn't work as well as it should with Clang/C2.

So is the "warn about "all"" behavior of Clang/C2 the same as that of Clang/LLVM, the same as that of MSVC, or different from both?

sgeto:
> But it's neither - it's Clang/C2, rather than {Microsoft's C front end}/L2; if CMake reports them both as "MSVC", then one might say CMake doesn't work as well as it should with Clang/C2.

If it were to report it as Clang then one might also assume that it is the full fletched LLVM toolchain when all you have is the clang compiler. 
I'd say it's more MSVC than LLVM but I guess it depends on how you look at it.

> So is the "warn about "all"" behavior of Clang/C2 the same as that of Clang/LLVM, the same as that of MSVC, or different from both?

It's the same as that of (the windows port of) Clang/LLVM

guyharris:
Well, having provoked a build with `/Wall`, some of them might actually be worth trying to fix, e.g. figuring out what's MSVCese for `__attribute__(unused)`, so we can use `_U_` to suppress uninteresting "unused parameter" warnings and get those warnings when they're real.  "'N' bytes padding after data member 'XXX'", however, is noise. and that seems to be about 99% of the warnings, so just suppressing C4820 might make a big difference.

PR #683, opened 2018-03-22
Prevent SIGSEGV

ardikars:
char errbuf[255];
pcap_t *pcap = pcap_open_dead(1, 1);
pcap_setnonblock(pcap, 1, errbuf); // SIGSEGV
pcap_close(pcap);


guyharris:
Fixed in 4a16d675abb3a5bc561d2dea4255a0ffd7b62777 by setting more function pointers to point to routines that return errors.

PR #684, opened 2018-04-01
Rename SOCK_ASSERT to SOCK_MESSAGE

jmayer:
SOCK_ASSERT was always called with a true expression ("1") thus only ever executing the message part.
Rename the macro, remove the assert part and remove the "1" parameter.


guyharris:
Maybe it should be called `SOCK_DEBUG_MESSAGE()`, given that it's only used for debugging messages.

PR #685, opened 2018-04-02
Declare some variables static (found via -Wmissing-variable-declarations)

jmayer:
Question: What is the preferred way to get rid of the fowllowing warnings, where the comments state that this is intentional:

pcap.c:180:19: warning: no previous extern declaration for non-static variable 'pcap_version' [-Wmissing-variable-declarations]
nametoaddr.c:549:28: warning: no previous extern declaration for non-static variable 'eproto_db' [-Wmissing-variable-declarations]

- Create a pcap-deprecated.h file and add it there
- Add DIAG_OFF/ON_MISSING_STATIC macros to diag-control.h
- Add generic DIAG_OFF/ON_WARNING macros to diag-control.h
- Use #pragma clang directly in the code
- Don't enable the warning at all

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1674,6 +1674,7 @@ if(EXISTS ${CMAKE_SOURCE_DIR}/.devel OR EXISTS ${CMAKE_BINARY_DIR}/.devel)
         check_and_add_compiler_option(-Wmissing-noreturn)
         # Warns about safeguards added in case the enums are extended
         # check_and_add_compiler_option(-Wcovered-switch-default)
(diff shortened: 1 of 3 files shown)


guyharris:
> What is the preferred way to get rid of the fowllowing warnings, where the comments state that this is intentional:

Another possibility would be to put an external declaration in the .c file, if that's sufficient to squelch the warning.

guyharris:
> Another possibility would be to put an external declaration in the .c file, if that's sufficient to squelch the warning.

I've done that in e54ab809ed63e243c53b15ac3b1f40adfe333cc1, with an accompanying addition to the comment, saying why we're doing that.

PR #686, opened 2018-04-04
Trivial cleanups to bpf_filter.c

jmayer:
- Remove all code when this file is used to compile a kernel module,
  i.e. inside KERNEL or _KERNEL (not done anymore)
- Remove all(?) helper stuff that was only needed for KERNEL stuff
- Fix an unused warning when VLAN_TAG matching is not done via aux_data
- Expand macro ANCILLARY - it didn't help readability
- Assume that SKF_AD_VLAN_TAG exists if SKF_AD_VLAN_TAG_PRESENT exists

### DIFF
--- a/bpf/net/bpf_filter.c
+++ b/bpf/net/bpf_filter.c
@@ -44,36 +44,17 @@
 
 #include <pcap/pcap-inttypes.h>
 #include "pcap-types.h"
+#include "varattrs.h"	// for _U_
 
 #ifndef _WIN32
 #include <sys/param.h>
 #include <sys/types.h>
 #include <sys/time.h>
-
-#if defined(sun) && (defined(__SVR4) || defined(__svr4__))
-#define SOLARIS
-#else
-#undef SOLARIS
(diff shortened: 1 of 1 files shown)


jmayer:
Would removing SUNOS4/ (sunos kernel modules) make sense as well?

guyharris:
> Would removing SUNOS4/ (sunos kernel modules) make sense as well?

At this point I suspect nobody other than hobbyists have machines capable of running SunOS versions prior to 5.x (and x is probably at least 8 at this point...), so, unless somebody protests that they really *need* binary modules for 4.x and before, we can probably get rid of them.

jmayer:
Anything I can do to get this accepted? Split into a series of smaller patches perhaps?

guyharris:
Perhaps this should be split into 1) changes to remove the kernel-mode support and 2) changes to get clean up the ancillary-data code?

jmayer:
Will split into two dedicated branches

PR #691, opened 2018-04-17
Make sure we don't overflow the bids array when running a binary that…

jmayer:
… was compiled with BDEBUG.

Fix a few compile warnings that pop up only when BDEBUG is defined.


guyharris:
Fixed with bounds checks for all references to `bids[]`, and just not reading or writing the array if the index is out of bounds, in 9bac75ac0222df31e29ecacf2da423702b3e5c3d.

PR #693, opened 2018-04-20
Add a minimalistic manpage for rpcapd

jmayer:
Very barebones content and install process - but you have to start somewhere


sgeto:
> you have to start somewhere

[Here ](https://www.winpcap.org/docs/docs_412/html/group__remote.html) maybe?

PR #694, opened 2018-04-20
cross compile for android error duo to the command line is too long

uscq:
When compiling it use the android NDK standalone toolchains, the `CC` command will be long enough with its extra options. After all of this, the command line may be too long to the environment (like Windows).

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -400,7 +400,7 @@ shared: libpcap.$(DYEXT)
 libpcap.so: $(OBJ)
 	@rm -f $@
 	VER=`cat $(srcdir)/VERSION`; \
-	MAJOR_VER=`sed 's/\([0-9][0-9]*\)\..*/\1/' $(srcdir)/VERSION`; \
+	MAJOR_VER=`sed 's/\([0-9][0-9]*\)\..*/\1/' $(srcdir)/VERSION`
 	@V_SHLIB_CMD@ $(LDFLAGS) @V_SHLIB_OPT@ @V_SONAME_OPT@$@.$$MAJOR_VER \
 	    -o $@.$$VER $(OBJ) $(ADDLOBJS) $(LIBS)
 


jmayer:
I just looked at the error message:

-- Installing: /tmp/bin/rpcapd
CMake Error at rpcapd/cmake_install.cmake:60 (file):
  file cannot create directory: /man8.  Maybe need administrative privileges.
Call Stack (most recent call first):
  cmake_install.cmake:435 (include)


make: *** [install] Error 1

So for some reason on your platform (or in your cross compile environment) the variable CMAKE_INSTALL_MANDIR seems to be undefined or empty

jmayer:
Just created a pull request (#695) with a fix.

guyharris:
> Just created a pull request (#695) with a fix.

A fix for this *particular* problem, or a fix for the problem with `rpcapd/CMakeLists.txt` that's causing Travis CMake builds with remote capture enabled to fail?

I need to test this change (not your change) to see whether it breaks building and installing shared libraries on systems that use `.so` as the shared-library suffix (meaning "all UN\*Xes except for macOS and 32-bit HP-UX").

guyharris:
> I need to test this change (not your change) to see whether it breaks building and installing shared libraries on systems that use `.so` as the shared-library suffix

It does.  On my Ubuntu 16.04 virtual machine, that change causes it to build `libpcap.so.` rather than `libpcap.so.`*version*.

uscq:
Maybe I had a configure issue, but before changing this, I got the following error:

VER=`cat ./VERSION`; \
        MAJOR_VER=`sed 's/\([0-9][0-9]*\)\..*/\1/' ./VERSION`; \
        aarch64-linux-android-gcc -fPIE -pie -std=gnu99  -shared -Wl,-soname,libpcap.so.$MAJOR_VER \
            -o libpcap.so.$VER pcap-linux.o pcap-usb-linux.o pcap-netfilter-linux.o fad-gifc.o pcap.o gencode.o optimize.o nametoaddr.o etherent.o fmtutils.o savefile.o sf-pcap.o sf-pcapng.o pcap-common.o bpf_image.o bpf_dump.o  scanner.o grammar.o bpf_filter.o
: Invalid argumentb-standalone-toolchain/bin/../lib/gcc/aarch64-linux-android/4.9.x/../../../../aarch64-linux-android/bin/ld.exe: cannot open output file libpcap.so.1.9.0-PRE-GIT
collect2.exe: error: ld returned 1 exit status
make: *** [libpcap.so] Error 1

PR #692, opened 2018-04-19
Remove kernel code

jmayer:

PR #701, opened 2018-04-26
Remove kernel code

jmayer:
Redo #692 - I failed to push the resolved merge conflict to GitHub


bpf/net/bpf_filter.c
```diff
@@ -623,19 +437,6 @@ bpf_validate(const struct bpf_insn *f, int len)
 			case BPF_ABS:
 			case BPF_IND:
 			case BPF_MSH:
-				/*
```
guyharris: I'd put the comment about there being no maximum packet data size back, to indicate why we're not checking the offset field in those instructions.

jmayer:
No idea what I do wrong, but after updating the patch I again couldn't push it. Continued in #702

PR #709, opened 2018-05-02
Fix warning when compiling with BDEBUG defined

jmayer:
warning: no previous extern declaration for non-static variable 'dflag' [-Wmissing-variable-declarations]

### DIFF
--- a/grammar.y
+++ b/grammar.y
@@ -65,6 +65,10 @@ struct rtentry;
 #include <arpa/inet.h>
 #endif /* _WIN32 */
 
+#ifdef YYDEBUG
+extern int pcap_debug;
+#endif
+
 #include <stdio.h>
 
 #include "diag-control.h"
--- a/missing/getopt.h
+++ b/missing/getopt.h
@@ -2,6 +2,6 @@
  * Header for the getopt() we supply if the platform doesn't supply it.
  */
 extern char *optarg;			/* getopt(3) external variables */
-extern int optind, opterr, optopt;
+extern int optind, opterr, optopt, optreset;
 
 extern int getopt(int nargc, char * const *nargv, const char *ostr);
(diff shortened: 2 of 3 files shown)


gvanem:
This is with gcc v6? I get such warnings with clang-cl v6 too. E.g. when I build with `-DBDEBUG`, I also set `-DYYDEBUG=1` causing:
```bash
grammar.c(1698,5):  warning: no previous extern declaration for non-static variable 'pcap_debug' [-Wmissing-variable-declarations]
int yydebug;
    ^
grammar.c(71,25):  note: expanded from macro 'yydebug'
#define yydebug         pcap_debug
                        ^
```
And (MSVC/clang-cl only):
```bash
missing/getopt.c(47,2):  warning: no previous extern declaration for non-static variable 'optreset' [-Wmissing-variable-declarations]
        optreset;               /* reset getopt */
        ^
```
The latter is easily fixed by:
```diff
--- a/missing/getopt.h 2015-11-02 09:59:52
+++ b/missing/getopt.h 2018-05-02 09:48:25
@@ -2,6 +2,6 @@
  * Header for the getopt() we supply if the platform doesn't supply it.
  */
 extern char *optarg;                   /* getopt(3) external variables */
-extern int optind, opterr, optopt;
+extern int optind, opterr, optreset, optopt;

 extern int getopt(int nargc, char * const *nargv, const char *ostr);
```
But the warning in `grammar.y` would need a patch to `pcap-int.h`. Maybe:
```diff
--- a/pcap-int.h 2018-05-01 09:50:39
+++ b/pcap-int.h 2018-05-02 09:55:46
@@ -496,6 +496,10 @@

 int    pcap_strcasecmp(const char *, const char *);

+#ifdef YYDEBUG
+extern int pcap_debug;
+#endif
+
 #ifdef __cplusplus
 }
 #endif
```

guyharris:
I've checked in Gisle's patches, which should address both of these issues.  (His change to `missing/config.h` is the same as the one in this pull request, except for the order of the variables being declared.)

guyharris:
(Plus some additional changes.)

PR #710, opened 2018-05-02
Add -Wformat-nonliteral and fix most warnings that come up

jmayer:
The remaining ocurrences are genuine and intentional.
I have no idea why a "const char *" doesn't qualify but I replaced it with a #define

warning: format string is not a string literal [-Wformat-nonliteral]

### DIFF
--- a/optimize.c
+++ b/optimize.c
@@ -2187,7 +2187,7 @@ convert_code_r(compiler_state_t *cstate, conv_state_t *conv_state,
 	    {
 		u_int i;
 		int jt, jf;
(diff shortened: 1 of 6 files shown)


optimize.c
```diff
@@ -2138,48 +2138,48 @@ convert_code_r(compiler_state_t *cstate, conv_state_t *conv_state,
 	    {
 		u_int i;
 		int jt, jf;
-		const char *ljerr = "%s for block-local relative jump: off=%d";
```
guyharris: What happens with

    static const char ljerr[] = "%s for block-local relative jump: off=%d";

PR #712, opened 2018-05-04
Warning unreachable code

jmayer:
Testing and request for feedback for now


jmayer:
Is there a way to access the generated grammar.c file of the AppVeyor build?

guyharris:
> Is there a way to access the generated grammar.c file of the AppVeyor build?

No easy one.

Note also that, at least with Bison, the canned code generated by Bison include some unreachable code in the parser function itself, and there's no obvious way to turn the "unreachable code" warnings off in the canned code and back on for generated code, so I don't see a way to get rid of the warnings from code completely generated by Bison but allow them for code that we wrote, so the Big Hammer - which is now hammering those warnings down with MSVC as well - may be the best we can do.

guyharris:
Also, `$$` is a YACC token, processed by YACC/Bison/Berkeley YACC, so you can't use it in a `#define`; doing

    #define PFVAR_RET $$ =

and

    pfvar:	  PF_IFNAME ID		{ PFVAR_RET gen_pf_ifname(cstate, $2); }

causes compile errors.

PR #713, opened 2018-05-04
Fix getopt.c(85): warning C4706: assignment within conditional expres…

jmayer:
…sion


jmayer:
Didn't work - next attempt later

PR #714, opened 2018-05-04
Fix getopt.c(85): warning C4706: assignment within conditional expres…

jmayer:
…sion

Not 100% identical but not too wastful either


guyharris:
I did this a bit differently in 9c87d9d5f086597ba064b2b0ac4b89644065b9f5.

PR #716, opened 2018-05-07
Expand "non-promisc" packet filter to include edge cases

dmiller-nmap:
First reported in 2013 by Sebastian Walther:
https://www.winpcap.org/pipermail/winpcap-bugs/2013-May/001620.html

The hardware packet filters for non-promisc mode on Windows were based
on an incomplete understanding. Npcap developers reviewing MS
documentation agree with Sebastian's assessment and propose the combined
packet filter of NDIS_PACKET_TYPE_ALL_LOCAL | NDIS_PACKET_TYPE_DIRECTED
| NDIS_PACKET_TYPE_BROADCAST | NDIS_PACKET_TYPE_MULTICAST for the
inverse of "promiscuous mode" on Windows.

Packet filter reference:
https://docs.microsoft.com/en-us/windows-hardware/drivers/network/oid-gen-current-packet-filter

### DIFF
--- a/pcap-npf.c
+++ b/pcap-npf.c
@@ -1034,7 +1034,17 @@ pcap_activate_npf(pcap_t *p)
 	}
 	else
 	{
-		if (PacketSetHwFilter(pw->adapter,NDIS_PACKET_TYPE_ALL_LOCAL) == FALSE)
+		/* NDIS_PACKET_TYPE_ALL_LOCAL selects "All packets sent by installed
+		 * protocols and all packets indicated by the NIC" but if no protocol
+		 * drivers (like TCP/IP) are installed, NDIS_PACKET_TYPE_DIRECTED,
+		 * NDIS_PACKET_TYPE_BROADCAST, and NDIS_PACKET_TYPE_MULTICAST are needed to
+		 * capture incoming frames.
+		 */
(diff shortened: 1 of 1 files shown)


guyharris:
So the issue is that

> All packets sent by installed protocols and all packets indicated by the NIC that is identified by a given NdisBindingHandle .

means that the NIC won't indicate (received) packets if they're for a disabled protocol?

dmiller-nmap:
@guyharris That's my understanding. In both the reported case linked here and the one privately reported by a Npcap licensee, the software using libpcap via wpcap.dll is implementing a userspace network stack, and they want to completely disable Windows' own networking, but doing so means the only way they can see packets is by enabling promiscuous mode. I don't see why that would really be a problem in that case, but it does create a disparity between Linux and Windows in their treatment of non-promisc pcap handles.

guyharris:
If you want all packets sent by the machine on which you're capturing, all packets sent directly to the interface on which you're capturing (e.g., sent to its MAC address if it's on a network with MAC addresses), multicast to a group to which that interface belongs, or broadcast, and *don't* need packets that are sent directly to some other machine on the network from some other machine on the network, you shouldn't have to enable promiscuous mode.

PR #718, opened 2018-05-16
Dag status API

sfd:
Link DAG Configuration and Status API library for pcap-dag.

Allows access to statistics such as stream drop, and configuration such as load balancing in future for FANOUT, IF_CONNECTION_STATUS support etc.

The 32-bit DAG stream drop counter provides a better ps_drop value than integrating 16-bit lctr, which may saturate under heavy load.


guyharris:
So this is intended to require that libdagconf is present, and fail if it can't be found?

PR #719, opened 2018-05-19
Add captured direction in pcap_pkthdr

teknoraver:


guyharris:
This changes a structure in ways that break the ABI; programs using `pcap_next()` that were built with a version of libpcap without the added member of `struct pcap_pkthdr` will, if they call `pcap_next()` when running with a version of libpcap *with* that member, get data following the structure overwritten.

teknoraver:
I need this feature to show the direction in tcpdump:

the-tcpdump-group/tcpdump#678

How can we do it without breaking the ABI?

teknoraver:
I solved it in a differente way, I allow to force a cooked capture as the datalink headers often provides this information.

Please see  #784

PR #721, opened 2018-05-24
Add SSL option for RPCAP TCP sockets

rixed:


guyharris:
Please don't change the indentation.

sslutils.c
```diff
@@ -0,0 +1,196 @@
...
+static int ssl_init_once(int is_server, char *errbuf, size_t errbuflen)
+{
+  static int inited = 0;
+  if (inited) return 0;
+
+  SSL_library_init();
+  SSL_load_error_strings();
+  OpenSSL_add_ssl_algorithms();
+
+  SSL_METHOD const *meth = SSLv23_method();
```
gvanem: Data declarations after code needs a C99-compiler. Don't think that's a requirement in libpcap.

rpcapd/rpcapd.c
```diff
@@ -194,7 +198,13 @@ int main(int argc, char *argv[])
 	mainhints.ai_socktype = SOCK_STREAM;
 
 	// Getting the proper command line options
-	while ((retval = getopt(argc, argv, "b:dhip:4l:na:s:f:v")) != -1)
```
guyharris: Or

    #ifdef HAVE_OPENSSL
        #define SSL_OPTS  "SK:C:"
    #else
        #define SSL_CLOPTS
    #endif

    #define CLOPTS "b:dhip:4l:na:s:f:v" SSL_CLOPTS

and then use `CLOPTS` as the third argument to `getopt()`.

gvanem: Switch to [optparse()](https://github.com/skeeto/optparse) instead? One single re-entrant file.

guyharris:
Why is there a "client must connect with TLS" flag in the "start capture" reply?  TLS should be established before you even try to log in to the server, so the password doesn't go over the wire in clear text; by the time you've gotten to starting a capture, you've either already negotiated TLS, in which case you should use TLS if the packets will be arriving over TCP and DTLS if the packets will be arriving over UDP, or you haven't negotiated TLS, in which case either you or the server probably don't support TLS and you won't be able to use TLS.

(Adding a flag to that message either means 1) revising the protocol, increasing the protocol version number, with the older version not having the flag field and the newer version having it, and using the negotiation mechanism to figure out which version to use or 2) interpreting the padding field in that message as a reserved field, and then un-reserving it and making it a flag field.

Also, using C bitfields in a structure representing an over-the-wire message is a Very Bad Idea, given that C makes no guarantees about byte order or other implementation details that matter for data exchanged between programs.)

sslutils.c
guyharris: We *do* require compiler support for at least some C99 features, including that one; we now require the same level of C99 support as Wireshark does:

> In general, not all C99 features can be used since some C compilers used to compile Wireshark, such as Microsoft's C compiler, don't support all C99 features.  The C99 features that can be used are:
> 
>  - flexible array members
>  - compound literals
>  - designated initializers
>  - "//" comments
>  - mixed declarations and code
>  - new block scopes for selection and iteration statements (that is, declaring the type in a for-loop like: for (int i = 0; i < n; i++) ;)
>  - macros with a variable number of arguments (variadic macros)
>  - trailing comma in enum declarations
>  - inline functions (guaranteed only by use of glib.h)

with "only by use of glib.h" replaced by "only by use of portability.h".

Our configure script doesn't enforce C99 support the way the Wireshark one does (in pre-3.0 releases; 3.0 requires CMake and doesn't have a configure script), by failing if it doesn't get a C99 compiler, but it should.

What I'm not sure of is why we're getting build errors with autotools *even though we're asking for a C99 compiler in the configure script*  and *even though it enables -std=gnu99* - perhaps the compiler is old enough that even gnu99 doesn't support mixed declarations and code.  I'll look into that.

rpcapd/rpcapd.c
guyharris: That might be worth considering, but it's orthogonal to adding SSL and to the various ways of making the availability of SSL-specific flags conditional.

sslutils.c
guyharris: I've removed the -s flag from the make command, so it should print the commands being used at build time.

guyharris:
Please rebase your change to the current contents of the master branch.

sslutils.c
guyharris: We're getting those build errors *because we're asking for them* - we're building with `-Wdeclaration-after-statement`

rixed:
> TLS should be established before you even try to log in to the server, so the password doesn't go over the wire in clear text;

Ideally yes, but I wanted first to encrypt the forwarded traffic, thus the focus on the "data socket".
Now that it seems using SSL with rpcap would be a welcomed addition, I will also similarly encrypt the "control socket".

TLS for the control socket can be done in several ways though:

1. `rpcapd -S` could wait for a TLS handshake right after the connection establishment of the control socket, but then clients must already know they have to issue this handshake, preferably without adding a command line parameter to every programs using libpcap. So I suggest we add a "rpcaps://" scheme so that libpcap can deal with this on its own. The "rpcaps://" scheme would _fail_ if the server does not run with -S, or the client libpcap is not compiled with openssl support.

2. Or `rpcapd -S` can send a special message before the authentication to request the client to establish TLS, aka a STARTTLS command. I regard opportunistic TLS like this as an ugly hack that makes the protocol more complicated to implement and debug, but the advantage is that older clients can still interoperate (it's not always an advantage though. One of the thing we really want to avoid is any kind of fallback to non-encrypted forward once rpcapd has been started with -S).

I strongly favor the first alternative and will work on that.

Re. DTLS, I was unaware openssl supported DTLS, but it does so I will give it a go.

Re. the ssl bit flag: Of course, having a specific scheme makes this controversial bit in the start-capture message useless and I will therefore remove it. But I wanted to acknowledge that yes indeed you are right and a bit field was a very bad idea given bit ordering within the chosen storage unit is indeed implementation specified.

Rebasing and reindenting will follow.

rixed:
Note regarding authentication: In the back of my mind, I'm of the opinion that password based authentication should be replaced by a client-certificate check, but it would require more code and configuration that I'm able to contribute ATM (on rpcapd side, a list of trusted client identities and on the client side dealing with pkcs11 to get the user cert)

guyharris:
> In the back of my mind, I'm of the opinion that password based authentication should be replaced/completed by a client-certificate check

That'd involve not only code changes but administrative changes, so "replaced", in the sense of "don't support password-based authentication at all", might be more than users would be willing to accept.  Offering it as an additional authentication flavor would work, however.

guyharris:
> TLS for the control socket can be done in several ways though:

...

> I strongly favor the first alternative and will work on that.

I've also thought TLS support would be a Very Good Thing, and was thinking of alternative 1) as well.

> So I suggest we add a "rpcaps://" scheme so that libpcap can deal with this on its own. The "rpcaps://" scheme would fail if the server does not run with -S, or the client libpcap is not compiled with openssl support.

Ultimately, I want to support remote capture using the create/activate APIs.  That would require changes in client applications - but that would have to happen *anyway*, even with `pcap_open()`, as opening a remote device isn't

    attempt to open device
    if that fails
        report an error and give up
    endif
    start capturing

it's

    start with null authentication
    loop doing
        attempt to open capture device
        if that succeeds
            we're done, break out of the loop
        endif
        if that fails with a non-authentication error
            give up
        endif
        tell the user about the authentication error
        ask for new credentials or a "forget it" indication
        if we get a "forget it" indication (control-C, "Cancel" button, etc.)
            give up
        endif
        set up to use the new credentials
     endloop

We could have "try to authenticate" also return a "would you like to upgrade to a more secure protocol?" indication and, in that case, have the loop enable TLS, or whatever, and continue.

That would allow application-side automatic upgrading to rpcaps, similar to browser-side upgrading to https.

mcr:
Guy Harris <notifications@github.com> wrote:
    > That'd involve not only code changes but administrative changes, so
    > "replaced", in the sense of "don't support password-based
    > authentication at all", might be more than users would be willing to
    > accept.  Offering it as an additional authentication flavor would work,
    > however.

I'd much rather we remove password authentication completely and insist that it
only run under sshd or stunnel.   I.e. stdin/out only, no socket code at all.

Perhaps that ship has sailed.

rixed:
I'm going to have a look at that windows compilation issue.

rixed:
Also, the active mode API is a bit of a mystery to me. To be honest the all protocol looks over complicated with the distinct control/data sockets, and probably it all stems from some use case for active mode that I don't really grasp.
Anyway, when is the client supposed to call pcap_remoteact_accept, exactly?
And in the SSL mode, where should we store the SSL handler? As a global parameter alongside sockmain, would that be ok?

rixed:
I'm afraid to enable SSL on the active API the client would have to instruct the lib if TLS is on or off, by adding a flag to pcap_remoteact_accept(). Would that be ok, or am I missing something more obvious?

guyharris:
> Also, the active mode API is a bit of a mystery to me. To be honest the all protocol looks over complicated with the distinct control/data sockets, and probably it all stems from some use case for active mode that I don't really grasp.

[This page in the WinPcap documentation](https://www.winpcap.org/docs/docs_40_2/html/group__remote.html) indicates the reason for active mode:

> The Active Mode is useful in case the remote daemon is behind a firewall and it cannot receive connections from the external world. In this case, the daemon can be configured to establish the connection to a given host, which will have been configured in order to *wait* for that connection. After establishing the connection, the protocol continues its job in almost the same way in both Active and Passive Mode.

So you start the capturing application, which presumably has to display a port number to which to connect, and then, on the remote machine, you run rpcapd in active mode, telling it to connect to that port and the capturing application's host.

The API for doing this is extremely ugly, and neither tcpdump nor Wireshark use it.  I think the [Analyzer sniffer](https://web.archive.org/web/20140325235147/http://analyzer.polito.it) from the Politecnico di Torino (the original WinPcap and rpcap/rpcapd developers) may have implemented it.

Active mode might work better with the create/activate API, as, instead of an API that starts with nothing and ends up with a connection or a failure, you could call `pcap_activate()` after specifying active mode, get back a "ready to wait for the connection" "error" indication, get the port number and report it, and then call `pcap_activate()` again to wait for the remote machine to connect.

pcap.c
```diff
@@ -1900,12 +1903,18 @@ pcap_parsesrcstr(const char *source, int *type, char *host, char *port,
 		return (0);
 	}
 
-	if (strcmp(scheme, "rpcap") == 0) {
+	int is_rpcap = 0;
+	if (strcmp(scheme, "rpcaps") == 0) {
+		is_rpcap = 1;
+		if (uses_ssl) *uses_ssl = 1;
+	}
+	is_rpcap |= strcmp(scheme, "rpcap") == 0;
```
guyharris: I might be inclined to do

    if (strcmp(scheme, "rpcaps") == 0) {
        is_rpcap = 1;
        if (uses_ssl)
            *uses_ssl = 1;
    } else if (strcmp(scheme, "rpcap") == 0)
        is_rpcap = 1;

as there's no need to check for "rpcap" if we already know it's "rpcaps".

rixed: > as there's no need to check for "rpcap" if we already know it's "rpcaps".

Ah true, I though `|=` would shortcut if `is_rpcap` were already true, but of course not as it's a _logical_ or. Will do.

infrastation:
I have briefly looked through the discussion and would like to make a point that if you are going to implement authentication that involves a password, that should not be a clear text password inside an encrypted channel, instead it should be something like CHAP, i.e. a method that can prove both parties know the same secret without disclosing the secret. Otherwise the connection will be subject to a man in the middle attack.

rixed:
By reading the log, I must confess I have no clue about the undefined symbols. Can someone with ssh access to the travis VM give me a hint?

sgeto:
You might want to add `$(OPENSSL_LIBS)` to `RPCAPD_LIBS` instead.
I also think you need to parentheses around $OPENSSL_CFLAGS and $OPENSSL_LIBS.
See `-lcrypt` in `configure.ac` for an example.

sgeto:
> ~~I also think you need parentheses around $OPENSSL_CFLAGS and $OPENSSL_LIBS.~~

Strike that.

rixed:
> You might want to add $(OPENSSL_LIBS) to RPCAPD_LIBS instead.

If I do that, openssl is no longer a dependency of libpcap.so and so clients (like tcpdump) fail as soon as they dereference any openssl symbol.

sgeto:
okay how about adding them to $LIBS then?

rixed:
The appVeyor build, which used to pass but now fails, apparently broke while downloading an external dependency (note to self: never depends on external systems for reproductible tests).
Will try to trigger the build again later so appease the gods of the green flag.

rixed:
I won't have time to add DTLS any time soon. I suggest this patch gets merged without, as it already encrypts the channel where the password is transmitted and the channel where the traffic is forwarded unless UDP is selected.

pcap-rpcap.c
```diff
@@ -2204,6 +2204,14 @@ pcap_t *pcap_open_rpcap(const char *source, int snaplen, int flags, int read_tim
 	pr = fp->priv;
 	pr->rmt_flags = flags;
 
+	// Also, SSL is not supported with UDP at the moment, so if the user
+	// asks for both we'd better bail out now:
+	if (pr->uses_ssl && (pr->rmt_flags & PCAP_OPENFLAG_DATATX_UDP)) {
+	  pcap_snprintf(errbuf, PCAP_ERRBUF_SIZE,
+			"SSL not supported with UDP forward of remote packets");
+		return NULL;
```
guyharris: You need to do a `pcap_close(fp)` before returning, so as not to leak the pcap_t.

pcap/pcap.h
```diff
@@ -837,7 +840,7 @@ PCAP_API pcap_t	*pcap_open(const char *source, int snaplen, int flags,
 PCAP_API int	pcap_createsrcstr(char *source, int type, const char *host,
 	    const char *port, const char *name, char *errbuf);
 PCAP_API int	pcap_parsesrcstr(const char *source, int *type, char *host,
-	    char *port, char *name, char *errbuf);
+	    char *port, char *name, unsigned char *uses_ssl, char *errbuf);
```
guyharris: `pcap_parsesrcstr()` is an API, so you can't change its calling sequence.

You would have to add a new `pcap_parsesrcstr_ex()` routine, with the new API, and make `pcap_parsesrcstr()` call it, passing a pointer to to internal variable for `uses_ssl`, and fail if that variable is non-zero when `pcap_parsesrcstr_ex()` returns.  Then have `pcap_open_rpcap()`, and other internal routines, call `pcap_parsesrcstr_ex()`.

guyharris: Also, there's no real benefit to having the Boolean flag be one byte long; the 3 bytes making it an `int` won't cost much, there might be padding causing it to eat all 4 bytes anyway, and some processors might do less internal work updating a 4-byte quantity than updating a 1-byte quantity.

pcap-rpcap.c
rixed: Fixed.

pcap/pcap.h
rixed: I reverted the change to the signature of `pcap_parsesrcstr()` and introduced a `pcap_parsesrcstr_ex()` that I call in the only place where it's really needed. For other call points, `pcap_parsesrcstr()` merely call the new function with a `NULL` as the ssl flag location.

> there's no real benefit to having the Boolean flag be one byte long

The actual benefit is that the only place where this pointer is needed is to initialize the flag within the `struct pcap_rpcap`, so making this function parameter a pointer to int instead would mean either introducing a temporary value (aka source code noise) or making the flag an int, but uint8 in `struct pcap_rpcap` seems the norm.

WDYT?

rixed:
What's the status of this?

rixed:
Rebased once again.
Could I have a yes or a no regarding that one?

infrastation:
I have briefly looked through the set of commits, there is a number of "fix this" and "fix that" standalone commits that it would help to fixup/squash into the changes they put right; also the huge "Rebasing" commit at the end does not look right (was there a rebase conflict?). I would suggest to comb through the set of commits in this pull request and make it look as clean and obvious as you can. This way it will be easier to spot any outstanding issues during the next review round, if that is your current goal.

rixed:
Ok, will do that.
Thank you for the answer.

rixed:
I believe the log is much easier to follow now.
I can of course squash it more if needed.

guyharris:
So why is `aclocal.m4` being renamed to `acinclude.m4`?  At least according to [section 3 "Making `configure` scripts" of the autoconf manual](https://www.gnu.org/software/autoconf/manual/autoconf.html#Making-configure-Scripts), "when using just Autoconf" `aclocal.m4` is an input file, and only "if you use Automake" is `acinclude.m4` used to generate `aclocal.m4`, and we're not using Automake.

guyharris:
Presumably the title should be changed to "Add SSL option for RPCAP TCP sockets", as 1) it does use TLS on the control connection and 2) it doesn't use DTLS if the captured packets are being sent over UDP.

pcap-rpcap.c
```diff
@@ -295,7 +303,7 @@ rpcap_deseraddr(struct rpcap_sockaddr *sockaddrin, struct sockaddr_storage **soc
 		if ((*sockaddrout) == NULL)
 		{
 			pcap_fmt_errmsg_for_errno(errbuf, PCAP_ERRBUF_SIZE,
-			    errno, "malloc() failed");
+				errno, "malloc() failed");
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -428,8 +444,8 @@ static int pcap_read_nocb_remote(pcap_t *p, struct pcap_pkthdr *pkt_header, u_ch
 	if (pr->rmt_flags & PCAP_OPENFLAG_DATATX_UDP)
 	{
 		/* Read the entire message from the network */
-		msglen = sock_recv_dgram(pr->rmt_sockdata, p->buffer,
-		    p->bufsize, p->errbuf, PCAP_ERRBUF_SIZE);
+		msglen = sock_recv_dgram(pr->rmt_sockdata, pr->data_ssl, p->buffer,
+			p->bufsize, p->errbuf, PCAP_ERRBUF_SIZE);
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -535,8 +549,8 @@ static int pcap_read_nocb_remote(pcap_t *p, struct pcap_pkthdr *pkt_header, u_ch
 	/*
 	 * Did the server specify the version we negotiated?
 	 */
-	if (rpcap_check_msg_ver(pr->rmt_sockdata, pr->protocol_version,
-	    header, p->errbuf) == -1)
+	if (rpcap_check_msg_ver(pr->rmt_sockdata, pr->data_ssl, pr->protocol_version,
+		header, p->errbuf) == -1)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -708,54 +722,68 @@ static void pcap_cleanup_rpcap(pcap_t *fp)
 	if (!active)
 	{
 		rpcap_createhdr(&header, pr->protocol_version,
-		    RPCAP_MSG_CLOSE, 0, 0);
+			RPCAP_MSG_CLOSE, 0, 0);
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -708,54 +722,68 @@ static void pcap_cleanup_rpcap(pcap_t *fp)
...
 
 		/*
 		 * Send the close request; don't report any errors, as
 		 * we're closing this pcap_t, and have no place to report
 		 * the error.  No reply is sent to this message.
 		 */
-		(void)sock_send(pr->rmt_sockctrl, (char *)&header,
-		    sizeof(struct rpcap_header), NULL, 0);
+		(void)sock_send(pr->rmt_sockctrl, pr->ctrl_ssl, (char *)&header,
+			sizeof(struct rpcap_header), NULL, 0);
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -708,54 +722,68 @@ static void pcap_cleanup_rpcap(pcap_t *fp)
...
 
 		/*
 		 * Send the end capture request; don't report any errors,
 		 * as we're closing this pcap_t, and have no place to
 		 * report the error.
 		 */
-		if (sock_send(pr->rmt_sockctrl, (char *)&header,
-		    sizeof(struct rpcap_header), NULL, 0) == 0)
+		if (sock_send(pr->rmt_sockctrl, pr->ctrl_ssl, (char *)&header,
+			sizeof(struct rpcap_header), NULL, 0) == 0)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -708,54 +722,68 @@ static void pcap_cleanup_rpcap(pcap_t *fp)
...
 			 * Wait for the answer; don't report any errors,
 			 * as we're closing this pcap_t, and have no
 			 * place to report the error.
 			 */
-			if (rpcap_process_msg_header(pr->rmt_sockctrl,
-			    pr->protocol_version, RPCAP_MSG_ENDCAP_REQ,
-			    &header, NULL) == 0)
+			if (rpcap_process_msg_header(pr->rmt_sockctrl, pr->ctrl_ssl,
+				pr->protocol_version, RPCAP_MSG_ENDCAP_REQ,
+				&header, NULL) == 0)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -708,54 +722,68 @@ static void pcap_cleanup_rpcap(pcap_t *fp)
...
 			{
-				(void)rpcap_discard(pr->rmt_sockctrl,
-				    header.plen, NULL);
+				(void)rpcap_discard(pr->rmt_sockctrl, pr->ctrl_ssl,
+					header.plen, NULL);
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -878,20 +906,20 @@ static struct pcap_stat *rpcap_stats_rpcap(pcap_t *p, struct pcap_stat *ps, int
 	    RPCAP_MSG_STATS_REQ, 0, 0);
 
 	/* Send the PCAP_STATS command */
-	if (sock_send(pr->rmt_sockctrl, (char *)&header,
-	    sizeof(struct rpcap_header), p->errbuf, PCAP_ERRBUF_SIZE) < 0)
+	if (sock_send(pr->rmt_sockctrl, pr->ctrl_ssl, (char *)&header,
+		sizeof(struct rpcap_header), p->errbuf, PCAP_ERRBUF_SIZE) < 0)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -878,20 +906,20 @@ static struct pcap_stat *rpcap_stats_rpcap(pcap_t *p, struct pcap_stat *ps, int
...
 		return NULL;		/* Unrecoverable network error */
 
 	/* Receive and process the reply message header. */
-	if (rpcap_process_msg_header(pr->rmt_sockctrl, pr->protocol_version,
-	    RPCAP_MSG_STATS_REQ, &header, p->errbuf) == -1)
+	if (rpcap_process_msg_header(pr->rmt_sockctrl, pr->ctrl_ssl, pr->protocol_version,
+		RPCAP_MSG_STATS_REQ, &header, p->errbuf) == -1)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -878,20 +906,20 @@ static struct pcap_stat *rpcap_stats_rpcap(pcap_t *p, struct pcap_stat *ps, int
...
 		return NULL;		/* Error */
 
 	plen = header.plen;
 
 	/* Read the reply body */
-	if (rpcap_recv(pr->rmt_sockctrl, (char *)&netstats,
-	    sizeof(struct rpcap_stats), &plen, p->errbuf) == -1)
+	if (rpcap_recv(pr->rmt_sockctrl, pr->ctrl_ssl, (char *)&netstats,
+		sizeof(struct rpcap_stats), &plen, p->errbuf) == -1)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -1177,19 +1209,19 @@ static int pcap_startcapture_remote(pcap_t *fp)
 	if (pcap_pack_bpffilter(fp, &sendbuf[sendbufidx], &sendbufidx, &fp->fcode))
 		goto error_nodiscard;
 
-	if (sock_send(pr->rmt_sockctrl, sendbuf, sendbufidx, fp->errbuf,
-	    PCAP_ERRBUF_SIZE) < 0)
+	if (sock_send(pr->rmt_sockctrl, pr->ctrl_ssl, sendbuf, sendbufidx, fp->errbuf,
+		PCAP_ERRBUF_SIZE) < 0)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -1177,19 +1209,19 @@ static int pcap_startcapture_remote(pcap_t *fp)
...
 		goto error_nodiscard;
 
 	/* Receive and process the reply message header. */
-	if (rpcap_process_msg_header(pr->rmt_sockctrl, pr->protocol_version,
-	    RPCAP_MSG_STARTCAP_REQ, &header, fp->errbuf) == -1)
+	if (rpcap_process_msg_header(pr->rmt_sockctrl, pr->ctrl_ssl, pr->protocol_version,
+		RPCAP_MSG_STARTCAP_REQ, &header, fp->errbuf) == -1)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -1177,19 +1209,19 @@ static int pcap_startcapture_remote(pcap_t *fp)
...
 		goto error_nodiscard;
 
 	plen = header.plen;
 
-	if (rpcap_recv(pr->rmt_sockctrl, (char *)&startcapreply,
-	    sizeof(struct rpcap_startcapreply), &plen, fp->errbuf) == -1)
+	if (rpcap_recv(pr->rmt_sockctrl, pr->ctrl_ssl, (char *)&startcapreply,
+		sizeof(struct rpcap_startcapreply), &plen, fp->errbuf) == -1)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -1514,19 +1567,19 @@ static int pcap_updatefilter_remote(pcap_t *fp, struct bpf_program *prog)
 	if (pcap_pack_bpffilter(fp, &sendbuf[sendbufidx], &sendbufidx, prog))
 		return -1;
 
-	if (sock_send(pr->rmt_sockctrl, sendbuf, sendbufidx, fp->errbuf,
-	    PCAP_ERRBUF_SIZE) < 0)
+	if (sock_send(pr->rmt_sockctrl, pr->ctrl_ssl, sendbuf, sendbufidx, fp->errbuf,
+		PCAP_ERRBUF_SIZE) < 0)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -1514,19 +1567,19 @@ static int pcap_updatefilter_remote(pcap_t *fp, struct bpf_program *prog)
...
 		return -1;
 
 	/* Receive and process the reply message header. */
-	if (rpcap_process_msg_header(pr->rmt_sockctrl, pr->protocol_version,
-	    RPCAP_MSG_UPDATEFILTER_REQ, &header, fp->errbuf) == -1)
+	if (rpcap_process_msg_header(pr->rmt_sockctrl, pr->ctrl_ssl, pr->protocol_version,
+		RPCAP_MSG_UPDATEFILTER_REQ, &header, fp->errbuf) == -1)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -1759,19 +1812,19 @@ static int pcap_setsampling_remote(pcap_t *fp)
 	sampling_pars->method = (uint8)fp->rmt_samp.method;
 	sampling_pars->value = (uint16)htonl(fp->rmt_samp.value);
 
-	if (sock_send(pr->rmt_sockctrl, sendbuf, sendbufidx, fp->errbuf,
-	    PCAP_ERRBUF_SIZE) < 0)
+	if (sock_send(pr->rmt_sockctrl, pr->ctrl_ssl, sendbuf, sendbufidx, fp->errbuf,
+		PCAP_ERRBUF_SIZE) < 0)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -1759,19 +1812,19 @@ static int pcap_setsampling_remote(pcap_t *fp)
...
 		return -1;
 
 	/* Receive and process the reply message header. */
-	if (rpcap_process_msg_header(pr->rmt_sockctrl, pr->protocol_version,
-	    RPCAP_MSG_SETSAMPLING_REQ, &header, fp->errbuf) == -1)
+	if (rpcap_process_msg_header(pr->rmt_sockctrl, pr->ctrl_ssl, pr->protocol_version,
+		RPCAP_MSG_SETSAMPLING_REQ, &header, fp->errbuf) == -1)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -1985,15 +2038,15 @@ static int rpcap_sendauth(SOCKET sock, uint8 *ver, struct pcap_rmtauth *auth, ch
...
 
 	/* Receive the reply */
-	if (rpcap_recv_msg_header(sock, &header, errbuf) == -1)
+	if (rpcap_recv_msg_header(sock, ssl, &header, errbuf) == -1)
 		return -1;
 
-	if (rpcap_check_msg_type(sock, RPCAP_MSG_AUTH_REQ, &header,
-	    &errcode, errbuf) == -1)
+	if (rpcap_check_msg_type(sock, ssl, RPCAP_MSG_AUTH_REQ, &header,
+		&errcode, errbuf) == -1)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -2261,29 +2338,30 @@ pcap_t *pcap_open_rpcap(const char *source, int snaplen, int flags, int read_tim
...
-	    PCAP_ERRBUF_SIZE) < 0)
+	if (sock_send(sockctrl, ssl, sendbuf, sendbufidx, errbuf,
+		PCAP_ERRBUF_SIZE) < 0)
 		goto error_nodiscard;
 
 	/* Receive and process the reply message header. */
-	if (rpcap_process_msg_header(sockctrl, protocol_version,
-	    RPCAP_MSG_OPEN_REQ, &header, errbuf) == -1)
+	if (rpcap_process_msg_header(sockctrl, ssl, protocol_version,
+		RPCAP_MSG_OPEN_REQ, &header, errbuf) == -1)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -2261,29 +2338,30 @@ pcap_t *pcap_open_rpcap(const char *source, int snaplen, int flags, int read_tim
...
 		goto error_nodiscard;
 	plen = header.plen;
 
 	/* Read the reply body */
-	if (rpcap_recv(sockctrl, (char *)&openreply,
-	    sizeof(struct rpcap_openreply), &plen, errbuf) == -1)
+	if (rpcap_recv(sockctrl, ssl, (char *)&openreply,
+		sizeof(struct rpcap_openreply), &plen, errbuf) == -1)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -2422,25 +2508,41 @@ pcap_findalldevs_ex_remote(char *source, struct pcap_rmtauth *auth, pcap_if_t **
...
 	}
 
 	/* RPCAP findalldevs command */
 	rpcap_createhdr(&header, protocol_version, RPCAP_MSG_FINDALLIF_REQ,
 	    0, 0);
 
-	if (sock_send(sockctrl, (char *)&header, sizeof(struct rpcap_header),
-	    errbuf, PCAP_ERRBUF_SIZE) < 0)
+	if (sock_send(sockctrl, ssl, (char *)&header, sizeof(struct rpcap_header),
+		errbuf, PCAP_ERRBUF_SIZE) < 0)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -2422,25 +2508,41 @@ pcap_findalldevs_ex_remote(char *source, struct pcap_rmtauth *auth, pcap_if_t **
...
 		goto error_nodiscard;
 
 	/* Receive and process the reply message header. */
-	if (rpcap_process_msg_header(sockctrl, protocol_version,
-	    RPCAP_MSG_FINDALLIF_REQ, &header, errbuf) == -1)
+	if (rpcap_process_msg_header(sockctrl, ssl, protocol_version,
+		RPCAP_MSG_FINDALLIF_REQ, &header, errbuf) == -1)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -2459,8 +2561,8 @@ pcap_findalldevs_ex_remote(char *source, struct pcap_rmtauth *auth, pcap_if_t **
 		tmpstring2[PCAP_BUF_SIZE] = 0;
 
 		/* receive the findalldevs structure from remote host */
-		if (rpcap_recv(sockctrl, (char *)&findalldevs_if,
-		    sizeof(struct rpcap_findalldevs_if), &plen, errbuf) == -1)
+		if (rpcap_recv(sockctrl, ssl, (char *)&findalldevs_if,
+			sizeof(struct rpcap_findalldevs_if), &plen, errbuf) == -1)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -2508,8 +2610,8 @@ pcap_findalldevs_ex_remote(char *source, struct pcap_rmtauth *auth, pcap_if_t **
 			}
 
 			/* Retrieve adapter name */
-			if (rpcap_recv(sockctrl, tmpstring,
-			    findalldevs_if.namelen, &plen, errbuf) == -1)
+			if (rpcap_recv(sockctrl, ssl, tmpstring,
+				findalldevs_if.namelen, &plen, errbuf) == -1)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -2541,8 +2643,8 @@ pcap_findalldevs_ex_remote(char *source, struct pcap_rmtauth *auth, pcap_if_t **
 			}
 
 			/* Retrieve adapter description */
-			if (rpcap_recv(sockctrl, tmpstring,
-			    findalldevs_if.desclen, &plen, errbuf) == -1)
+			if (rpcap_recv(sockctrl, ssl, tmpstring,
+				findalldevs_if.desclen, &plen, errbuf) == -1)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -2574,9 +2676,9 @@ pcap_findalldevs_ex_remote(char *source, struct pcap_rmtauth *auth, pcap_if_t **
 			struct rpcap_findalldevs_ifaddr ifaddr;
 
 			/* Retrieve the interface addresses */
-			if (rpcap_recv(sockctrl, (char *)&ifaddr,
-			    sizeof(struct rpcap_findalldevs_ifaddr),
-			    &plen, errbuf) == -1)
+			if (rpcap_recv(sockctrl, ssl, (char *)&ifaddr,
+				sizeof(struct rpcap_findalldevs_ifaddr),
+				&plen, errbuf) == -1)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -2882,21 +3020,27 @@ int pcap_remoteact_close(const char *host, char *errbuf)
...
 				 * just cleaning up.
 				 */
-				if (sock_send(temp->sockctrl,
-				    (char *)&header,
-				    sizeof(struct rpcap_header), errbuf,
-				    PCAP_ERRBUF_SIZE) < 0)
+				if (sock_send(temp->sockctrl, temp->ssl,
+					(char *)&header,
+					sizeof(struct rpcap_header), errbuf,
+					PCAP_ERRBUF_SIZE) < 0)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -2997,13 +3148,13 @@ int pcap_remoteact_list(char *hostlist, char sep, int size, char *errbuf)
...
+static int rpcap_recv_msg_header(SOCKET sock, SSL *ssl, struct rpcap_header *header, char *errbuf)
 {
 	int nrecv;
 
-	nrecv = sock_recv(sock, (char *) header, sizeof(struct rpcap_header),
-	    SOCK_RECEIVEALL_YES|SOCK_EOF_IS_ERROR, errbuf,
-	    PCAP_ERRBUF_SIZE);
+	nrecv = sock_recv(sock, ssl, (char *) header, sizeof(struct rpcap_header),
+		SOCK_RECEIVEALL_YES|SOCK_EOF_IS_ERROR, errbuf,
+		PCAP_ERRBUF_SIZE);
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -3131,14 +3282,14 @@ static int rpcap_process_msg_header(SOCKET sock, uint8 expected_ver, uint8 reque
...
+	if (rpcap_check_msg_ver(sock, ssl, expected_ver, header, errbuf) == -1)
 		return -1;
 
 	/*
 	 * Check the message type.
 	 */
-	return rpcap_check_msg_type(sock, request_type, header,
-	    &errcode, errbuf);
+	return rpcap_check_msg_type(sock, ssl, request_type, header,
+		&errcode, errbuf);
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -3160,8 +3311,8 @@ static int rpcap_recv(SOCKET sock, void *buffer, size_t toread, uint32 *plen, ch
 		pcap_snprintf(errbuf, PCAP_ERRBUF_SIZE, "Message payload is too short");
 		return -1;
 	}
-	nread = sock_recv(sock, buffer, toread,
-	    SOCK_RECEIVEALL_YES|SOCK_EOF_IS_ERROR, errbuf, PCAP_ERRBUF_SIZE);
+	nread = sock_recv(sock, ssl, buffer, toread,
+		SOCK_RECEIVEALL_YES|SOCK_EOF_IS_ERROR, errbuf, PCAP_ERRBUF_SIZE);
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -3183,9 +3334,9 @@ static void rpcap_msg_err(SOCKET sockctrl, uint32 plen, char *remote_errbuf)
 		 * Message is too long; just read as much of it as we
 		 * can into the buffer provided, and discard the rest.
 		 */
-		if (sock_recv(sockctrl, remote_errbuf, PCAP_ERRBUF_SIZE - 1,
-		    SOCK_RECEIVEALL_YES|SOCK_EOF_IS_ERROR, errbuf,
-		    PCAP_ERRBUF_SIZE) == -1)
+		if (sock_recv(sockctrl, ssl, remote_errbuf, PCAP_ERRBUF_SIZE - 1,
+			SOCK_RECEIVEALL_YES|SOCK_EOF_IS_ERROR, errbuf,
+			PCAP_ERRBUF_SIZE) == -1)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -3209,9 +3360,9 @@ static void rpcap_msg_err(SOCKET sockctrl, uint32 plen, char *remote_errbuf)
 	}
 	else
 	{
-		if (sock_recv(sockctrl, remote_errbuf, plen,
-		    SOCK_RECEIVEALL_YES|SOCK_EOF_IS_ERROR, errbuf,
-		    PCAP_ERRBUF_SIZE) == -1)
+		if (sock_recv(sockctrl, ssl, remote_errbuf, plen,
+			SOCK_RECEIVEALL_YES|SOCK_EOF_IS_ERROR, errbuf,
+			PCAP_ERRBUF_SIZE) == -1)
```
guyharris: Please don't change the indentation of existing code.

pcap-rpcap.c
```diff
@@ -3267,9 +3418,10 @@ static int rpcap_read_packet_msg(SOCKET sock, pcap_t *p, size_t size)
 		 * We haven't read all of the packet header yet.
 		 * Read what remains, which could be all of it.
 		 */
-		bytes_read = sock_recv(sock, bp, size - cc,
-		    SOCK_RECEIVEALL_NO|SOCK_EOF_IS_ERROR, p->errbuf,
-		    PCAP_ERRBUF_SIZE);
+		bytes_read = sock_recv(rp->rmt_sockdata, rp->data_ssl, bp, size - cc,
+			SOCK_RECEIVEALL_NO|SOCK_EOF_IS_ERROR, p->errbuf,
```
guyharris: Please don't change the indentation of existing code.

rpcapd/rpcapd.c
```diff
@@ -1190,8 +1266,8 @@ accept_connection(SOCKET listen_sock)
 		// This is passive mode, so we don't care whether we were
 		// told by the client to close.
 		//
-		(void)daemon_serviceloop(sockctrl, sockctrl, 0,
-		    nullAuthAllowed);
+		(void)daemon_serviceloop(sockctrl, sockctrl, ssl, 0,
+			nullAuthAllowed, uses_ssl);
```
guyharris: Please don't change the indentation of existing code.

rixed:
Regarding the renaming of aclocal to acinclude, I guess it's because I've used autoreconf (instead of autoconf). I fixed it, manually appending the PKG_* macros to the original aclocal.m4.

Regarding the issue with tabs (the *other* billion dollar mistake), I've rewritten the history as much as I could to use what I think is the proper style. Would you welcome a PR adding modelines for vim and emacs in every source file?

pcap-rpcap.c
```diff
@@ -715,47 +729,61 @@ static void pcap_cleanup_rpcap(pcap_t *fp)
...
 		 */
-		(void)sock_send(pr->rmt_sockctrl, (char *)&header,
+		(void)sock_send(pr->rmt_sockctrl, pr->ctrl_ssl, (char *)&header,
 		    sizeof(struct rpcap_header), NULL, 0);
 	}
 	else
 	{
 		rpcap_createhdr(&header, pr->protocol_version,
-		    RPCAP_MSG_ENDCAP_REQ, 0, 0);
+			RPCAP_MSG_ENDCAP_REQ, 0, 0);
```
guyharris: Another unnecessary change of indentation.

rpcapd/rpcapd.c
```diff
@@ -381,13 +421,24 @@ int main(int argc, char *argv[])
...
+#endif
 		//
 		// Handle this client.
 		// This is passive mode, so we don't care whether we were
 		// told by the client to close.
 		//
-		(void)daemon_serviceloop(sockctrl_in, sockctrl_out, 0,
-		    nullAuthAllowed);
+		(void)daemon_serviceloop(sockctrl_in, sockctrl_out, ssl, 0,
+			nullAuthAllowed, uses_ssl);
```
guyharris: Don't change the indentation, just add `uses_ssl` as an argument.

guyharris:
> Would you welcome a PR adding modelines for vim and emacs in every source file?

As a *separate* pull request, adding mostly-kernel-normal-form style, yes.

rixed:
Fixed (with a few others).
As far as my regexs can tell, others space/tab or curly braces style mismatches predate this PR.

rixed:
And rebased.

mcr:
Cedric Cellier <notifications@github.com> wrote:
    > ...and again.  Is this project accepting PR at all?

Quite a few.
We are still trying to dig out of a year-long CVE innudation.

rixed:
Understood; sorry for being impatient.

guyharris:
From a quick look at the documentation, there can be multiple connections for a given context, so I assume that this supports having multiple `pcap_t`s open to different servers.

What about the "TODO lock?" comment?  openssl/openssl#2165 seems to indicate that "An `SSL_CTX` may be used on multiple threads *provided it is not reconfigured*.", so perhaps a lock on the context isn't necessary.  (libpcap itself doesn't guarantee that a given `pcap_t` can be safely used by multiple threads, with the exception of one thread calling `pcap_breakloop()` on a `pcap_t` and another thread doing all the work on that `pcap_t`.)

rixed:
That comment mean that given a cursory look at the doc and a very limited experience with openssl, I think a lock is not required, but should one be needed in the future this is where to add it.

guyharris:
The conflicts appear to be due to merging your other pull request #722; you might want to rebase.

rixed:
True, and it was only half fixe in that branch.

guyharris:
> True, and it was only half fixe in that branch.

So is there anything left to fix in the master branch?  If so, that should be a separate pull request, so we can backport that to the 1.9 branch.

rixed:
> So is there anything left to fix in the master branch?

Not that I know about.

guyharris:
> I'm afraid to enable SSL on the active API the client would have to instruct the lib if TLS is on or off, by adding a flag to pcap_remoteact_accept(). Would that be ok, or am I missing something more obvious?

`pcap_remoteact_accept()` is a libpcap/WinPcap/Npcap API, so its signature should not be changed.  You could, however, add `pcap_remoteact_accept_ex()`, with the extra argument, and have `pcap_remoteact_accept()` be a wrapper that calls it with the "this is SSL" flag being zero.

guyharris:
> > I'm afraid to enable SSL on the active API the client would have to instruct the lib if TLS is on or off, by adding a flag to pcap_remoteact_accept(). Would that be ok, or am I missing something more obvious?
> 
> `pcap_remoteact_accept()` is a libpcap/WinPcap/Npcap API, so its signature should not be changed. You could, however, add `pcap_remoteact_accept_ex()`, with the extra argument, and have `pcap_remoteact_accept()` be a wrapper that calls it with the "this is SSL" flag being zero.

I've checked in a change to do that.

PR #722, opened 2018-05-25
Fix using uninitialised file descriptor?

rixed:
I think at this point pr->rmt_sockctrl is still uninitialised, and that we want to discard sockctrl instead.
Please check.

### DIFF
--- a/pcap-rpcap.c
+++ b/pcap-rpcap.c
@@ -2276,7 +2276,7 @@ pcap_t *pcap_open_rpcap(const char *source, int snaplen, int flags, int read_tim
 		goto error;
 
 	/* Discard the rest of the message, if there is any. */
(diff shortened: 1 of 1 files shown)


guyharris:
> I think at this point pr->rmt_sockctrl is still uninitialised, and that we want to discard sockctrl instead.

Yes, copy-and-pasteo.

The error case probably needs to do the same thing.

guyharris:
You can also remove the question mark from the commit message. :-)

rixed:
What's the status of this?

guyharris:
> What's the status of this?

You fixed the error case, but you haven't removed the question mark from the commit message. :-)

rixed:
I did actually, but the PR title is not always the same as the git commit message.
But the real question I think is how worthy is a bug fix compared to a typographic glitch?

PR #726, opened 2018-06-22
documentation: make it clear that stream passed to pcap_dump_fopen() is closed by pcap_dump_close()

anthonykirby:
It isn't stated in the documentation whether the FILE* passed to pcap_dump_fopen() will be closed by a subsequent pcap_dump_close().

This lack of clarity can cause the user to close the FILE* themselves after calling pcap_dump_close(), leading to SIGABRT on some platforms.


infrastation:
On the web-site.

Issue #127, opened 2013-04-15
Display/filter on packet direction and interface name


guyharris:
Converted from [SourceForge issue 2850109](http://sourceforge.net/support/tracker.php?aid=2850109), submitted by leonerd

When capturing from a single interface, libpcap only allows the EN10MB link type, which contains only the ethernet frame, with no context such as would provide packet direction. This makes it impossible to dump a single interface and know the direction.

When capturing from "-i any", libpcap on Linux provides LINUX_SLL, which seems to throw away such useful information as ifindex, making it impossible to know which of the many interfaces the packet was found on.

Please therefore provide some way to have all of this; if by no other mechanism, lets have LINUX_SLL2 which is a complete copy of the entire struct sockaddr_ll, which will therefore have the sll_pkttype (for direction) and sll_ifindex (for interface index) fields. Allow any capture on Linux, of a single interface or all, to switch to this frame type.

This would allow us to make tcpdump incantations such as the following, which right now are impossible:

  # Capture packets pertaining to the locally running SSH server, ignoring any locally running SSH clients
  tcpdump inbound tcp dest port 22 or outbound tcp source port 22 

  # I am a firewalling bridge, capture packets flowing in one direction only, to see what is being dropped
  tcpdump -i any inbound ifname "eth0" or outbound ifname "eth1"

Using this newfound information, it would then be possible for tcpdump's packet output to display it; I might suggest something like:
# tcpdump -n -i eth0 icmp

tcpdump: verbose output suppressed, use -v or -vv for full protocol decode
listening on eth0, link-type EN10MB (Ethernet), capture size 96 bytes
17:25:08.727472 Tx IP 192.168.2.221 > 192.168.2.1: ICMP echo request, id 62524, seq 3, length 64
17:25:08.727642 Rx IP 192.168.2.1 > 192.168.2.221: ICMP echo reply, id 62524, seq 3, length 64
# tcpdump -n -i any

tcpdump: verbose output suppressed, use -v or -vv for full protocol decode
listening on any, link-type LINUX_SLL (Linux cooked), capture size 96 bytes
17:26:42.944458 Rx(eth0) ARP, Request who-has 192.168.2.133 tell 192.168.2.135, length 46
17:26:43.105021 Rx(eth0) ARP, Request who-has 192.168.0.251 tell 192.168.0.1, length 46

leonerd:
[...]

PR #729, opened 2018-07-11
Add DLT_LINUX_SLL2 and LINKTYPE_LINUX_SLL2

pevik:
with linktype value 276.

This is a preparation for implementing Linux "cooked" capture
encapsulation v2.

See: GH the-tcpdump-group/libpcap#127

Signed-off-by: Petr Vorel <pvorel@suse.cz>

### DIFF
--- a/doc/DLT_ALLOCATE_HOWTO.md
+++ b/doc/DLT_ALLOCATE_HOWTO.md
@@ -26,10 +26,4 @@ DEVELOPER NOTES
 
 When allocating a new DLT_ value, a corresponding value needs to be
 added to pcap-common.c.
-It not necessary to copy to commonts from dlt.h to pcap-common.c.
-
-
-
-
-
-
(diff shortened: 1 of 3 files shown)


guyharris:
I don't like that header.  See my comment in the-tcpdump-group/tcpdump-htdocs#3; I prefer something like

    struct header {
    	uint16_t hatype;		/* link-layer address type */
    	uint8_t  pkttype;		/* packet type */
    	uint8_t  halen;			/* link-layer address length */
    	uint8_t  addr[SLL_ADDRLEN];	/* link-layer address */
    	int32_t  if_index;		/* 1-based interface index */
    	uint16_t hatype_specific;	/* dependent on sll3_hatype */
    	uint16_t protocol;		/* protocol */
    };

fxlb:
Perhaps this new header should be discussed on the mailing list for a possible `LINKTYPE_LINUX_SLL3/DLT_LINUX_SLL3`?

PR #741, opened 2018-08-06
Proper breakloop for linux on pcap_dispatch

### LINKED ISSUE #734: pcap_breakloop doesn't always interrupt pcap_dispatch
for example:

```
pcap_breakloop();
pcap_dispatch(...);
```

on a device that has no traffic will block until pcap_breakloop is later called from inside a signal handler.

One might argue that this is not normal usage but take for example the following case:

Thread 1 runs pcap_dispatch and some other things
Thread 2 does the following
1. set up some things to do after a `pcap_dispatch` interrupt
2. send a signal to the thread doing `pcap_dispatch`, in the signal handler call `pcap_breakloop`
3. set up some more work to do after a `pcap_dispatch` interrupt
4. same as 2

Now if 4 happens too fast after 2 and on Thread 1 we are not back into a `pcap_dispatch` blocking system call, then it will not be interrupted.

Problems seems to be that `pcap_dispatch` does a poll with an infinite timeout (at least on most linux versions). In this condition I propose pcap uses a non-blocking [eventfd](http://man7.org/linux/man-pages/man2/eventfd.2.html).

libpcap version:  a5341c3b81a4d7f00b5318c6d8fa8673eda81611
operating system: `Linux 4.17.3-1-ARCH #1 SMP PREEMPT Tue Jun 26 04:42:36 UTC 2018 x86_64 GNU/Linux`

Also reproduced on `Linux  4.9.0-6-amd64 #1 SMP Debian 4.9.88-1+deb9u1 (2018-05-07) x86_64 GNU/Linux`

No special configure options.

adrian-budau:
On platforms that support it use an eventfd to exit any polling.

This is meant to fix #734 . It does not currently handle `pcap_read_packet`, I could also add poll there so it wouldn't block. I probably didn't search enough, but I could not find any tests to run to check everything is in order, so any help in that direction would be helpful.

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -142,6 +142,11 @@
 #include "pcap/sll.h"
 #include "pcap/vlan.h"
 
+#ifdef HAVE_SYS_EVENTFD_H
+#include <sys/eventfd.h>
+#endif
+
+
 /*
  * If PF_PACKET is defined, we can use {SOCK_RAW,SOCK_DGRAM}/PF_PACKET
  * sockets rather than SOCK_PACKET sockets.
@@ -324,6 +329,10 @@ struct pcap_linux {
 	unsigned char *current_packet; /* Current packet within the TPACKET_V3 block. Move to next block if NULL. */
 	int packets_left; /* Unhandled packets left within the block from previous call to pcap_read_linux_mmap_v3 in case of TPACKET_V3. */
 #endif
+#ifdef HAVE_SYS_EVENTFD_H
+	int poll_breakloop_fd; /* fd to an eventfd to break from blocking operations */
+#endif
+
 };
 
 /*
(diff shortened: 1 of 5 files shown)


pcap-linux.c
```diff
@@ -142,6 +142,11 @@
 #include "pcap/sll.h"
 #include "pcap/vlan.h"
 
+#ifdef HAVE_SYS_EVENTFD_H
+#include <sys/eventfd.h>
+#endif
+
+
```
guyharris: Only one blank line is needed.

pcap.c
```diff
@@ -3604,6 +3611,13 @@ pcap_remove_from_pcaps_to_close(pcap_t *p)
 	}
 }
 
+void
+pcap_breakloop_common(pcap_t *p)
+{
+	p->break_loop = 1;
+}
+
+
```
guyharris: One blank line is sufficient here as well.

guyharris:
The check for `sys/eventfd.h` needs to be done in autoconf as well; I added that in commit dfce68d037bad5f57341db488392553cd479eaab.

guyharris:
> It does not currently handle `pcap_read_packet`, I could also add poll there so it wouldn't block.

Currently, I don't think that will ever be called on a system new enough to have eventfd, so it's probably not an issue.

However, a case could be made that immediate mode should use socket reads rather than TPACKET_V[12], in which case it would be an issue.

(Then again, a case could be made that most programs using immediate mode should instead use a library that provides a socket-like API for raw network I/O; that library wouldn't even necessarily use the same underlying mechanism as libpcap does, and might be able to use the same mechanism in a different fashion more appropriate for implementing a particular protocol, e.g. setting the protocol on a PF_PACKET socket so as to get only packets for that protocol.)

PR #748, opened 2018-08-24
Fuzzing libpcap

catenacyber:
One target for a pcap file
One target for a bpf filter

We should combine both, but I am not sure what is the best way to do this.

This is for oss-fuzz integration

### DIFF
--- a/testprogs/CMakeLists.txt
+++ b/testprogs/CMakeLists.txt
@@ -38,3 +38,5 @@ add_test_executable(threadsignaltest ${CMAKE_THREAD_LIBS_INIT})
 if(NOT WIN32)
   add_test_executable(valgrindtest)
 endif()
+
+add_subdirectory(fuzz)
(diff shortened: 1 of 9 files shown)


catenacyber:
I guess the Cmakelists.txt files can be improved so that we compile fuzz targets only when we need tests, and so that it works on windows (if I understand correctly why appveyor failed)

guyharris:
> Could you fix this please?

It appears that `testprogs/fuzz/fuzz_pcap.c` uses several UN\*X APIs, such as `open()` and `ftruncate()`, so, in order for it to work on Windows, it'd have to be ported; not including `unistd.h` when built on Windows would be part of that process (but obviously not all of it).

So the simplest fix is *not* to add the `fuzz` subdirectory on Windows.

catenacyber:
Thanks @guyharris I had come to the same conclusion. I will try a bit to make my code portable

catenacyber:
That should be good to go now :-)

infrastation:
Let's try to have these changes in the master branch to unblock OSS-Fuzz for libpcap (as discussed earlier). Team, if you have substantial objections to merging this pull request, please post your comments within few days. @catenacyber, please rebase this work on the current master branch (I have briefly checked and found a small rebase conflict) and make it one nice commit.

catenacyber:
@infrastation here is the rebase + commits squashing after a few days (and I had no conflict)

infrastation:
These changes are in the master branch now, thank you.

PR #749, opened 2018-08-25
Manpage formatting

dmiller-nmap:
All manpage references such as pcap_create(3PCAP) will now be formatted
with the identifier (e.g. "pcap_create") in **bold** and the section
name (e.g. "(3PCAP)") in roman (default) face. This is how most manpages
seem to be formatted and makes things more consistent.

I'm assuming due to some of the errors found that it is appropriate to submit a patch for the roff files directly, and that they are not generated from some other source. Do correct me if I am wrong, thanks.


guyharris:
> This is how most manpages seem to be formatted

On which operating system?  On macOS, whose man pages are probably pretty close to those of other BSD-flavored UN\*Xes, the open(2) and getpwent(3) man pages' SEE ALSO sections do *not* boldface the man page names, for example.

infrastation:
If it helps, boldfacing of man page names in SEE ALSO:
* happens on a 2018 Linux in `man 2 open` from glibc, and
* happens on Solaris 11 in `man fopen` from libc, but
* does not happen on Solaris 11 in `man open` from Tcl.

dmiller-nmap:
This keeps the SEE ALSO section in concordance with the manpage references added in a710ddc83c9ec174d417c695d5f0ada310ff9660, and also fixes some inconsistencies in how the section portion of the manpage reference was bolded.

infrastation:
For clarity, I do not know how many styles exist. If there is more than one style, I do not have a preferred one. That said, it would be nice to use one style/convention consistently in the man pages as it keeps maintenance simple.

dmiller-nmap:
For my part, I was looking to use https://github.com/bagder/roffit to
render these docs for Npcap, and it only links pages that are in bold.
Presumably you're using a different renderer on tcpdump.org that doesn't
have the same limitation.

infrastation:
Historically www.tcpdump.org had the HTML man pages generated with plain man2html on a very irregular basis. At some point I had scripted that process and started to add layers of man2html-specific post-processing to make the output HTML more useful for the web-site. Hence what you see in the browser now isn't exactly what man2html delivers (including hyperlinks to other man pages, see `regen_manpages.sh` in tcpdump-htdocs if interested).

Potentially it is possible to switch to a different HTML generator and to amend the post-processing as required. I remember trying some other tools (one definitely being mandoc) and not seeing better deliverables. Maybe I did not look well enough, in that case if somebody else is able to see the gain of a different HTML generator and deliver it without regressions, I would not get in the way of such a contribution.

PR #750, opened 2018-08-26
Ignore net only on root directory

orgads:
There are `/bpf/net` and `/Win32/Include/net` which are part of the repository.


infrastation:
As of release 1.9.0 libpcap has no `Include` in `Win32`.

orgads:
Sorry, I needed this for for an earlier version and didn't notice that these directories were removed.

PR #752, opened 2018-08-30
Fix multiple declaration of bpf_ functions

### LINKED ISSUE #751: Redudant declaration of bpf_filter and bpf_validate
- *version*: master (18ea67cd) exists in previous version, tried 1.7, 1.6 and 1.5 branches
- *OS*: archlinux (gcc 8.2.0), debian stretch (gcc 6.3.0)
- *flags*: No configure flags

## Issue Description

**NOTE**: This is only showing when using a locally built *libpcap* and not installing it on system, see explanations below.

The declarations of *bpf_filter()* and *bpf_validate()* exists in both [pcap/pcap.h](https://github.com/the-tcpdump-group/libpcap/blob/master/pcap/pcap.h#L516-L528) and [pcap/bpf.h](https://github.com/the-tcpdump-group/libpcap/blob/master/pcap/bpf.h#L257-L259)

So when building an object when using with *-Wredudant-decls* AND a local *libpcap* we will have multiple declarations of these functions and gcc will output a warning:

> ./pcap/pcap.h:524:18: warning: redundant redeclaration of ‘bpf_filter’ [-Wredundant-decls]

**WHY**: What actually happens is that when using an installed libpcap, the preprocessor will disable some of the warning, as explained [here](https://gcc.gnu.org/onlinedocs/cpp/Preprocessor-Output.html) on the '3' description:

> ‘3’ This indicates that the following text comes from a system header file, so certain warnings should be suppressed.

When generating a file including *pcap.h* from system the output of a gcc -E command will look like:

```
# 1 "/usr/include/pcap.h" 1 3 4
```

While the same with a local header will be:

```
# 1 "./pcap.h" 1
```

Just to let you know, I stumbled upon this as a project at  work needs *ENABLE_REMOTE*, so currently we build the lib and link to static lib locally.

## How to reproduce

Using an empty code including *pcap.h*:

```C
#include "pcap.h"

int main(void)
{
    return 0;
}
```

Compile it from a libpcap folder using local header, with the appropriate warning:

```
$ gcc -Wredundant-decls -I . -c /tmp/test.c -o test.o                                                                                                               
In file included from ./pcap.h:43,                                                                                                                                                                               
                 from /tmp/test.c:1:                                                                                                                                                                             
./pcap/pcap.h:524:18: warning: redundant redeclaration of ‘bpf_filter’ [-Wredundant-decls]                                                                                                                       
   PCAP_API u_int bpf_filter(const struct bpf_insn *, const u_char *, u_int, u_int);                                                                                                                             
                  ^~~~~~~~~~                                                                                                                                                                                     
In file included from ./pcap/pcap.h:88,                                                                                                                                                                          
                 from ./pcap.h:43,                                                                                                                                                                               
                 from /tmp/test.c:1:                                                                                                                                                                             
./pcap/bpf.h:258:16: note: previous declaration of ‘bpf_filter’ was here                                                                                                                                         
 PCAP_API u_int bpf_filter(const struct bpf_insn *, const u_char *, u_int, u_int);                                                                                                                               
                ^~~~~~~~~~                                                                                                                                                                                       
In file included from ./pcap.h:43,                                                                                                                                                                               
                 from /tmp/test.c:1:                                                                                                                                                                             
./pcap/pcap.h:526:14: warning: redundant redeclaration of ‘bpf_validate’ [-Wredundant-decls]                                                                                                                     
 PCAP_API int bpf_validate(const struct bpf_insn *f, int len);                                                                                                                                                   
              ^~~~~~~~~~~~                                                                                                                                                                                       
In file included from ./pcap/pcap.h:88,                                                                                                                                                                          
                 from ./pcap.h:43,                                                                                                                                                                               
                 from /tmp/test.c:1:                                                                                                                                                                             
./pcap/bpf.h:257:14: note: previous declaration of ‘bpf_validate’ was here                                                                                                                                       
 PCAP_API int bpf_validate(const struct bpf_insn *, int);
```

Using system header shows no warning:

```
$ gcc -Wredundant-decls -c /tmp/test.c -o test.o

$
```

## Additional notes

The preprocessor output:

```
$ gcc -Wredundant-decls -c /tmp/test.c -o test.e -E && grep pcap.h test.e | head -n 3                                                                             
# 1 "/usr/include/pcap.h" 1 3 4                                                                                                                                                                                  
# 43 "/usr/include/pcap.h" 3 4
# 1 "/usr/include/pcap/pcap.h" 1 3 4

$ gcc -Wredundant-decls -I . -c /tmp/test.c -o test.e -E && grep pcap.h test.e | head -n 3                                                                        
# 1 "./pcap.h" 1                                                                                                                                                                                                 
# 43 "./pcap.h"
# 1 "./pcap/pcap.h" 1
```

bleader:
This fixes #751.
I'm thinking this may be the proper way to do it, based  on my understanding. I tested a *tcpdump* compilation with this and it worked. If this isn't the right way to do it let me know.

### commit log
pcap.h will include pcap/pcap.h and pcap/bpf.h that both defines the
bpf_filter and bpf_validate functions.

In most cases this won't be an issue, most people are using the libpcap
as a system library, gcc preprocessor will therefore disable part of the
warnings according to the `3` in
https://gcc.gnu.org/onlinedocs/cpp/Preprocessor-Output.html

In case libpcap is included in a project that actually have more warning
(-Wredundant-decls), and also using local headers we end up having quite
a lot of warning.

### DIFF
--- a/pcap/bpf.h
+++ b/pcap/bpf.h
@@ -254,9 +254,20 @@ struct bpf_aux_data {
 #define BPF_STMT(code, k) { (u_short)(code), 0, 0, k }
 #define BPF_JUMP(code, k, jt, jf) { (u_short)(code), jt, jf, k }
 
-PCAP_API int bpf_validate(const struct bpf_insn *, int);
-PCAP_API u_int bpf_filter(const struct bpf_insn *, const u_char *, u_int, u_int);
+/*
+ * On at least some versions of NetBSD and QNX, we don't want to declare
+ * bpf_filter() here, as it's also be declared in <net/bpf.h>, with a
+ * different signature, but, on other BSD-flavored UN*Xes, it's not
+ * declared in <net/bpf.h>, so we *do* want to declare it here, so it's
+ * declared when we build pcap-bpf.c.
+ */
+#if !defined(__NetBSD__) && !defined(__QNX__)
+  PCAP_API u_int	bpf_filter(const struct bpf_insn *, const u_char *, u_int, u_int);
+#endif
 extern u_int bpf_filter_with_aux_data(const struct bpf_insn *, const u_char *, u_int, u_int, const struct bpf_aux_data *);
+PCAP_API int	bpf_validate(const struct bpf_insn *f, int len);
+PCAP_API char	*bpf_image(const struct bpf_insn *, int);
+PCAP_API void	bpf_dump(const struct bpf_program *, int);
 
 /*
  * Number of scratch memory words (for BPF_LD|BPF_MEM and BPF_ST).
--- a/pcap/pcap.h
+++ b/pcap/pcap.h
@@ -513,20 +513,6 @@ PCAP_API void	pcap_freealldevs(pcap_if_t *);
  */
 PCAP_API const char *pcap_lib_version(void);
 
-/*
- * On at least some versions of NetBSD and QNX, we don't want to declare
- * bpf_filter() here, as it's also be declared in <net/bpf.h>, with a
- * different signature, but, on other BSD-flavored UN*Xes, it's not
- * declared in <net/bpf.h>, so we *do* want to declare it here, so it's
- * declared when we build pcap-bpf.c.
- */
-#if !defined(__NetBSD__) && !defined(__QNX__)
-  PCAP_API u_int	bpf_filter(const struct bpf_insn *, const u_char *, u_int, u_int);
-#endif
-PCAP_API int	bpf_validate(const struct bpf_insn *f, int len);
-PCAP_API char	*bpf_image(const struct bpf_insn *, int);
-PCAP_API void	bpf_dump(const struct bpf_program *, int);
-
 #if defined(_WIN32)
 
   /*


bleader:
It looks like the failed test is the setup of the CI that fails, if I'm not mistaken.

infrastation:
After restarting the failed job the build has passed.

guyharris:
> It looks like the failed test is the setup of the CI that fails, if I'm not mistaken. Not sure if there's an easy way to retrigger a specific job.

If you go to the page for a Travis build, each of the jobs should have a restart button on the right-hand edge of the entry for the job; pushing that will restart the job.

guyharris:
> I'm thinking this may be the proper way to do it, based on my understanding. I tested a tcpdump compilation with this and it worked. If this isn't the right way to do it let me know.

This raises a bigger issue, namely that if `net/bpf.h` declares `bpf_filter()` one way and libpcap defines it another way, even `pcap-bpf.c` needs a declaration that matches how libpcap defines it, not how `net/bpf.h` (mistakenly) declares it.  ("Mistakenly" because it should *not* be declaring the *kernel's* version of `bpf_filter()` unless it's being used in a *kernel* build; other \*BSDs, and macOS, declare it only in kernel builds by testing for a `#define` such as `KERNEL` or `KERNEL_PRIVATE`, but NetBSD doesn't - it *should*, but it doesn't.)

So I have a bigger change that should address this; it renames the internal-to-pcap filtering routine as `pcap_filter()`, which is *not* exported from the library, and has `bpf_filter()` as a wrapper around it that *is* exported.  Everything inside libpcap should be using `pcap_filter()` (unless it uses `pcap_filter_with_aux_data()`), which will be declared by `pcap-int.h`.

bleader:
> If you go to the page for a Travis build, each of the jobs should have a restart button on the right-hand edge of the entry for the job; pushing that will restart the job.

I didn't find it, my understanding, is that this button is only accessible to people with write permission on the repository, the suggested way for others is to close and reopen the pull-request, which felt pretty ugly, so I didn't do it.

> This raises a bigger issue, namely that if net/bpf.h declares bpf_filter() one way and libpcap defines it another way, even pcap-bpf.c needs a declaration that matches how libpcap defines it, not how net/bpf.h (mistakenly) declares it.
> 
> So I have a bigger change that should address this; it renames the internal-to-pcap filtering routine as pcap_filter(), which is not exported from the library, and has bpf_filter() as a wrapper around it that is exported. Everything inside libpcap should be using pcap_filter() (unless it uses pcap_filter_with_aux_data()), which will be declared by pcap-int.h.

I agree, and the solution seems a good way to do it, not that you need my approval anyway, but as it is related, I felt like giving my opinion.

Thanks for the merge!

PR #754, opened 2018-09-03
Cleanly avoid warnings in grammar.y

### LINKED ISSUE #753: bison warning on grammar.y
- *version*: master (18ea67cd)  and previous versions, 1.7, 1.6, 1.5.
- *OS*: archlinux (bison 3.0.5), debian stretch (bison 3.0.4), debian jessie (bison 3.0.2)
- *flags*: no flags configured

**issue / reproduciton**:

When building:

    ./configure && make
    [...]
    bison -y -p pcap_ -o grammar.c -d grammar.y
    grammar.y: warning: 38 shift/reduce conflicts [-Wconflicts-sr]
    [...]

I understand you may not really care about this one, feel free to close if you won't want to bother with it.

I don't have a real solution beside ignoring the warning as I don't know enough about *bison* to try to actually fix it.

For my tests I used *-Wno-conflicts-sr* in **CMakeLists.txt**, but I haven't looked at all the supported build frameworks, and I'm not sure if *bison* is used in all cases.

bleader:
Fixes #753 
While looking for a way to pass *-Wno-conflicts-sr* from our build framework, I stumbled on something I missed last time I looked at this in the [bison documentation](https://www.gnu.org/software/bison/manual/html_node/Expect-Decl.html) which explains what they suggest is the proper way to fix these warnings.

So here is a patch using just that suggested *expect* in the case of *grammar.y*.

### Commitlog
Shift/reduce warnings are apparently mostly harmless:
https://www.gnu.org/software/bison/manual/html_node/Expect-Decl.html

The suggested way to handle them in bison documentation is to monitor
the amount isn't changing over time.

The usual output is:
grammar.y: warning: 38 shift/reduce conflicts [-Wconflicts-sr]

So I added the %expect 38 in grammar.y to avoid warnings.

### DIFF
--- a/grammar.y
+++ b/grammar.y
@@ -19,6 +19,14 @@
 %lex-param   {void *yyscanner}
 
 /*
+ * According to bison documentation, shift/reduce conflicts are not an issue
+ * in most parsers as long as the number does not evolve over time:
+ * https://www.gnu.org/software/bison/manual/html_node/Expect-Decl.html
+ * So, following the advice use %expect to check the amount of shift/reduce
+ * warnings.
+ */
+%expect 38
+/*
  * And we need to pass the compiler state to the scanner.
  */
 %parse-param { compiler_state_t *cstate }


guyharris:
Hopefully the claim in [the Berkeley YACC documentation](https://invisible-island.net/byacc/manpage/yacc.html) that "Bison's support for "%expect" is broken in more than  one  release." either 1) doesn't apply to releases that are still likely to be used or 2) merely means that the warning won't be suppressed.

bleader:
Sorry I missed that, I hope for 1)

PR #755, opened 2018-09-04
add pcap stats reset

wy7980:
pcap_stats function get a stastic of pcap, but ps_recv, ps_drop, ps_ifdrop can be overflowed, so I add this api for pcap stats reset.


guyharris:
Implementations of the API are needed for *all* platforms, not just Linux.

infrastation:
What is the use case for the new function from a library user point of view?

infrastation:
Closing due to lack of feedback.

PR #756, opened 2018-09-06
replace CMAKE_SOURCE_DIR by CMAKE_CURRENT_SOURCE_DIR

stubbfel:
This is an approach to use CMAKE_CURRENT_SOURCE_DIR intead of CMAKE_SOURCE_DIR.

https://cmake.org/cmake/help/latest/variable/CMAKE_CURRENT_SOURCE_DIR.html
http://johnnado.com/cmake-directory-variables/
http://fujii.github.io/2015/10/10/cmake-best-practice

So its possible add libpcap as subdirectory ( https://cmake.org/cmake/help/v3.0/command/add_subdirectory.html ) in other "cmake-projects".

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,7 @@ if(POLICY CMP0042)
     cmake_policy(SET CMP0042 OLD)
 endif()
 
-set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/Modules)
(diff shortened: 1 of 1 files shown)


guyharris:
> replace CMAKE_CURRENT_SOURCE_DIR by CMAKE_SOURCE_DIR

You mean "replace CMAKE_SOURCE_DIR by CMAKE_CURRENT_SOURCE_DIR"

PR #761, opened 2018-09-18
add ifindex field in pcap_pkthdr struct otherwise can not get it when…

zhangkaiheb:
… capture packets using 'any' interface


guyharris:
That change breaks the libpcap ABI; the `pcap_next()` routine takes a pointer to a `struct pcap_pkthdr` as an argument, and thus, if a program compiled with the old definition of the structure is run with a shared library using the new definition of the structure, libpcap will write past the end of the structure.

Instead, you can request the `DLT_LINUX_SLL2` link-layer header type when capturing on the "any" device; it includes the interface index in the pseudo-header.  That avoids making a possibly ABI-breaking

PR #771, opened 2018-10-19
Allow rpcapd to rebind more rapidly

### LINKED ISSUE #765: rpcapd: add reuseaddr so daemon can be restarted quicky
In some cases (lost link then killing daemon for example), the listen port may still be in use when restarting the daemon. Using *SO_REUSEADDR* flag before *bind()* could help with this issue.

Given this is handled by the *sock_open()* function, I made my patch adding a call to *setsockopt()* before the *bind()* in case of server socket. This will also be applied in case of active connection, which may be unecessary but shouldn't create any issue, as far as I can tell.

The way I did it is on [my fork](https://github.com/bleader/libpcap/commit/f3cc1dc38dbacc2067a197ca66f96cc71a1a3129), I wasn't able to test it on windows, but that should work, so I'll do a pull-request to check if CI builds fine.

ether42:
Sometimes rpcapd can't rebind due to lingering sockets, here is a basic patch to avoid some of the corner cases.

Closes #765.

### DIFF
--- a/sockutils.c
+++ b/sockutils.c
@@ -346,6 +346,18 @@ SOCKET sock_open(struct addrinfo *addrinfo, int server, int nconn, char *errbuf,
 	/* This is a server socket */
 	if (server)
 	{
+		/*
+		 * Allow a new server to bind the socket after the old one exited,
+		 * even if lingering sockets are still present.
+		 */
+		int optval = 1;
+		if (setsockopt(sock, SOL_SOCKET, SO_REUSEADDR, &optval, sizeof(optval)) < 0)
+		{
+			sock_geterror("socket(): ", errbuf, errbuflen);
+			/* dont treat an error as a failure */
+			SOCK_DEBUG_MESSAGE(errbuf);
+		}
+
 #if defined(IPV6_V6ONLY) || defined(IPV6_BINDV6ONLY)
 		/*
 		 * Force the use of IPv6-only addresses.


fxlb:
New warning:
```
./sockutils.c:352:4: warning: implicit declaration of function 'SOCK_DEBUG_MESSAGE' is invalid in C99 [-Wimplicit-function-declaration]
                        SOCK_DEBUG_MESSAGE(errbuf);
                        ^
./sockutils.c:352:4: warning: this function declaration is not a prototype [-Wstrict-prototypes]

```

fxlb:
Build is broken:
https://travis-ci.org/the-tcpdump-group/libpcap/jobs/524462162

fxlb:
See: 245ead8bc052d2dcf88c536511af895696946afd (Get rid of the last SOCK_DEBUG_MESSAGE() calls).

fxlb:
(A pull request should always be asked for a rebase before merging to avoid such problems and to have a better git history)

mcr:
Wow. I guess the "safe to merge" check lied to me.
I will fix that today.

mcr:
fxlb <notifications@github.com> wrote:
    > (A pull request should always be asked for a rebase before merging to
    > avoid such problems and to have a better git history)

I agree in general, but the patch seemed so small :-)

fxlb:
Small, yes, but 352 commits between merge commit and pr commit, make a not very readable history:
```
fddbd52f Merge pull request #771 from ether42/reuseaddr
...
[352 commits]
...
c51bcb50 Allow rpcapd to rebind more rapidly
```

Issue #766, opened 2018-10-03
rpcap & rpcapd: lost link isn't detected and may get stuck.


bleader:
When using rpcapd + a client, if the link goes down long enough (about 60s seems to be enough), the connection won't drop, but nothing will go through between the two anymore.

In case we want to have a fairly reliable way to get traffic it can be a pretty important issues.

I tested this with a master rpcapd + tcpdump built with the matching libpcap, as well as another client using the libpcap, put the link down on the machine hosting the rpcapd, then putting it back up. Both the control and data sockets stay in **ESTABLISHED** mode. As far as I could tell, the rpcapd will timeout in about 15 minutes (which match linux's *tcp_replies2* setting as seen [here](https://pracucci.com/linux-tcp-rto-min-max-and-tcp-retries2.html)), and I never waited long enough for it to drop on the client side.

One idea is to use *TCP*'s *keepalive* and *user_timeout*, *keepalive* so that an inactive socket will send keepalive probes to check if link is still active, this will make the socket end up in **FIN_WAIT1**, and *user_timeout* will allow this to actually timeout faster. This is meant to detect quickly the disconnections and allow reconnection to happen in an acceptable delay.

I added a **PCAP_OPENFLAG_KEEPALIVE** that can be passed to *pcap_open()* and trigger the setting of these flags.

The main issue here is that these flags don't exist on windows beside **SO_KEEPALIVE**, which alone won't have much use I think. I also don't have a mean to test this on windows.

I have [this patch](https://github.com/bleader/libpcap/commit/0d132ca8ab1544cb9b86036264761c43a6a3ebb8) which is currently used in the project where it is required, but I don't think doing a pull-request with something that will fail CI is wise, it is more an input to see what can be done here, open a discussion if need be.

PR #772, opened 2018-10-19
Verify rpcap control socket status during reception

kevinboulain:
This is probably the first of a few patches so that a client can more reliably handle different connections failure.

In this patch, I chose to extend the `select` on the control socket during packet retrieval.
When using TCP, it's fine to totally avoid checking the state of the control socket (as if the data socket dies, the control socket would probably too), but when using UDP, the client may be stuck indefinitely even though the control socket surely died.

I'm not sure there is a better way than using `MSG_PEEK` (which should be safe in this case as the control socket seems to be discarded until the end of the packet capture). `sock_recv` didn't seem suitable (can't pass the flag, message oriented, etc).

The advantage of this solution being that TCP keep-alives  (probably via `SO_KEEPALIVE` and `TCP_USER_TIMEOUT`) could be set on the control socket to be more reactive when the connection drops (#766 is related but it isn't suitable for UDP and probably too hard-coded).

Do you have any idea on what would be the best way to provide keep-alives from the API standpoint?
It can think of two ways:
 - introduce a new flag, for example `PCAP_OPENFLAG_KEEPALIVE` but the downside is that we shouldn't probably hard-code timeouts 
 - introduce helpers functions, that could be used after the call to `pcap_open`

What do you think?

Regards.

**EDIT**: would have liked to use the `MAX` macro in the `select` but it's defined in `optimize.c` and there doesn't seem to have some internal/miscellaneous header.

### DIFF
--- a/pcap-rpcap.c
+++ b/pcap-rpcap.c
@@ -396,10 +396,16 @@ static int pcap_read_nocb_remote(pcap_t *p, struct pcap_pkthdr *pkt_header, u_ch
 	/*
 	 * 'fp->rmt_sockdata' has always to be set before calling the select(),
 	 * since it is cleared by the select()
+	 *
+	 * While not strictly necessary, it's best to include the control socket.
+	 * It allows us to check for a connection drop as the data socket may use UDP
+	 * and as such, is without any mean to report back any error to the client.
 	 */
+	FD_SET(pr->rmt_sockctrl, &rfds);
 	FD_SET(pr->rmt_sockdata, &rfds);
 
-	retval = select((int) pr->rmt_sockdata + 1, &rfds, NULL, NULL, &tv);
+	const int max_fd = pr->rmt_sockctrl > pr->rmt_sockdata ? pr->rmt_sockctrl : pr->rmt_sockdata;
+	retval = select((int) max_fd + 1, &rfds, NULL, NULL, &tv);
 	if (retval == -1)
 	{
 #ifndef _WIN32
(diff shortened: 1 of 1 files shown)


pcap-rpcap.c
```diff
@@ -413,8 +419,31 @@ static int pcap_read_nocb_remote(pcap_t *p, struct pcap_pkthdr *pkt_header, u_ch
...
 
+	/*
+	 * In the rpcap protocol, once the capture starts, the control socket isn't
+	 * used anymore until the capture ends.
+	 * However, it's the only way to check for connection errors
+	 * as the data socket may uses UDP.
+	 */
+	if (FD_ISSET(pr->rmt_sockctrl, &rfds)) {
+		uint8 byte;
+		const ssize_t nread = recv(pr->rmt_sockctrl, &byte, sizeof(byte), MSG_PEEK);
```
guyharris: This should call `sock_recv()`, to filter out UN\*X vs. Windows differences.  Note that `sock_recv()` returns an `int`; `ssize_t` is a UN\*X type but not a Windows type.

ether42: Yep, I just saw the builder failure and exported it in the header before seeing your comment.

The problem with `sock_recv` (as quickly explained above) is that I should add another parameter just to handle `MSG_PEEK` or a special flag and that there are already special flags conflicting with this use case (like `SOCK_EOF*`). That seemed less clean to me at first :)

guyharris: >  I should add another parameter just to handle `MSG_PEEK` or a special flag

No, `SOCK_MSG_PEEK` already been added in d491c4d68d38902ce71fcceed49aeac1f574f533, so you don't have to add it, just use it.

pcap-rpcap.c
```diff
@@ -413,8 +419,31 @@ static int pcap_read_nocb_remote(pcap_t *p, struct pcap_pkthdr *pkt_header, u_ch
...
+		if (nread <= 0)
+		{
+			if (nread == -1)
+				sock_geterror("recv(): ", p->errbuf, PCAP_ERRBUF_SIZE);
+			/* control socket has been properly closed or got an error, stop capture */
+			SOCK_DEBUG_MESSAGE("Control socket closed, stopping capture");
+			return -1;
```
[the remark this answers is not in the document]
guyharris: `sock_recv()` already fills in the error message.

pcap-rpcap.c
```diff
@@ -396,10 +396,16 @@ static int pcap_read_nocb_remote(pcap_t *p, struct pcap_pkthdr *pkt_header, u_ch
...
+	 *
+	 * While not strictly necessary, it's best to include the control socket.
+	 * It allows us to check for a connection drop as the data socket may use UDP
+	 * and as such, is without any mean to report back any error to the client.
 	 */
+	FD_SET(pr->rmt_sockctrl, &rfds);
 	FD_SET(pr->rmt_sockdata, &rfds);
 
-	retval = select((int) pr->rmt_sockdata + 1, &rfds, NULL, NULL, &tv);
+    const int max_fd = pr->rmt_sockctrl > pr->rmt_sockdata ? pr->rmt_sockctrl : pr->rmt_sockdata;
```
guyharris: Indent this line the same way the adjacent line is indented.

pcap-rpcap.c
ether42: Thank you!

ether42:
AppVeyor crashed two times on a network fetch, looks like it's building the code otherwise.

infrastation:
Is there a good reason against rebasing and merging these changes?

PR #774, opened 2018-10-23
Initializing np value to NULL

catenacyber:
Found by oss-fuzz :
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=10919

Apparently, we cannot trust libc to assign value in `getnetbyname_r` even when it returns no error

So, the fix is simple

PS : @guyharris 👍 for all the fixes so far

### DIFF
--- a/nametoaddr.c
+++ b/nametoaddr.c
@@ -239,7 +239,7 @@ pcap_nametonetaddr(const char *name)
 	/*
 	 * UN*X.
 	 */
-	struct netent *np;
+	struct netent *np = NULL;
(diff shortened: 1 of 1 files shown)


guyharris:
> Apparently, we cannot trust libc to assign value in `getnetbyname_r` even when it returns no error

I.e., [this](http://man7.org/linux/man-pages/man3/getnetbyname_r.3.html) lies when it says

> If the function call successfully obtains a network record, then \**result* is set pointing to *result_buf*; otherwise, \**result* is set to NULL.

PR #776, opened 2018-10-23
Avoids undefined shift

catenacyber:
Found by oss-fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=10931

Reproduce with filter "mpls 637625"


guyharris:
Fixed in 71abe040fa49dfcb8ba3e08e5f21896b0c7cc6e8, which also 1) reports an error for MPLS labels that don't fit in 20 bits and 2) does an unsigned shift so there's no sign bit into which to shift.

(And isn't 637625 & 0xfffff equal to 637625?)

PR #777, opened 2018-10-23
Avoids infinite loop in filter optimzation

catenacyber:
Found by oss-fuzz.
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=10930

This is obviously not the best patch, as there must be some graph logic flaw to be fixed, but at least we do not loop forever

### DIFF
--- a/optimize.c
+++ b/optimize.c
@@ -1834,7 +1834,7 @@ static void
 opt_loop(compiler_state_t *cstate, opt_state_t *opt_state, struct icode *ic,
     int do_stmts)
 {
-
+	int loopCounter = 0;
 #ifdef BDEBUG
 	if (pcap_optimizer_debug > 1 || pcap_print_dot_graph) {
 		printf("opt_loop(root, %d) begin\n", do_stmts);
@@ -1855,6 +1855,10 @@ opt_loop(compiler_state_t *cstate, opt_state_t *opt_state, struct icode *ic,
 			opt_dump(cstate, ic);
 		}
 #endif
+		loopCounter++;
+		if (loopCounter > 1000) {
+			break;
+		}
 	} while (!opt_state->done);
 }
 


guyharris:
> as there must be some graph logic flaw to be fixed

If "assuming that there's a fixed point in the iteration" counts as a graph logic flaw, yes, but, the last time I saw an issue of this sort, that was the problem - the logic was fine, it just happened to get to a stage where the optimizing path shuffled some instructions but didn't reduce the number of instructions, and the next step shuffled them back, so it was stuck in a loop where the code kept changing back and forth between two equivalent programs.

catenacyber:
> it just happened to get to a stage where the optimizing path shuffled some instructions but didn't reduce the number of instructions, and the next step shuffled them back, so it was stuck in a loop where the code kept changing back and forth between two equivalent program

Indeed, it gets to the same behavior with `opt_blks`.
So, should we add a watchdog to avoid infinite loop or change the iteration so that it reaches a fixed point ?
(By the way, thanks @infrastation for the pointer to the slides)

And I found another strange behavior : I was looking to understand the priority order of the `and` `or` operations but I did not manage to get it.
`tcp port (11 or !12 and 10 or (!11 and 10) or 13)` gets into an infinite loop
So I was assuming that I could add parenthesis on the highest priority operation and keep the same behavior.

But :
- `tcp port (11 or (!12 and 10) or (!11 and 10) or 13)` ends ok
- `tcp port ((11 or !12) and 10 or (!11 and 10) or 13)` errors with `can't parse filter expression: syntax error`
- `tcp port (11 or !12 and (10 or (!11 and 10)) or 13)` ends ok
- `tcp port (11 or !12 and 10 or ((!11 and 10) or 13))`  errors
- `tcp port ((11 or !12 and 10) or (!11 and 10) or 13)` errors 
- `tcp port ((11 or !12 and 10 or (!11 and 10)) or 13)` errors
- `tcp port (11 or (!12 and 10 or (!11 and 10)) or 13)` ends ok
- `tcp port (11 or (!12 and 10 or (!11 and 10) or 13))` ends ok
- `tcp port (11 or !12 and (10 or (!11 and 10) or 13))` ends ok

Am I missing something ?

catenacyber:
ping @guyharris @infrastation ?

inferno-chromium:
ping @guyharris @infrastation

Please consider uptaking this PR. This will make fuzzing in OSS-Fuzz project efficient. Currently fuzz targets are easily running into timeout making them less effective.

catenacyber:
We can try a different patch : we can check if the number of nodes decreases. If it does not decrease in 100 iterations, we can break out of the loop.

mcr:
Before I merge this, I wanted to know if it needed to be as high as 1000?

catenacyber:
My experience was that optimization can take 20 iterations
So I wrote 1000 to be safe.

Oppen:
Why not measuring the clock-time instead? I think that's what the user is going to perceive. Also, maybe add a way to configure this limit?

guyharris:
> Why not measuring the clock-time instead? I think that's what the user is going to perceive. Also, maybe add a way to configure this limit?

*Ideally*, we'd like to figure out whether the optimizer is really just shuffling a couple of chunks of code back and forth and back and forth without making anything resembling forward progress (or, if possible, somehow prevent it from shuffling blocks of code at all if it won't improve the code in the general case, rather than just interchanging two tests in a way that makes some cases faster and others slower), but that would require a deeper dive into 1) compiler optimization in general and 2) the BPF compiler's optimizer than I've had time to do.

(It'd be wonderful if we could just count on LLVM being available and use that.)

Oppen:
> (It'd be wonderful if we could just count on LLVM being available and use that.)

I actually thought of that option, but:
A) I don't have the skillset to do it, and while I'll learn a bit of the programming side (I enrolled to an assignature at my College for November), I lack a lot on the theoretical background, so I didn't want to ask someone else to do the job.
B) Assumed the dependency would be unwanted.

Oppen:
> _Ideally_, we'd like to figure out whether the optimizer is really just shuffling a couple of chunks of code back and forth and back and forth without making anything resembling forward progress

I'm probably wrong, but isn't that undecidable or something?

Oppen:
Now that I think about it, can't we *optionally* depend on LLVM, and fall back to libpcap's own compiler if absent?

guyharris:
I've checked change f2d84366a864f7b41f59ef47334f6a53aa914b32, which also uses a loop counter, but only counts consecutive optimizer passes that do nothing other than move code.

catenacyber:
Thanks Guy for your better patch

PR #778, opened 2018-10-24
Initializes value before use

catenacyber:
Found by oss-fuzz:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=10991

Another solution would be to use another array of Booleans wether the memory is initialized or not


catenacyber:
@guyharris are you working on this ? How can I help ?

PR #779, opened 2018-10-25
Fix build on Mojave

trufae:
Last macOS (darwin18) is 64bit only, so enforcing those -arch flags is making the build fail.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1703,7 +1703,21 @@ darwin*)
 			V_PROG_CCOPT_FAT="-arch x86_64 -arch i386"
 			V_PROG_LDFLAGS_FAT="-arch x86_64 -arch i386"
 			;;
-
+		darwin18*)
+			#
+			# Post-Snow Leopard.  Build libraries for x86-64
+			# and 32-bit x86, with x86-64 first, and build
+			# executables only for x86-64.  (That's what
+			# Apple does.)  This requires no special flags
+			# for programs.
+			# XXX - update if and when Apple drops support
+			# for 32-bit x86 code and if and when Apple adds
+			# ARM-based Macs.  (You're on your own for iOS
+			# etc.)
+			#
+			V_LIB_CCOPT_FAT="-arch x86_64"
+			V_LIB_LDFLAGS_FAT="-arch x86_64"
+			;;
 		darwin*)
 			#
 			# Post-Snow Leopard.  Build libraries for x86-64


guyharris:
> Last macOS (darwin18) is 64bit only

macOS has, for *several* releases, only run on 64-bit processors.

However, Xcode has continued to support building 32-bit code.

Xcode 10 on Mojave still does, if 1) you install the command-line tools package and 2) then install the package at `/Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg`, which puts the include files where they *belong* on a UN\*X system - under `/usr/include`.

trufae:
The problem i see is that i think those flags shouldnt be hardcoded by the configure script. The packager must specify them in the commandline. Also, i dont see a way to get rid of them without patching the configure script

trufae:
The problem is that those flags are unnecessary. If you dont pass them the build goes as expected. They should be in a release.sh or build.sh to get same reproducible builds in your farms or so. But enforcing cflags at configure time is imho a bad approach.

I understand they are there to ensure all libpcap builds support 32 and 64 bit targets, but this is something arbitrary. And users should have the ability to make 32 or 64 bit only builds without having to patch a configure script

agnosticdev:
I will defer to @guyharris on this, but I would imagine that in macOS 10.15 that the 32 bit option will be completely removed by Apple.

This is topic also discussed on https://github.com/the-tcpdump-group/libpcap/issues/760

The point being, it seems logical to me to only provide the 64 bit option in the build script when Apple officially drops 32 bit support.

jmayer:
> > Last macOS (darwin18) is 64bit only
> 
> macOS has, for _several_ releases, only run on 64-bit processors.
> 
> However, Xcode has continued to support building 32-bit code.
> 
> Xcode 10 on Mojave still does, if 1) you install the command-line tools package and 2) then install the package at `/Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg`, which puts the include files where they _belong_ on a UN*X system - under `/usr/include`.

On my HighSierra, it was enough to install the XCode command line tools package without any further steps to make it compile and link. How about doing something like 

elseif(EXISTS /Library/Developer/CommandLineTools/usr/pick/sensible/file/here)
  set(OSX_LIBRARY_ARCHITECTURES "x86_64;i386")
else()
  set(OSX_LIBRARY_ARCHITECTURES "x86_64")
endif()

mcr:
jmayer <notifications@github.com> wrote:
    > On my HighSierra, it was enough to install the XCode command line tools package
    > without any further steps to make it compile and link. How about doing
    > something like

Can we do this for Travis-CI builds, I wonder?

    > elseif(EXISTS /Library/Developer/CommandLineTools/usr/pick/sensible/file/here)
    > set(OSX_LIBRARY_ARCHITECTURES "x86_64;i386")
    > else()
    > set(OSX_LIBRARY_ARCHITECTURES "x86_64")
    > endif()

send patch/pull-request?

guyharris:
This should now be fixed, for both High Sierra and Mojave, by 0466b96b44e764ca00ba2a7eddce971f594be255, e8f24aa79fc6530573ffbe89f0be044cbbf70f02, and 4dcea058851263dbe5aa624d9a77ff874b75e5b5.  They will check - both with autotools *and* CMake, as *both* have to be changed - whether building for 32-bit x86 is supported and, if it's not, build only for x86-64 and issue a warning suggesting how to get 32-bit x86 builds to work.

trufae:
Thanks!

PR #780, opened 2018-10-26
Avoids integer overflow in bpf optimization

catenacyber:
Found by oss-fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=11106

We cannot minus an unsigned integer greater than 0x80000000 so we report an error

### DIFF
--- a/bpf_filter.c
+++ b/bpf_filter.c
@@ -58,7 +58,6 @@
 #include <pcap-int.h>
 
 #include <stdlib.h>
-#include <limits.h>
 
 #ifdef __linux__
 #include <linux/types.h>
@@ -362,7 +361,7 @@ pcap_filter_with_aux_data(const struct bpf_insn *pc, const u_char *p,
 			 * can't be unsigned; throw some casts to
 			 * specify what we're trying to do.
 			 */
-			if (A > INT_MAX) {
+			if (A == 0x80000000) {
 				return 0;
 			}
 			A = (uint32_t)(-(int32_t)A);
--- a/bpf_filter.c
+++ b/bpf_filter.c
@@ -361,10 +361,7 @@ pcap_filter_with_aux_data(const struct bpf_insn *pc, const u_char *p,
 			 * can't be unsigned; throw some casts to
 			 * specify what we're trying to do.
 			 */
-			if (A == 0x80000000) {
-				return 0;
-			}
-			A = (uint32_t)(-(int32_t)A);
+			A = 0U - A;
 			continue;
 
 		case BPF_MISC|BPF_TAX:
--- a/bpf_filter.c
+++ b/bpf_filter.c
@@ -58,6 +58,7 @@
 #include <pcap-int.h>
 
 #include <stdlib.h>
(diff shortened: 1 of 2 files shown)


catenacyber:
And another similar commit for https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=11062

guyharris:
> We cannot minus an unsigned integer greater than 0x80000000

Most CPUs can negate any arbitrary collection of 32 bits, and BPF includes a (simulated) CPU, so....

I'd have to do a lot more language-lawyering on the C99 spec to be sure, but, as far as I know, the expression (0U - *unsigned_variable*) is subtracting one unsigned value from another, and, if you view unsigned *n*-bit twos-complement arithmetic as arithmetic modulo 2^*n*, it's well-defined for all values of *unsigned_variable*, so evaluating (0U - *unsigned_variable*) should never involve undefined behavior.

catenacyber:
Indeed, all is good as long as we keep it unsigned.
But the negate of 0x80000000, ie -2147483648, is +2147483648 and it cannot be represented in an `int32_t` type and that is the bug.

guyharris:
> But the negate of 0x80000000, ie -2147483648, is +2147483648 and it cannot be represented in an int32_t type and that is the bug.

If we don't use any signed types, there's no bug, if (0U - 0x80000000U) is defined in C, which I think it is.

catenacyber:
Maybe, I am not sure, but the comment says  
```Most BPF arithmetic is unsigned, but negation can't be unsigned; throw some casts to specify what we're trying to do.
A = (uint32_t)(-(int32_t)A);```

guyharris:
The casts were thrown in by commit 2fc98f9a355a97a69f8701cd7822ed5d6537c57, the commit message for which was "Squelch warnings for negating an unsigned value."  I guess some C compiler whined about applying the `-` operator to an unsigned value; that was one way of squelching the compiler warning, but just doing `A = (0U - A);` should squelch them as well *and* prevent UBSan from putting in checks to make sure we're not negating an un-negatable *signed* value (-2147483648).

catenacyber:
Yes that would be better for the case in optimize.c
For the one in bpf_filter.c, I am not sure what is expected as it looks to be signed integers from the beginning... You know best I guess.

cheers for this explanation 👍

guyharris:
And now it's pretty much the same as what's currently checked in after 52a246fb5291e97ba97bc07fc66e0d598484e471.

guyharris:
> For the one in bpf_filter.c, I am not sure what is expected as it looks to be signed integers from the beginning... 

Nope:

        register uint32_t A, X;
        register bpf_u_int32 k;
        uint32_t mem[BPF_MEMWORDS];

catenacyber:
Great for the commit. For bpf_filter.c i menant the const_val field as you casted or.

guyharris:
The changes in 52a246fb5291e97ba97bc07fc66e0d598484e471 fix the problems with the two overflow test cases, using UBSan on my Mac.

PR #785, opened 2018-11-29
Check whether i386 builds on current macOS

jmayer:
Properly check for i386 build support


guyharris:
This should now be fixed, for both High Sierra and Mojave, by 0466b96b44e764ca00ba2a7eddce971f594be255, e8f24aa79fc6530573ffbe89f0be044cbbf70f02, and 4dcea058851263dbe5aa624d9a77ff874b75e5b5.  They will check - both with autotools *and* CMake, as *both* have to be changed - whether building for 32-bit x86 is supported and, if it's not, build only for x86-64 and issue a warning suggesting how to get 32-bit x86 builds to work.

PR #790, opened 2018-12-16
Enable DPDK for libpcap, tested with DPDK v18.11

jingleyang:
# Description

1. Pcap-dpdk provides libpcap the ability to use DPDK with the device name as dpdk:[portid], such as dpdk:0.
2. DPDK is a set of libraries and drivers for fast packet processing. (https://www.dpdk.org/)

# Limitations

1. By default enable_dpdk is no, unless you set inlcudes and lib dir
by --with-dpdk-includes= --with-dpdk-libraries=
2. Only support link libdpdk.so dynamicly, because the libdpdk.a will not work correctly.
3. Only support read operation, and packet injection has not been supported yet.
4. I have tested on DPDK v18.11.

# Usage

1. compile DPDK as shared library and install.(https://github.com/DPDK/dpdk.git)

You shall modify the file $RTE_SDK/$RTE_TARGET/.config and set:
CONFIG_RTE_BUILD_SHARED_LIB=y

2. launch l2fwd that is one of DPDK examples correctly, and get device information.

You shall learn how to bind nic with DPDK-compatible driver by $RTE_SDK/usertools/dpdk-devbind.py, such as igb_uio.
And enable hugepages by dpdk-setup.sh

Then launch the l2fwd with dynamic dirver support. For example:
$RTE_SDK/examples/l2fwd/$RTE_TARGET/l2fwd -dlibrte_pmd_e1000.so -dlibrte_pmd_ixgbe.so -dlibrte_mempool_ring.so -- -p 0x1

3. compile libpcap with dpdk options.

you shall run the following command to generate a new configure

make clean
autoreconf

Then, run configure with dpdk options.
For Ubuntu, they are --with-dpdk-includes=/usr/local/include/dpdk/ --with-dpdk-libraries=/usr/local/lib

4. link your own program with libpcap, and use DPDK with the device name as dpdk[portid], such as dpdk:0.
And you shall set DPDK configure options by environment variable DPDK_CFG
For example, the testprogs/capturetest could be lanched by:

env DPDK_CFG="--log-level=debug -l0 -dlibrte_pmd_e1000.so -dlibrte_pmd_ixgbe.so -dlibrte_mempool_ring.so" ./capturetest -i dpdk:0

The program will print the following message on my computer:

	USER1: dpdk cfg: libpcap_dpdk --log-level=debug -l0 -dlibrte_pmd_e1000.so -dlibrte_pmd_ixgbe.so -dlibrte_mempool_ring.so
	EAL: Detected 4 lcore(s)
	EAL: Detected 1 NUMA nodes
	EAL: Multi-process socket /var/run/dpdk/rte/mp_socket
	EAL: No free hugepages reported in hugepages-1048576kB
	EAL: Probing VFIO support...
	EAL: PCI device 0000:00:19.0 on NUMA socket -1
	EAL:   Invalid NUMA socket, default to 0
	EAL:   probe driver: 8086:1559 net_e1000_em
	USER1: pcap_dpdk_activate device dpdk:0 portid 0, pci_addr: 0000:00:19.0
	USER1: Port 0 Link Up. Speed 1000 Mbps - full-duplex
	USER1: Port 0, MAC address: [MAC ADDR]

	Listening on dpdk:0
	USER1: dpdk: lcoreid=0 runs for portid=0


jingleyang:
I found this pull request fail to pass the CI with the error msg: 

    make: *** No rule to make target `@DPDK_SRC@'

`autoreconf` should be add for travis-ci, because this pull request adds pcap-dpdk.c|h and modifies the configure.ac and Makefile.in

How to solve this problem?
1. Could I commit a new configure file ? or 
2. Could admin modify the travis-ci configure file ?

Thanks.

mcr:
jingyu yang <notifications@github.com> wrote:
> 1 Could I commit a new configure file ? or
> 2 Could admin modify the travis-ci configure file ?

You can edit .travis.yml in your fork to add the right things you might need
and cycle on it until you get it right.

It is likely that your DPDK extensions will fail if they are not tested
regularly, so it would be great if you added a new Travis matrix option that
installed all the right DPDK libraries and built libpcap with dpdk support
(as well as without).

--
]   Michael Richardson, Sandelman Software Works        |    IoT architect   [

guyharris:
> `autoreconf` should be add for travis-ci

No, `autoreconf` should be run by people who change `configure.ac` or `aclocal.m4`, before they commit a change, because, currently, the `configure` file is in the repository.

Perhaps it shouldn't be, and the process of building a release tarball should run `autoconf`.

> How to solve this problem?
> 
> 1. Could I commit a new configure file ?

Yes.

guyharris:
You need to update `CMakeLists.txt` too.

jingleyang:
Thank you and I will try to fix it.

jingleyang:
I feel so excited that all checks have passed.
1. I have tested with my own .travis.yml to enable dpdk. but this PR remains the original .travis.yml, because It will take a long time to compile DPDK.
The following screenshot show pcap-dpdk.c is compiled correctly.
dpdk_ci

dpdk_compile

2. Cmake is supported.

guyharris:
Is it possible to enumerate all the available DPDK devices, giving their names?  If the port IDs are numeric, this presumably means knowing what port IDs exist.

If so, `pcap_dpdk_findalldevs()` should do so, as it'd make it easier to capture on those devices from Wireshark, where "easier" means "the Wireshark developers wouldn't have to provide a UI to allow arbitrary device names to be specified"; [somebody's already asked about capturing on DPDK devices in Wireshark](https://ask.wireshark.org/question/6559/capturing-on-dpdk-interface).  (Not that Wireshark *shouldn't* have such a UI....)

configure.ac
```diff
@@ -2128,6 +2128,57 @@ AC_INCLUDES_DEFAULT
...
+],[])
+
+if test "x$enable_dpdk" != "xno" ; then
+	# If necessary, set default paths for DPDK API headers and libraries.
+	if test -z "$dpdk_root"; then
+		dpdk_root=/usr/local
+	fi
+
+	if test -z "$dpdk_include_dir"; then
+		dpdk_include_dir="$dpdk_root/include/dpdk"
```
guyharris: Why isn't this just `$dpdk_root/include`, with the `#include` statements including `dpdk/{includefile}`?

If it's always the case that the DPDK include directory is `{rootdirectory}/include` and the the DPDK library directory is `{rootdirectory}/lib`, for some value of `{rootdirectory}`, then you can get rid of the `--with-dpdk-includes` and `--with-dpdk-libraries` options, and simply have a `--with-dpdk` option that takes an optional argument that, if present, is `{rootdirectory}`.  That's how the other optional capture device support options, such as `--with-dag`, work.

pcap-dpdk.c
```diff
@@ -0,0 +1,742 @@
...
+	unsigned nb_mbufs = DPDK_NB_MBUFS;
+	struct rte_eth_rxconf rxq_conf;
+	struct rte_eth_txconf txq_conf;
+	struct rte_eth_conf local_port_conf = port_conf;
+	struct rte_eth_dev_info dev_info;
+	int is_port_up = 0;
+	struct rte_eth_link link;
+	if (p == NULL)
+	{
+		return PCAP_ERROR;
```
guyharris: 1. This should never happen - the activate routine is only called by `pcap_activate()`, and, before it calls it, it'll call `pcap_check_activated()`, which will check `p->activated`, and that will crash, on most if not all platforms, if `p` is null.

1. If you return `PCAP_ERROR`, you *have* to set `p->errbuf` with an error message.

pcap-dpdk.c
```diff
@@ -0,0 +1,742 @@
...
+	int prefix_len = strlen(DPDK_PREFIX);
+	unsigned long ret_ul = 0L;
+
+	if (len<=prefix_len || strncmp(device, DPDK_PREFIX, prefix_len)) // check prefix dpdk:
+	{
+		return ret;
+	}
+	if (device[prefix_len]>='0' && device[prefix_len]<='9')
+	{ // is digital
+		ret_ul = strtoul(&(device[prefix_len]), NULL, 10);
```
guyharris: This should check whether `strtoul()` found a number.

That means that the second argument should point to a `char *` pointer and, when it returns, you should report an error if either 1) the pointer is equal to `&device[prefix_len])` (meaning the first character wasn't a digit) or 2) the pointer doesn't point to a `\0' (meaning that there's a non-digit in the string).

However, you should still do the test for the first character being a digit, because, for some reason, the argument to `strtoul()` can begin with a `-`, the "u" in "strtoul" standing for "unsigned" nonwithstanding.

guyharris:
> Thanks. DPDK provides `./usertools/dpdk-devbind.py -status` to provide some information about devices. I will figure it out on the weekend.

If it is the case that all ports from 0 to `rte_eth_dev_count_avail() - 1` exist, then all you need to do is to return ports from `dpdk:0` through `dpdk:{rte_eth_dev_count_avail() - 1}`.

pcap-dpdk.c
```diff
@@ -0,0 +1,742 @@
...
+			pcap_fmt_errmsg_for_errno(p->errbuf, PCAP_ERRBUF_SIZE,
+			    errno, "dpdk error: portid is invalid. device %s",
+			    p->opt.device);
+			ret = PCAP_ERROR;
+			break;
+		}
+
+		if (portid >= nb_ports)
+		{
+			pcap_fmt_errmsg_for_errno(p->errbuf, PCAP_ERRBUF_SIZE,
```
guyharris: This should perhaps just return `PCAP_ERROR_NO_SUCH_DEVICE`.

pcap-dpdk.c
```diff
@@ -0,0 +1,742 @@
...
+	struct rte_eth_conf local_port_conf = port_conf;
+	struct rte_eth_dev_info dev_info;
+	int is_port_up = 0;
+	struct rte_eth_link link;
+	if (p == NULL)
+	{
+		return PCAP_ERROR;
+	}
+
+	do{
```
guyharris: Does a process need any special permissions to open DPDK devices, e.g. must run as root, or as a member of a special group?

If so, do any routines called in this block return a particular error code meaning "permission denied" if the process doesn't have the required permissions?

If so, then if any of those routines return such an error code, the activate routine should return `PCAP_ERROR_PERM_DENIED`.

If the device exists but isn't "up" in some sense, is it possible to open it?

If so, do any routines called in this block return a particular error code meaning "device isn't up" if the device isn't "up" in that sense?

If so, then if any of those routines return such an error code, the activate routine should return `PCAP_ERROR_IFACE_NOT_UP`.

jingleyang:
Thank you very much for your review.
I will try to modify my code, and provide more detailed error code, such as PCAP_ERROR_XXX.
Cheers.

pcap-dpdk.c
```diff
@@ -0,0 +1,742 @@
+/*
+ * Copyright (C) 2018 All rights reserved.
```
guyharris: Is 2018 the person who holds the copyright to this code or is it an organization?

Or has the person or organization holding the copyright been omitted from the comment?

pcap-dpdk.c
```diff
@@ -0,0 +1,742 @@
...
+
+You shall learn how to bind nic with DPDK-compatible driver by $RTE_SDK/usertools/dpdk-devbind.py, such as igb_uio.
+And enable hugepages by dpdk-setup.sh
+
+Then launch the l2fwd with dynamic dirver support. For example:
+$RTE_SDK/examples/l2fwd/$RTE_TARGET/l2fwd -dlibrte_pmd_e1000.so -dlibrte_pmd_ixgbe.so -dlibrte_mempool_ring.so -- -p 0x1
+
+3. compile libpcap with dpdk options.
+
+you shall run the following command to generate a new configure
```
guyharris: That shouldn't be necessary - the libpcap download should come with a generated configure script.

pcap-dpdk.c
```diff
@@ -0,0 +1,742 @@
...
+
+make clean
+autoreconf
+
+Then, run configure with dpdk options.
+For Ubuntu, they are --with-dpdk-includes=/usr/local/include/dpdk/ --with-dpdk-libraries=/usr/local/lib
```
guyharris: Or run CMake with the DPDK options.

pcap-dpdk.c
```diff
@@ -0,0 +1,742 @@
...
+
+4. link your own program with libpcap, and use DPDK with the device name as dpdk[portid], such as dpdk:0.
```
guyharris: Presumably you mean "as dpdk:{portid}" - the : is necessary, and the curly brackets may more clearly indicate that they're not a syntactic element of the name, but an indication that {portid} is a variable.

pcap-dpdk.c
```diff
@@ -0,0 +1,742 @@
...
+And you shall set DPDK configure options by environment variable DPDK_CFG
+For example, the testprogs/capturetest could be lanched by: 
+
+env DPDK_CFG="--log-level=debug -l0 -dlibrte_pmd_e1000.so -dlibrte_pmd_ixgbe.so -dlibrte_mempool_ring.so" ./capturetest -i dpdk:0
```
guyharris: So why do *the names of some shared objects* have to be part of the configure options?

pcap-dpdk.c
jingleyang: No. This pull request is written by me individually, and I am willing to contribute to libpcap. 
The original comment is copy from other files.
So how should I modify this comment?
Thanks

pcap-dpdk.c
```diff
@@ -0,0 +1,742 @@
...
+		if (pd->break_loop){
+			break;
+		}
+		nb_rx = (int)rte_eth_rx_burst(portid, 0, pkts_burst, MAX_PKT_BURST);
+		pkt_cnt += nb_rx;
+		for ( i = 0; i < nb_rx; i++) {
+			m = pkts_burst[i];
+			calculate_timestamp(&(pd->ts_helper),&(pcap_header.ts));
+			data_len = rte_pktmbuf_data_len(m);
+			pcap_header.caplen = data_len;
```
guyharris: You should probably cut the packet short by the snapshot length by setting `pcap_header.caplen` to the minimum of `data_len` and `p->snapshot`, so that if something's capturing packets with a snapshot length, and writing them to a file, they will write no more than the snapshot length to the file (that being one of the purposes of a snapshot length) without having to do anything other than writing out `pcap_header.caplen` bytes of packet data.

pcap-dpdk.c
```diff
@@ -0,0 +1,742 @@
...
+			calculate_timestamp(&(pd->ts_helper),&(pcap_header.ts));
+			data_len = rte_pktmbuf_data_len(m);
+			pcap_header.caplen = data_len; 
+			pcap_header.len = data_len; 
+			// volatile prefetch
+			rte_prefetch0(rte_pktmbuf_mtod(m, void *));
+			bp = NULL;
+			if (m->nb_segs == 1)
+			{
+				bp = rte_pktmbuf_mtod(m, u_char *);
```
guyharris: Perhaps the setting of `pcap_header.len` and `pcap_header.caplen` should be moved here.

pcap-dpdk.c
```diff
@@ -0,0 +1,742 @@
...
+			}else{
+				if (m->pkt_len <= ETHER_MAX_JUMBO_FRAME_LEN)
+				{
+					gather_len = dpdk_gather_data(pd->pcap_tmp_buf, m);
+					bp = pd->pcap_tmp_buf;
+					pcap_header.caplen = gather_len;
```
guyharris: `pcap_header.caplen` should be clamped at `p->snapshot` here too.

pcap-dpdk.c
```diff
@@ -0,0 +1,742 @@
...
+					pcap_header.len = gather_len;
+				}else{
+					// size too large
+					// why only free this pkt
+					rte_pktmbuf_free(m);
```
guyharris: If it's too large, are you discarding it?

pcap-dpdk.c
```diff
@@ -0,0 +1,742 @@
...
+	// free pcap_dpdk?
+	pcap_cleanup_live_common(p);
+} 
+
+static int pcap_dpdk_setfilter(pcap_t *p, struct bpf_program *fp)
+{
+	//init bpf for dpdk, only support userspace bfp 
+	struct pcap_dpdk * pd = p->priv;
+	int ret=0;
+	ret = install_bpf_program(p, fp);
```
guyharris: If you're calling `install_bpf_program()`, and not passing the filter to the DPDK code to do the filtering at a lower level, you're *always* doing userland filtering, so:

1. you should set `p->setfilter_op` to `install_bpf_program`, as other pcap-XXX.c files that do filtering in userland do (see, for example, `pcap-rdmasniff.c`
1. you should eliminate the `filter_in_userland` structure member, and call `pcap_filter()` if `p->fcode.bf_insns` is non-null.

pcap-dpdk.c
jingleyang: It's a long story.
1. Currently, we can not use static link with DPDK in libpcap. only *dynamic link* works correctly.

Becuase, DPDK uses global static variable to initialize NIC device. When DPDK is static linked by other program, this variable is NULL. As a result, the device can not be found when rte_eal_init() is called. More information is at: https://github.com/DPDK/dpdk/blob/master/drivers/bus/pci/pci_common.c lineno:548 . I tried `--whole-archive` flags, but DPDK provides many .a files not only one. And some .a files are mutually exclusive.

 2. So dynamic link is the only choice. But DPDK only provides runtime option `-d` to load device related plugins.

For example:
librte_pmd_e1000.so is for e1000 driver and is usually used for 1G ethernet NICs.
librte_pmd_ixgbe.so is for igxbe driver and is usually used for 10G ethernet NICs.
librte_mempool_ring.so is used for mempool.
More information is at: http://mails.dpdk.org/archives/dev/2015-May/017789.html

Finally, I designed this way to configure DPDK correctly.
I will provide two most common plugin default, but user can overwrite env variable DPDK_CFG, to launch their NIC related driver plugins.

pcap-dpdk.c
guyharris: If the code is copyright by you, put your name into the copyright notice, e.g. "Copyright 2018 Jingle Yang All rights reserved."

Do this in other files, such as `pcap-dpdk.h`, as well.

pcap-dpdk.c
```diff
@@ -0,0 +1,740 @@
...
+		mbuf=mbuf->next;
+	}
+	return total_len;
+}
+
+static void pcap_dpdk_breakloop(pcap_t *p)
+{
+	pcap_breakloop_common(p);
+	struct pcap_dpdk *pd = (struct pcap_dpdk*)(p->priv);
+	pd->break_loop = p->break_loop;
```
guyharris: So why does there need to be a private copy of the "break out of the loop" variable?

pcap-dpdk.c
```diff
@@ -0,0 +1,740 @@
...
+	return total_len;
+}
+
+static void pcap_dpdk_breakloop(pcap_t *p)
+{
+	pcap_breakloop_common(p);
+	struct pcap_dpdk *pd = (struct pcap_dpdk*)(p->priv);
+	pd->break_loop = p->break_loop;	
+}
+static void dpdk_dispatch_inter(void *dpdk_user)
```
guyharris: This routine will be run in the thread that's calling `pcap_dispatch()`, `pcap_loop()`, `pcap_next()`, or `pcap_next_ex()`, right?  Libpcap doesn't support delivering packets asynchronously in a separate thread - packets are only supposed to be delivered when one of those routines are called.

pcap-dpdk.c
jingleyang: OK, the check if (p==NULL) will be removed safely.

pcap-dpdk.c
jingleyang: OK, I will check the rest char of device are all digital then to the strtoul.
The following bad cases will be rejected: 
dpdk:1e2
dpdk:1u

pcap-dpdk.c
jingleyang: 1. DPDK needs root permision.
2. Two kinds of error code (PCAP_ERROR_PERM_DENIED, PCAP_ERROR_IFACE_NOT_UP ) are add.
Cheers.

pcap-dpdk.c
jingleyang: OK, Thanks

pcap-dpdk.c
jingleyang: Yes, caplen = min(pkt_len, p->snapshot).
I have modified the code.

pcap-dpdk.c
jingleyang: This part of code is re-organized.

pcap-dpdk.c
jingleyang: This part of code is re-organized.

pcap-dpdk.c
jingleyang: It is a bug.
Regarding a very large packet, the beginning caplen of data will be processed.

pcap-dpdk.c
jingleyang: OK, the code have been modified.

pcap-dpdk.c
jingleyang: OK, p->break_loop will be used directly.

pcap-dpdk.c
jingleyang: Ok. Thanks.
dpdk_dispatch_internal will be called synchronously.

configure.ac
jingleyang: OK. I have updated configure.ac and CMakelists.txt

pcap-dpdk.c
jingleyang: yes

jingleyang:
Thank you so much for the code review. @guyharris 
I have modified the code and update the pull request.
Feel free to give me any feedback.
Merry Christmas and Happy New Year.

pcap-dpdk.c
```diff
@@ -0,0 +1,800 @@
...
+}
+
+static int pcap_dpdk_dispatch(pcap_t *p, int max_cnt, pcap_handler cb, u_char *pcap_user)
+{
+	unsigned lcore_id = 0;	
+	struct pcap_dpdk *pd = (struct pcap_dpdk*)(p->priv);
+	pd->rx_pkts=0;
+	pd->cb = cb;
+	pd->cb_arg = pcap_user;
+	pd->max_cnt = max_cnt;
```
guyharris: Why not just pass `max_cnt` as an argument to `dpdk_dispatch_internal()`?

pcap-dpdk.c
```diff
@@ -0,0 +1,800 @@
...
+	//only use master lcore
+	if (lcore_id != master_lcore_id){
+		return;
+	}
+	if (max_cnt>0 && max_cnt < MAX_PKT_BURST){
+		burst_cnt = max_cnt;
+	}else{
+		burst_cnt = MAX_PKT_BURST;
+	}
+	while( max_cnt==-1 || pkt_cnt < max_cnt){
```
guyharris: You should use `PACKET_COUNT_IS_UNLIMITED(max_cnt)` rather than just checking against -1; 0 should also mean "unlimited".

pcap-dpdk.c
```diff
@@ -0,0 +1,800 @@
...
+					large_buffer = (u_char *)malloc(caplen*sizeof(u_char));
+					gather_len = dpdk_gather_data(large_buffer, caplen, m); 
+					bp = large_buffer;
+				}
+				
+			}
+			if (bp){
+				//default accpet all
+				is_accepted=1;
+				if (pd->filter_in_userland && p->fcode.bf_insns!=NULL)
```
guyharris: Checking `pd->filter_in_userland`, and thus the `filter_in_userland` structure member, is unnecessary - if there's no filter, `p->fcode.bf_insns` will be null.

So you can do this more simply as

    if (p->fcode.bf_insns==NULL || pcap_filter(p->fcode.bf_insns, bp, pcap_header.len, pcap_header.caplen)){
    	cb(cb_arg, &pcap_header, bp);
    }else{
    	pd->bpf_drop++;
    }

and don't need the `is_accepted` variable.

pcap-dpdk.c
```diff
@@ -0,0 +1,800 @@
...
+	if(lcore_id == master_lcore_id){
+		RTE_LOG(DEBUG, USER1, "dpdk: lcoreid=%u runs for portid=%u\n", lcore_id, portid);
+	}else{
+		RTE_LOG(DEBUG, USER1, "dpdk: lcore %u has nothing to do\n", lcore_id);
+	}
+	//only use master lcore
+	if (lcore_id != master_lcore_id){
+		return;
+	}
+	if (max_cnt>0 && max_cnt < MAX_PKT_BURST){
```
guyharris: You could also use `PACKET_COUNT_IS_UNLIMITED()` here:

    if (! PACKET_COUNT_IS_UNLIMITED(max_cnt) && max_cnt < MAX_PKT_BURST){

pcap-dpdk.c
```diff
@@ -0,0 +1,800 @@
...
+	void *dpdk_user = p;	
+	dpdk_dispatch_internal(dpdk_user);	
+	return pd->rx_pkts;	
+}
+
+static int pcap_dpdk_inject(pcap_t *p, const void *buf _U_, int size _U_)
+{
+	//not implemented yet
+	pcap_fmt_errmsg_for_errno(p->errbuf, PCAP_ERRBUF_SIZE,
+		errno, "dpdk error: Inject function has not be implemented yet");
```
guyharris: That should be `"dpdk error: Inject function has not been implemented yet"`.

pcap-dpdk.c
```diff
@@ -0,0 +1,800 @@
...
+	uint32_t total_len = 0;
+	while (mbuf && (total_len+mbuf->data_len) < len ){
+		rte_memcpy(data+total_len, rte_pktmbuf_mtod(mbuf,void *),mbuf->data_len);
+		total_len+=mbuf->data_len;
+		mbuf=mbuf->next;
+	}
+	return total_len;
+}
+
+static void dpdk_dispatch_internal(void *dpdk_user)
```
guyharris: Where does this block waiting for packets to arrive?

pcap-dpdk.c
jingleyang: Yes. The code is simple and clean.

pcap-dpdk.c
jingleyang: OK. Thanks.

pcap-dpdk.c
jingleyang: It is cool.

pcap-dpdk.c
jingleyang: OK. Thanks.

pcap-dpdk.c
jingleyang: Opps. Thanks.

pcap-dpdk.c
jingleyang: The rte_eth_rx_burst() is key function to wait for packets to arrive.
nb_rx = (int)rte_eth_rx_burst(portid, 0, pkts_burst, burst_cnt);

pcap-dpdk.c
```diff
@@ -0,0 +1,800 @@
...
+	int cnt=0;
+	memset(dargv,0,sizeof(dargv[0])*DPDK_ARGC_MAX);	
+	//current process name
+	int skip_space = 1;
+	int i=0;
+	RTE_LOG(INFO, USER1,"dpdk cfg: %s\n",dpdk_cfg);
+	// find first non space char
+	// The last opt is NULL
+	for (i=0;dpdk_cfg[i] && cnt<DPDK_ARGC_MAX-1;i++){
+		if (skip_space && dpdk_cfg[i]!=0x20){ // not space
```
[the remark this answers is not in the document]
jingleyang: OK.

pcap-dpdk.c
```diff
@@ -0,0 +1,800 @@
...
+	{
+		return ret;
+	}
+	//check all chars are digital
+	for (int i=prefix_len; device[i]; i++){
+		if (device[i]<'0' || device[i]>'9'){
+			return ret;	
+		}
+	}
+	ret_ul = strtoul(&(device[prefix_len]), &pEnd, 10);
```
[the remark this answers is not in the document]
jingleyang: OK. Thanks.

pcap-dpdk.c
guyharris: So `rte_eth_rx_burst()` blocks, rather than returning 0, if there are no packets available?  [The documentation for `rte_eth_rx_burst()`](https://doc.dpdk.org/api/rte__ethdev_8h.html#a3e7d76a451b46348686ea97d6367f102) doesn't say how that works.

Is there a way to test whether there are any packets available, or to have `rte_eth_rx_burst()` not block?

Is there a way to get a UN\*X file descriptor on which you can do a `select()`, `poll()`, `epoll()`, etc. in order to wait until there are packets available?

pcap-dpdk.c
```diff
@@ -0,0 +1,800 @@
...
+	char *ptr_dpdk_cfg = NULL;
+	int ret = PCAP_ERROR; 
+	// globale var
+	if (is_dpdk_pre_inited)
+	{
+		// already inited
+		return 0;
+	}
+	// check for root permission
+	if( geteuid() != 0)
```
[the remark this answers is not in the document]
jingleyang: Yes. DPDK need root permission.
The following error msg is attached when I run dpdk-18.11/examples/l2fwd

```
EAL: Probing VFIO support...
EAL: Couldn't get fd on hugepage file
EAL: Couldn't get fd on hugepage file
error allocating rte services array
EAL: FATAL: rte_service_init() failed
EAL: rte_service_init() failed
EAL: Error - exiting with code: 1
```

pcap-dpdk.c
```diff
@@ -0,0 +1,800 @@
...
+		for (int i=0; i<nb_ports; i++){
+			pcap_snprintf(dpdk_name,DPDK_DEV_NAME_MAX-1,"dpdk:%d",i);
+			// mac addr 
+			rte_eth_macaddr_get(i, &eth_addr);
+			eth_addr_str(&eth_addr,mac_addr,DPDK_MAC_ADDR_SIZE);	
+			// PCI addr
+			rte_eth_dev_get_name_by_port(i,pci_addr);
+			pcap_snprintf(dpdk_desc,DPDK_DEV_DESC_MAX-1,"%s %s, MAC:%s, PCI:%s", DPDK_DESC, dpdk_name, mac_addr, pci_addr);
+			// continue add all dev, even error happens
+			add_dev(devlistp, dpdk_name, 0, dpdk_desc, ebuf);
```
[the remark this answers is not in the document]
jingleyang: OK. Thanks.

pcap-dpdk.c
```diff
@@ -0,0 +1,800 @@
...
+		p->fd = pd->portid; 
+		if (p->snapshot <=0 || p->snapshot> MAXIMUM_SNAPLEN)
+		{
+			p->snapshot = MAXIMUM_SNAPLEN;
+		}
+		p->linktype = DLT_EN10MB; // Ethernet, the 10MB is historical.
+		p->selectable_fd = p->fd;
+		p->read_op = pcap_dpdk_dispatch;
+		p->inject_op = pcap_dpdk_inject;
+		// DPDK only support filter in userland now
```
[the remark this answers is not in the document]
jingleyang: 1. DPDK currently provides their own BPF API by rte_bpf.h. And it is also a user space solution.
https://doc.dpdk.org/guides/prog_guide/bpf_lib.html
2. However, DPDK will load BPF code from a ELF file by calling rte_bpf_elf_load(elf_file_name). It is difficult to use, because user needs compile BPF elf firstly, then pass it to DPDK programs.
https://www.dpdk.org/wp-content/uploads/sites/35/2018/10/pm-07-DPDK-BPFu6.pdf
3. Currently, I prefer use pcap_filter instead of rte_bpf.h. And later, if DPDK provides a user friendly API to compile BPF from string and make filtering more efficient, I will consider to change to rte_bpf.h
4. I will change the comment to make it clear.

jingleyang:
hi @guyharris ,
Thank you very much for the second round code review.
I have changed the code and pull request is updated.
Any feedback is welcome.
Cheers.

pcap-dpdk.c
```diff
@@ -0,0 +1,761 @@
...
+struct dpdk_ts_helper{
+	struct timeval start_time;
+	uint64_t start_cycles;
+	uint64_t hz;
+};
+struct pcap_dpdk{
+	pcap_t * orig;
+	uint16_t portid; // portid of DPDK
+	int must_clear_promisc;
+	int filter_in_userland;
```
guyharris: `filter_in_userland` isn't used any more, so you can remove it.

pcap-dpdk.c
```diff
@@ -0,0 +1,761 @@
...
+		p->fd = pd->portid; 
+		if (p->snapshot <=0 || p->snapshot> MAXIMUM_SNAPLEN)
+		{
+			p->snapshot = MAXIMUM_SNAPLEN;
+		}
+		p->linktype = DLT_EN10MB; // Ethernet, the 10MB is historical.
+		p->selectable_fd = p->fd;
+		p->read_op = pcap_dpdk_dispatch;
+		p->inject_op = pcap_dpdk_inject;
+		pd->filter_in_userland = 1;
```
guyharris: `filter_in_userland` isn't used any more, so you don't need to set it.

jingleyang:
Thank you for the code review.
I have update the PR and the code support non-blocking, blocking with (timeout or forever).
Cheers.

pcap-dpdk.c
```diff
@@ -0,0 +1,808 @@
...
+		// read once in non-blocking mode, or try many times waiting for timeout_ms.
+		// if timeout_ms == 0, it will be blocked until one packet arrives or break_loop is setted.
+		nb_rx = dpdk_read_with_timeout(p, portid, 0, pkts_burst, burst_cnt);
+		if (nb_rx == 0){ 
+			if (pd->nonblock){
+				RTE_LOG(DEBUG, USER1, "dpdk: no packets available in non-blocking mode.\n");
+			}else{
+				if (p->break_loop){
+					RTE_LOG(DEBUG, USER1, "dpdk: no packets available and break_loop is setted in blocking mode.\n");
+					p->break_loop = 1;
```
guyharris: This should set `p->break_loop` to 0, not 1.  It's *already* 1; it should be cleared.

pcap-dpdk.c
```diff
@@ -0,0 +1,808 @@
...
+struct dpdk_ts_helper{
+	struct timeval start_time;
+	uint64_t start_cycles;
+	uint64_t hz;
+};
+struct pcap_dpdk{
+	pcap_t * orig;
+	uint16_t portid; // portid of DPDK
+	int must_clear_promisc;
+	uint64_t rx_pkts;
```
guyharris: `rx_pkts` can just be a local variable in `pcap_dpdk_dispatch()`.

pcap-dpdk.c
```diff
@@ -0,0 +1,808 @@
...
+				// sleep for a very short while, but do not block CPU.
+				rte_delay_us_block(DPDK_DEF_MIN_SLEEP_MS*1000);
+				sleep_ms += DPDK_DEF_MIN_SLEEP_MS;
+			}
+		}
+	}
+	return nb_rx;
+}
+
+static int pcap_dpdk_dispatch(pcap_t *p, int max_cnt, pcap_handler cb, u_char *cb_arg)
```
guyharris: `rx_pkts` should just be a local variable here.

pcap-dpdk.c
```diff
@@ -0,0 +1,808 @@
...
+	uint64_t start_cycles;
+	uint64_t hz;
+};
+struct pcap_dpdk{
+	pcap_t * orig;
+	uint16_t portid; // portid of DPDK
+	int must_clear_promisc;
+	uint64_t rx_pkts;
+	uint64_t bpf_drop;
+	int nonblock;
```
guyharris: There should be an additional member `struct timeval required_select_timeout` here.  It will be set below.

pcap-dpdk.c
```diff
@@ -0,0 +1,808 @@
...
+		p->inject_op = pcap_dpdk_inject;
+		// using pcap_filter currently, though DPDK provides their own BPF function. Because DPDK BPF needs load a ELF file as a filter.
+		p->setfilter_op = install_bpf_program;
+		p->setdirection_op = NULL;
+		p->set_datalink_op = NULL;
+		p->getnonblock_op = pcap_dpdk_getnonblock;
+		p->setnonblock_op = pcap_dpdk_setnonblock;
+		p->stats_op = pcap_dpdk_stats;
+		p->cleanup_op = pcap_dpdk_close;
+		p->breakloop_op = pcap_breakloop_common;
```
guyharris: Here's where you set, and use, the `required_select_timeout` member of the private data structure.

Do

    pd->required_select_timeout.tv_sec = 0;
    pd->required_select_timeout.tv_usec = DPDK_DEF_MIN_SLEEP_MS*1000;
    p->required_select_timeout = &pd->required_select_timeout;

This way, the `pcap_get_required_select_timeout()` call will return that timeout, so that if code using pcap has a `select()`/`poll()`/`epoll()`/etc. loop, waiting for, among other input sources, a `pcap_t`, it can use that timeout to poll a DPDK `pcap_t`.  (This mechanism was introduced for DAG cards, but DPDK, like DAG, works by polling, rather than with interrupts and wakeups.)

pcap-dpdk.c
jingleyang: Yes. It's a bug.

pcap-dpdk.c
jingleyang: OK. No problem.

pcap-dpdk.c
jingleyang: OK. No problem.

pcap-dpdk.c
jingleyang: OK.

pcap-dpdk.c
jingleyang: OK. Thanks.

jingleyang:
Thank you for the 4th round code review.
The PR has been update.
Keep in touch.
Cheers.

configure
```diff
@@ -1440,6 +1446,10 @@ Optional Packages:
   --with-turbocap[=DIR]   include Riverbed TurboCap support (located in
                           directory DIR, if supplied). [default=yes, if
                           present]
+  --with-dpdk-includes=IDIR
```
guyharris: Will the DPDK include and library directories ever be anything other than the `include` and `lib` subdirectories of some particular higher-level directory, such as `/usr/local/include` and `/usr/local/lib`, or `/opt/dpdk/include` and `/opt/dpdk/lib`, or something such as that?

jingleyang: 1. You could find the system requirement about DPDK. https://doc.dpdk.org/guides/linux_gsg/sys_reqs.html

pcap-dpdk.c
```diff
@@ -0,0 +1,811 @@
...
+			rte_prefetch0(rte_pktmbuf_mtod(m, void *));
+			bp = NULL;
+			if (m->nb_segs == 1)
+			{
+				bp = rte_pktmbuf_mtod(m, u_char *);
+			}else{
+				// use fast buffer pcap_tmp_buf if pkt_len is small, no need to call malloc and free
+				if ( pkt_len <= ETHER_MAX_JUMBO_FRAME_LEN)
+				{
+					gather_len = dpdk_gather_data(pd->pcap_tmp_buf, RTE_ETH_PCAP_SNAPLEN, m);
```
AndersBroman: Discaimer: I'm do not know much about dpdk, my assumtion making this comment is that the packet(s) are in linked mbuffs and a packet may span 2 consecutive bugffers.

I think "zero-copy" would be much faster e.g pass the mbuff to the caller and let the writing code handle the case of a packet continuing in the next buffer. Reading about dpdk I got the impression that it was possible to create a header section in the buffer before the actual packet. If that is true it would be very efficent to leave space for a pcap/pcap-ng packet header in the buffer in that way the whol buffer co0uld be written to file in one go.

AndersBroman: This link may be interesting too https://www.napatech.com/dpdk-packet-capture-pdump/

configure
guyharris: Then why not just handle DPDK the same way other optional modules are handled, with a `--with-dpdk` option that takes an optional "DIR" argument, with the header directory being `{DIR}/include` and the library directory being `{DIR}/lib`, so that:

- if you don't specify `--with-dpdk`, it tries to find DPDK in `/usr` and `/usr/local` and, if it finds them, builds with DPDK, otherwise doesn't build with DPDK;
- if you specify `--with-dpdk`, it tries to find DPDK in `/usr` and `/usr/local` and, if it finds them, builds with DPDK, otherwise fails;
- if you specify `--with-dpdk=DIR`, it tries to find DPDK in DIR and, if it finds them, builds with DPDK, otherwise fails;
- if you specify `--without-dpdk`, it doesn't try to find DPDK and doesn't build with DPDK?

For CMake, have a `cmake/Modules/FindDPDK.cmake` module.

pcap-dpdk.c
guyharris: That article claims

> Basically the pcap_dump() function does a fwrite(pcap header of 16Bytes) followed by a fwrite(packet payload) and these two fwrite(s) cause two kernel traps, which are expensive and should be minimized.

which is an over-simplification.  `fwrite()` does not necessarily cause a kernel trap; with the standard I/O library, a kernel trap occurs only if it does a `write()` (on UN\*X) or an `NtWriteFile()` (on Windows; that will come from a Win32 `WriteFile()`, which would be done by an `_write()`), which will happen only in an `fwrite()` that requires that the buffer be emptied to make enough room to copy the data.

The typical buffer size for the standard I/O library is, these days, probably at least 4KiB, which is sufficient for at least one pcap record for a maximum-size Ethernet packet.  libpcap could increase that buffer size by making a `setvbuf()` call, which would reduce the number of system calls done.  I've sent the author fo that document a e-mail pointing that out and asking how big the buffer in question was.

guyharris: > I think "zero-copy" would be much faster e.g pass the mbuff to the caller and let the writing code handle the case of a packet continuing in the next buffer.

It's more like "...to the *callee*"; libpcap's core captured-packet-processing loops pass the packets to a callback routine.  Those routines are expecting the packet data to be in a contiguous array of size equal to the captured length, so if the packet is split between mbufs, it would have to be coalesced into a single buffer.

If the goal is to have a special fast path for dumping, that would have to be done through additional pcap APIs, e.g. a `pcap_dispatch_and_dump()` API that takes a `pcap_t` and a `pcap_dumper_t` as arguments and dumps each packet to the `pcap_dumper_t`.  That would possibly involve running a different captured-packet-processing loop for capture modules where packet coalescing is done (for other modules, `pcap_dispatch_and_dump()` would just be something such as

    int
    default_pcap_dispatch_and_dump_handler(pcap_t *p, int cnt, pcap_dumper_t *pdmp)
    {
       return pcap_dispatch(p, cnt, pcap_dump, pdmp);
    }

so that we don't need two loops for *all* pcap-XXX.c files).

configure
jingleyang: OK. I will try to do that on the weekend.

jingleyang: I rewrite the configuration files with your suggestion. Thanks.

configure
```diff
@@ -6691,6 +6696,15 @@ dag)
 	xxx_only=yes
 	;;
 
+dpdk)
+	#
+	# --with-pcap=dpdk is the only way to get here, and it means
+	# "DAG support but nothing else"
```
guyharris: You mean `"DPDK" support but nothing else"` here.

guyharris: But it's correct in `configure.ac`; perhaps you just need to run `autoreconf` and check in the result.

guyharris: I've done that.

Also, pcap-dpdk.c and pcap-dpdk.h need to be added to EXTRA_DIST, but I'll handle that as well.

jingleyang: Cool, Thank you so much for the help and the code review.

guyharris: > Cool, Thank you so much for the help and the code review.

And thank you for the code!

missyoyo:
I have test tcpdump witch libcap dpdk enable and can caputre dpdk:0 pcaket.
And I have a question:
Can diffrent program listening dpdk:0 in the same time.
I open two shell and run two tcpdump -i dpdk:0 with dpdk env:
The fist one works good
The second tcpdump get error below:
EAL: Cannot create lock on '/var/run/dpdk/rte/config'. Is another primary process running?
I try to run second tcpdump with out dpdk env
tcpdump -i dpdk:0
EAL: Cannot create lock on '/var/run/dpdk/rte/config'. Is another primary process running?
EAL: FATAL: Cannot init config
EAL: Cannot init config
tcpdump: Can't open device dpdk:0: dpdk error: dpdk_pre_init failed: Success
still get same error. So can two diffrent program listening dpdk:0 in the same time?Thank you

missyoyo:
Hi:
    I got a error when run ./capturetest
env DPDK_CFG="--log-level=debug -l0 -dlibrte_pmd_ixgbe.so -dlibrte_mempool_ring.so" ./capturetest -i dpdk:0
ERROR MESSAGE:
"ixgbe_dev_tx_queue_setup(): TX WTHRESH must be set to 0 if tx_rs_thresh is greater than 1. (tx_rs_thresh=8 port=0 queue=0)"
May I know how to set TX WTHRESH to 0 when run  ./capturetest ?
Thank you.

PR #793, opened 2019-01-09
Make CMake work for an installed DPDK on Ubuntu.

AndersBroman:
DPDK (EAL: RTE Version: 'DPDK 18.11.0')can be installed on Ubuntu 18.04 by adding this line
 to /etc/apt/sources.list
deb [trusted=yes] http://ftp.se.debian.org/debian/ testing main contrib
and doing sudo apt-get install dpdk dpdk-dev

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1427,34 +1427,14 @@ endif()
 
 # Check for DPDK sniffing support
 if (NOT DISABLE_DPDK)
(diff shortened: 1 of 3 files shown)


guyharris:
Please add `cmake/Modules/Finddpdk.cmake` to the `EXTRA_DIST` list in the top-level `Makefile.in`.

AndersBroman:
Done

PR #796, opened 2019-01-21
Add support for DSA link-layer types

ffainelli:
Linux kernel 4.20 and greater can report what type of Distributed Switch
Architecture tagging protocol is used on the DSA master/management
interface. We need to map the protocol to a specific DLT and linktype
value in the pcap file because these protocols typically cannot be
decoded simply by making use of heuristics. The sysfs attribute that is
being checked and parsed is documented in this commit:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=a3d7e01da06013dc580641a1da57c3b482d58157

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -332,7 +332,6 @@ struct pcap_linux {
 #ifdef HAVE_SYS_EVENTFD_H
 	int poll_breakloop_fd; /* fd to an eventfd to break from blocking operations */
 #endif
-
 };
 
 /*
@@ -505,6 +504,9 @@ static struct sock_filter	total_insn
 static struct sock_fprog	total_fcode
 	= { 1, &total_insn };
 #endif /* SO_ATTACH_FILTER */
+
+static int	iface_dsa_get_proto_info(const char *device, pcap_t *handle,
+    char *ebuf);
 
 pcap_t *
 pcap_create_interface(const char *device, char *ebuf)
@@ -3298,6 +3300,28 @@ static void map_arphrd_to_dlt(pcap_t *handle, int sock_fd, int arptype,
 		 * others?
 		 */
 		if (!is_wifi(sock_fd, device)) {
+			int ret;
+
+			/*
+			 * This is not a Wi-Fi device but it could be
+			 * a DSA master/management network device.
+			 */
+			ret = iface_dsa_get_proto_info(device, handle,
+						       handle->errbuf);
+			if (ret < 0)
+				return;
+
+			if (ret == 1) {
+				/*
+				 * This is a DSA master/management network
+				 * device linktype is already set by
+				 * iface_dsa_get_proto_info() set an
+				 * appropriate offset here.
+				 */
+				handle->offset = 2;
+				break;
+			}
+
 			/*
 			 * It's not a Wi-Fi device; offer DOCSIS.
 			 */
@@ -6899,6 +6923,78 @@ iface_get_offload(pcap_t *handle _U_)
 #endif /* HAVE_PACKET_RING */
 
 #endif /* HAVE_PF_PACKET_SOCKETS */
+
+static struct dsa_proto {
+	const char *name;
+	bpf_u_int32 linktype;
+} dsa_protos[] = {
+	/*
+	 * None is special and indicates that the interface does not have
+	 * any tagging protocol configured, and is therefore a standard
+	 * Ethernet interface.
+	 */
+	{ "none", DLT_EN10MB },
+	{ "brcm", DLT_DSA_TAG_BRCM },
+	{ "brcm-prepend", DLT_DSA_TAG_BRCM_PREPEND },
+};
+
+static int	iface_dsa_get_proto_info(const char *device, pcap_t *handle,
+    char *ebuf)
+{
+	char *pathstr;
+	unsigned int i;
+	char buf[256];
+	ssize_t r;
+	int fd;
+
+	fd = asprintf(&pathstr, "/sys/class/net/%s/dsa/tagging", device);
+	if (fd < 0) {
+		pcap_fmt_errmsg_for_errno(handle->errbuf, PCAP_ERRBUF_SIZE,
+					  fd, "asprintf");
+		return PCAP_ERROR;
+	}
+
+	fd = open(pathstr, O_RDONLY);
+	free(pathstr);
+	/*
+	 * This is not fatal, kernel >= 4.20 *might* expose this attribute
+	 */
+	if (fd < 0)
+		return 0;
+
+	r = read(fd, buf, sizeof(buf) - 1);
+	close(fd);
+	if (r <= 0) {
+		pcap_fmt_errmsg_for_errno(handle->errbuf, PCAP_ERRBUF_SIZE,
+					  r, "read");
+		return PCAP_ERROR;
+	}
+
+	/*
+	 * Buffer should be LF terminated.
+	 */
+	if (buf[r - 1] == '\n')
+		r--;
+	buf[r] = '\0';
+
+	for (i = 0; i < sizeof(dsa_protos) / sizeof(dsa_protos[0]); i++) {
+		if (strlen(dsa_protos[i].name) == r &&
+		    !strcmp(buf, dsa_protos[i].name)) {
+			handle->linktype = dsa_protos[i].linktype;
+			switch (dsa_protos[i].linktype) {
+			case DLT_EN10MB:
+				return 0;
+			default:
+				return 1;
+			}
+		}
+	}
+
+	pcap_snprintf(handle->errbuf, PCAP_ERRBUF_SIZE,
+		      "unsupported DSA tag: %s", buf);
+
+	return PCAP_ERROR;
+}
 
 /* ===== Functions to interface to the older kernels ================== */
 
--- a/pcap.c
+++ b/pcap.c
@@ -3002,6 +3002,8 @@ static struct dlt_choice dlt_choices[] = {
 	DLT_CHOICE(LINUX_SLL2, "Linux cooked v2"),
 	DLT_CHOICE(OPENVIZSLA, "OpenVizsla USB"),
 	DLT_CHOICE(EBHSCR, "Elektrobit High Speed Capture and Replay (EBHSCR)"),
+	DLT_CHOICE(DSA_TAG_BRCM, "Broadcom tag"),
+	DLT_CHOICE(DSA_TAG_BRCM_PREPEND, "Broadcom tag (prepended)"),
 	DLT_CHOICE_SENTINEL
 };
 
--- a/pcap/dlt.h
+++ b/pcap/dlt.h
@@ -1427,6 +1427,12 @@
 #define DLT_VPP_DISPATCH	280
 
 /*
+ * Broadcom Ethernet switches (ROBO switch) 4 bytes proprietary tagging format.
+ */
+#define DLT_DSA_TAG_BRCM	281
+#define DLT_DSA_TAG_BRCM_PREPEND	282
+
+/*
  * In case the code that includes this file (directly or indirectly)
  * has also included OS files that happen to define DLT_MATCHING_MAX,
  * with a different value (perhaps because that OS hasn't picked up
@@ -1436,7 +1442,7 @@
 #ifdef DLT_MATCHING_MAX
 #undef DLT_MATCHING_MAX
 #endif
-#define DLT_MATCHING_MAX	280	/* highest value in the "matching" range */
+#define DLT_MATCHING_MAX	282	/* highest value in the "matching" range */
 
 /*
  * DLT and savefile link type values are split into a class and
--- a/pcap-common.c
+++ b/pcap-common.c
@@ -1141,7 +1141,13 @@
  */
 #define LINKTYPE_VPP_DISPATCH	280
 
-#define LINKTYPE_MATCHING_MAX	280		/* highest value in the "matching" range */
+/*
+ * Broadcom Ethernet switches (ROBO switch) 4 bytes proprietary tagging format.
+ */
+#define LINKTYPE_DSA_TAG_BRCM	281
+#define LINKTYPE_DSA_TAG_BRCM_PREPEND	282
+
+#define LINKTYPE_MATCHING_MAX	282		/* highest value in the "matching" range */
 
 /*
  * The DLT_ and LINKTYPE_ values in the "matching" range should be the


pcap-linux.c
```diff
@@ -3298,6 +3300,26 @@ static void map_arphrd_to_dlt(pcap_t *handle, int sock_fd, int arptype,
...
+
+			/* This is not a Wi-Fi device but it could be
+			 * a DSA master/management network device.
+			 */
+			ret = iface_dsa_get_proto_info(device, handle,
+						       handle->errbuf);
+			if (ret < 0)
+				return;
+
+			/* This is a DSA master/management network device
```
guyharris: That moment should go inside the `if (ret == 1) {` block, as the statement in the comment is true only if `iface_dsa_get_proto_info()` returned 1.

pcap-linux.c
```diff
@@ -6900,6 +6922,70 @@ iface_get_offload(pcap_t *handle _U_)
...
+		if (!strncmp(buf, dsa_protos[i].name,
+			     strlen(dsa_protos[i].name)) &&
+		    dsa_protos[i].linktype) {
+			handle->linktype = dsa_protos[i].linktype;
+
+			return 1;
+		}
+	}
+
+	pcap_strlcpy(handle->errbuf, "unsupported DSA tag\n",
```
guyharris: You might want to use `pcap_snprintf()` and include the tag string in the message, to make it easier to fix the problem given the error message.

Also, error messages should *not* end in a newline.

pcap-linux.c
```diff
@@ -6900,6 +6922,70 @@ iface_get_offload(pcap_t *handle _U_)
...
+	r = read(fd, buf, sizeof(buf));
+	close(fd);
+	if (r <= 0) {
+		pcap_fmt_errmsg_for_errno(handle->errbuf, PCAP_ERRBUF_SIZE,
+					  r, "read");
+		return PCAP_ERROR;
+	}
+
+	for (i = 0; i < sizeof(dsa_protos) / sizeof(dsa_protos[0]); i++) {
+		if (!strncmp(buf, dsa_protos[i].name,
```
guyharris: If the tag for the device is longer than the name in the table, this will do prefix matching; is that what it should do, or should it do an exact match?

If the tag for the device is *shorter* than the name in the table, this will run past the end of the tag string, as it hasn't been null-terminated.

You should probably:

- make the buffer 255+1 bytes, with the +1 for a terminating null character, read into the first 255 bytes, and set `buf[r]` to `'\0'`;
- before the `strncmp()` call, check whether `strlen(dsa_protos[i].name)` is equal to `r`, and then do a `strcmp()` call between `buf` and `dsa_protos[i].name` if the lengths are equal.

pcap-linux.c
ffainelli: Will do, thanks.

guyharris:
We'd need detailed descriptions for the tag and its position in the packet for all of the new `DLT_` values.

ffainelli:
> We'd need detailed descriptions for the tag and its position in the packet for all of the new `DLT_` values.

Since I can only provide such a detailed description for the two types of Broadcom tags, should I just not reserve a DLT_* number for all other types?

guyharris:
> > We'd need detailed descriptions for the tag and its position in the packet for all of the new `DLT_` values.
> 
> Since I can only provide such a detailed description for the two types of Broadcom tags, should I just not reserve a DLT_* number for all other types?

For now, yes.  The other tags can be added as descriptions become available.  We can add an error for types for which no DLT_ is available for now.

ffainelli:
> > > We'd need detailed descriptions for the tag and its position in the packet for all of the new `DLT_` values.
> > 
> > 
> > Since I can only provide such a detailed description for the two types of Broadcom tags, should I just not reserve a DLT_* number for all other types?
> 
> For now, yes. The other tags can be added as descriptions become available. We can add an error for types for which no DLT_ is available for now.

That would be breaking backwards compatibility. Is there a standard way to issue a warning with libpcap, unless printing to stdout/stderr is acceptable for that case?

ffainelli:
Forced push an update to the same branch which contains full Broadcom tag description (per email conversation with Guy) as well as suggested changes.

guyharris:
> > For now, yes. The other tags can be added as descriptions become available. We can add an error for types for which no DLT_ is available for now.
> 
> That would be breaking backwards compatibility.

Having `pcap_activate()` return an error if you try to open a DSA device for which we haven't yet gotten around to writing up a description, rather than returning a `pcap_t *` with the wrong DLT_ type, breaks binary compatibility?

ffainelli:
> > > For now, yes. The other tags can be added as descriptions become available. We can add an error for types for which no DLT_ is available for now.
> > 
> > 
> > That would be breaking backwards compatibility.
> 
> Having `pcap_activate()` return an error if you try to open a DSA device for which we haven't yet gotten around to writing up a description, rather than returning a `pcap_t *` with the wrong DLT_ type, breaks binary compatibility?

It does not, I keep forgetting that map_arphrd_to_dlt() returns void and that we would simply fall back to DLT_EN10MB if iface_dsa_get_proto_info() returns < 0, therefore we are fine no matter what. I will make a few changes and force push an update.

ffainelli:
Here is another version.

guyharris:
Do any of the switch chips in question describe the tag format in publicly-available documentation?  I know BRCM used to bury a lot of their documentation in Henry Nicholas's basement dungeon, but other chip vendors might have published that stuff without an NDA.

If so, that might be a sufficient reference for the tcpdump.org link-layer header types page.

pcap.c
```diff
@@ -3002,6 +3002,8 @@ static struct dlt_choice dlt_choices[] = {
 	DLT_CHOICE(LINUX_SLL2, "Linux cooked v2"),
 	DLT_CHOICE(OPENVIZSLA, "OpenVizsla USB"),
 	DLT_CHOICE(EBHSCR, "Elektrobit High Speed Capture and Replay (EBHSCR)"),
+	DLT_CHOICE(DSA_TAG_BRCM, "Broadcom tag"),
+	DLT_CHOICE(DSA_TAG_BRCM, "Broadcom tag (prepended)"),
```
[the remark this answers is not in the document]
ffainelli: > You mean `DLT_DSA_TAG_BRCM_PREPEND` for the second item.

Indeed, thanks.

ffainelli:
> Do any of the switch chips in question describe the tag format in publicly-available documentation? I know BRCM used to bury a lot of their documentation in Henry Nicholas's basement dungeon, but other chip vendors might have published that stuff without an NDA.
> 
> If so, that might be a sufficient reference for the tcpdump.org link-layer header types page.

For the usual highly cooperative open source players, the best documentation is under net/dsa/tag_*.c unfortunately. The only public documentation that I know of is for Marvell's 88E6060 LinkStreet which is nearing 15 years old. Here are a few that I just quickly glanced over:

http://ww1.microchip.com/downloads/en/DeviceDoc/00002308A.pdf
http://ww1.microchip.com/downloads/en/devicedoc/00002392a.pdf

guyharris:
> For the usual highly cooperative open source players, the best documentation is under net/dsa/tag_*.c unfortunately.

One of the fundamental goals for

http://www.tcpdump.org/linktypes.html

is to allow people to write code to parse link-layer headers without having to read anybody else's code that parses them first; it's also there to allow them to do that without having to read anybody else's code that *generates* them first, so code that generates them should either 1) be used to generate a complete non-procedural description of the header or 2) be expanded to include such a description.

See, for example, the entry for `LINKTYPE_USB_LINUX_MMAPPED`/`DLT_USB_LINUX_MMAPPED` on that page; it refers to the `Documentation/usb/usbmon.txt` file in the Linux source tree, which has a definition of `struct usbmon_packet`.

> The only public documentation that I know of is for Marvell's 88E6060 LinkStreet which is nearing 15 years old. Here are a few that I just quickly glanced over:
> 
> http://ww1.microchip.com/downloads/en/DeviceDoc/00002308A.pdf
> http://ww1.microchip.com/downloads/en/devicedoc/00002392a.pdf

If those have sections that describe the tags and where they appear in the packet, the item for those `LINKTYPE_`/`DLT_` values could refer to that document, indicating the section or pages where the description appears.

ffainelli:
> > For the usual highly cooperative open source players, the best documentation is under net/dsa/tag_*.c unfortunately.
> 
> One of the fundamental goals for
> 
> http://www.tcpdump.org/linktypes.html
> 
> is to allow people to write code to parse link-layer headers without having to read anybody else's code that parses them first; it's also there to allow them to do that without having to read anybody else's code that _generates_ them first, so code that generates them should either 1) be used to generate a complete non-procedural description of the header or 2) be expanded to include such a description.
> 
> See, for example, the entry for `LINKTYPE_USB_LINUX_MMAPPED`/`DLT_USB_LINUX_MMAPPED` on that page; it refers to the `Documentation/usb/usbmon.txt` file in the Linux source tree, which has a definition of `struct usbmon_packet`.

Understood, which is why I tried to respond to your questions regarding the Broadcom tag format. I am not trying to say that "go look at the Linux code" is the way to go, but I am also not able to twist other silicon vendors' hand, so I can only help pave a road here to let them easily support their proprietary Ethernet switch tagging protocol in a way that hopefully requires a one liner change.

> 
> > The only public documentation that I know of is for Marvell's 88E6060 LinkStreet which is nearing 15 years old. Here are a few that I just quickly glanced over:
> > http://ww1.microchip.com/downloads/en/DeviceDoc/00002308A.pdf
> > http://ww1.microchip.com/downloads/en/devicedoc/00002392a.pdf
> 
> If those have sections that describe the tags and where they appear in the packet, the item for those `LINKTYPE_`/`DLT_` values could refer to that document, indicating the section or pages where the description appears.

If someone at Microchip wants to contribute a dissector for their tagging format, I will point them to our discussion and encourage them.

With everything that we have discussed so far, is this pull request in its current shape in an acceptable form to be included, or is there something else that I should be taking care of towards an inclusion?

pcap/dlt.h
```diff
@@ -1426,6 +1426,121 @@
  */
 #define DLT_VPP_DISPATCH	280
 
+/*
+ * Broadcom Ethernet switches (ROBO switch) 4 bytes proprietary tagging
+ * format.
+ *
```
guyharris: I've checked in a change to the link-layer header types page on tcpdump.org to add the two Broadcom switch tag LINKTYPE_/DLT_ values, along with a page that describes the Broadcom switch tag formats, so you don't need to have the switch tag format description here any more.  Those changes should show up on the Web site within 24 hours or so.

pcap-linux.c
```diff
@@ -6900,6 +6924,79 @@ iface_get_offload(pcap_t *handle _U_)
...
+					  r, "read");
+		return PCAP_ERROR;
+	}
+
+	/*
+	 * Buffer should be LF terminated.
+	 */
+	if (buf[r - 1] == '\n') {
+		r--;
+		buf[r] = '\0';
```
guyharris: This should probably be

    	if (buf[r - 1] == '\n')
    		r--;
    	buf[r] = '\0';

so that `buf` is guaranteed to be null-terminated.

ffainelli: Indeed, thanks

mcr:
Looks great, thank you!

fxlb:
FYI, news warnings:
```
./pcap-linux.c: In function ‘iface_dsa_get_proto_info’:
./pcap-linux.c:6981:34: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
   if (strlen(dsa_protos[i].name) == r &&
                                  ^
./pcap-linux.c:6942:11: warning: unused parameter ‘ebuf’ [-Wunused-parameter]
     char *ebuf)
           ^
```
(https://travis-ci.org/the-tcpdump-group/libpcap/jobs/483694166)

guyharris:
> FYI, news warnings:

The first is fixed by 87c3b959236479979ec262f69890a10caac7e24f; the second is fixed by 1a83bb6703bdcb3f07433521d95b788fa2ab4825.

ffainelli:
Thanks for fixing both warnings.

fxlb:
1 more new warning with clang  7.0.1:
```
./pcap-linux.c:6967:8: warning: implicit conversion loses integer precision: 'ssize_t' (aka 'long') to 'int' [-Wshorten-64-to-32]
                                          r, "read");
                                          ^
```

guyharris:
It shouldn't be passing `r`, it should be passing `errno` - `read()` returns a positive non-zero byte count on success, zero on EOF, and -1 on error, and the error number is in `errno`.

Fixed in d40631536d0cde11a6dacdcda7cdd8e7d56ac1d9.

ffainelli:
On 1/23/19 11:05 AM, Guy Harris wrote:
> *@guyharris* commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In pcap/dlt.h
> <https://github.com/the-tcpdump-group/libpcap/pull/796#discussion_r250331119>:
> 
>> @@ -1426,6 +1426,121 @@
>   */
>  #define DLT_VPP_DISPATCH	280
>  
> +/*
> + * Broadcom Ethernet switches (ROBO switch) 4 bytes proprietary tagging
> + * format.
> + *
> 
> I've checked in a change to the link-layer header types page on
> tcpdump.org to add the two Broadcom switch tag LINKTYPE_/DLT_ values,
> along with a page that describes the Broadcom switch tag formats, so you
> don't need to have the switch tag format description here any more.
> Those changes should show up on the Web site within 24 hours or so.

They do not seem to appear on tcpdump.org, do you expect them to show up
in the near future?
-- 
Florian

guyharris:
> They do not seem to appear on tcpdump.org, do you expect them to show up in the near future? -- Florian

Oops.  It helps if one remembers to push one's changes....

vivien:
Hi guys, just a notice that the link https://www.tcpdump.org/linktypes/broadcom-switch-tag.html has the same HTML title (as seen in the tab) as https://www.tcpdump.org/linktypes/LINKTYPE_LINUX_SLL.html (from which it seems to be based on.)

guyharris:
> Hi guys, just a notice that the link https://www.tcpdump.org/linktypes/broadcom-switch-tag.html has the same HTML title (as seen in the tab) as https://www.tcpdump.org/linktypes/LINKTYPE_LINUX_SLL.html (from which it seems to be based on.)

That's a Web site issue; those should be filed in [the GitHub repository for the Web site](https://github.com/the-tcpdump-group/tcpdump-htdocs).

That, and some other issues, are fixed in the-tcpdump-group/tcpdump-htdocs@1114a971bf3c8b5971e4af86d29aa920953e09b4.

PR #797, opened 2019-02-05
Fixes integer (and then buffer) overflow in rpcapd

catenacyber:
On a 32-bit system, this integer overflow will result in a buffer overflow a few lines later with `username[usernamelen] = '\0';`

### DIFF
--- a/rpcapd/daemon.c
+++ b/rpcapd/daemon.c
@@ -1234,6 +1234,11 @@ daemon_msg_auth_req(struct daemon_slpars *pars, uint32 plen)
 			uint32 usernamelen, passwdlen;
 
 			usernamelen = ntohs(auth.slen1);
+			if (usernamelen >= 0xFFFFFFFF) {
+				pcap_fmt_errmsg_for_errno(errmsgbuf,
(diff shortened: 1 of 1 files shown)


rpcapd/daemon.c
```diff
@@ -1234,6 +1234,11 @@ daemon_msg_auth_req(struct daemon_slpars *pars, uint32 plen)
 			uint32 usernamelen, passwdlen;
 
 			usernamelen = ntohs(auth.slen1);
+			if (usernamelen >= 0xFFFFFFFF) {
```
guyharris: `usernamelen` is set by a call to `ntohs()`, which puts a 16-bit quantity into host byte order from network byte order; as such, it should never produce a result > 65535.  How would it be possible for `usernamelen` to be > 65535?

guyharris:
> On a 32-bit system, this integer overflow will result in a buffer overflow a few lines later with `username[usernamelen] = '\0';`

How would a 16-bit quantity in the incoming packet cause an integer overflow on a *32-bit* system?  `usernamelen + 1` is at most 65536, which fits in a 32-bit integer.

catenacyber:
Sorry my mistake :-/

PR #799, opened 2019-02-05
DPDK: When using CMake PCAP_SUPPORT_DPDK is not set.

AndersBroman:


guyharris:
Those appear to be independent changes; the first is a change to CMakeFile.txt to fix it to properly set `PCAP_SUPPORT_DPDK` so that CMake builds work, the second is a change to the DPDK *code*, that is independent of whether it's built with autotools or CMake, causing it to return an error string from DPDK.

They should be submitted as separate pull requests.

AndersBroman:
That was the intention... Seems If I can figure out how to split it.


Den tis 5 feb. 2019 21:23 skrev Guy Harris <notifications@github.com>:

AndersBroman:
Done

PR #800, opened 2019-02-06
Fuzz rpcap

catenacyber:
The interesting target seems to be the server.
There is still performance optimization to be done.

This pull request also includes a small leak fix in pcap-linux.c (as well as the buffer overflows in #797)

### DIFF
--- a/rpcapd/daemon.c
+++ b/rpcapd/daemon.c
@@ -756,8 +756,11 @@ daemon_serviceloop(SOCKET sockctrl, int isactive, char *passiveClients,
 			tv.tv_usec = 0;
 
 			FD_SET(pars.sockctrl, &rfds);
-
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+			retval = 1;
+#else
 			retval = select((int)pars.sockctrl + 1, &rfds, NULL, NULL, &tv);
+#endif
 			if (retval == -1)
 			{
 				sock_geterror("select() failed", errmsgbuf, PCAP_ERRBUF_SIZE);
@@ -1105,6 +1108,9 @@ end:
 		session = NULL;
 	}
 
+	if (passiveClients) {
+		free(passiveClients);
+	}
 	//
 	// Finish using the SSL handle for the control socket, if we
 	// have an SSL connection, and close the control socket.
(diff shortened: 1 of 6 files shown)


guyharris:
There is still "making it build with CMake" work to be done. It is not building with CMake on Travis on Linux or macOS, and it is not building on AppVeyor on Windows (that is only done with CMake).

catenacyber:
@guyharris builds are now green, thanks :-)

The new fuzz targets get only built if `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` is defined in CFLAGS, because otherwise, we do not "fake" the sockets with a predefined buffer

rpcapd/daemon.c
```diff
@@ -2855,7 +2861,7 @@ void sleep_secs(int secs)
 {
 #ifdef _WIN32
```
guyharris, line 2862: It's probably cleaner to wrap the entire routine in

    #ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION

        ...

    #endif

sockutils.c
```diff
@@ -794,7 +819,9 @@ int sock_send(SOCKET sock, SSL *ssl _U_NOSSL_, const char *buffer, size_t size,
 		if (ssl) return ssl_send(ssl, buffer, remaining, errbuf, errbuflen);
 #endif
 
-#ifdef MSG_NOSIGNAL
```
guyharris, line 797: Same here.

sockutils.c
```diff
@@ -1004,7 +1031,9 @@ int sock_recv(SOCKET sock, SSL *ssl _U_NOSSL_, void *buffer, size_t size,
 	 * Win32.
 	 */
 	for (;;) {
-#ifdef HAVE_OPENSSL
+#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
+		nread = fuzz_recv(bufp, remaining);
+#elif defined(HAVE_OPENSSL)
 		if (ssl)
```
guyharris, line 1037: You can't remove `#ifdef HAVE_OPENSSL` - this has to build even without OpenSSL.

rpcapd/daemon.c
catenacyber: ok

sockutils.c
catenacyber: Not sure it is relevant here, but doing it

sockutils.c
catenacyber: It is not removed, it is moved 3 lines below.
fuzz socket now overrides both ssl and non-ssl sockets

PR #801, opened 2019-02-07
 DPDK: Don't return fatal ERROR(-) on find_all_devices()

AndersBroman:


pcap-dpdk.c
```diff
@@ -532,10 +534,18 @@ static int dpdk_pre_init(char * ebuf)
...
-	if (ret == 0){
-		is_dpdk_pre_inited = 1;
+	/* Copy the errno */
+	errno = rte_errno;
+
+	if (ret != 0){
+		pcap_fmt_errmsg_for_errno(ebuf, PCAP_ERRBUF_SIZE,
+			EACCES, "dpdk error: %s",
+			DPDK_ERR_RTE_EAL_INIT_FAIL);
+		errno = rte_errno;
```
agnosticdev: Is there a need to copy the `errno` here if it is being done above on line 538 too?

AndersBroman: Sorry didn't get that, Tried to improve the comment instead.

AndersBroman: I'm blind, see it now.

guyharris:
Most errors from `rte_eal_init()` *are* fatal; is the problem here that `ENOTSUP` shouldn't be treated as fatal?  If so, I've already done that in 8df8dadd730c74ec8cfc3f054a9558cb6c9357a3.  Or are there some *other* cases where `rte_eal_init()` can fail and that should be treated as "DPDK isn't available on this machine" rather than "DPDK is available but something's broken and needs to be fixed so that it *is* usable"?

AndersBroman:
Den tors 7 feb. 2019 20:04 skrev Guy Harris <notifications@github.com>:

> Most errors from rte_eal_init() *are* fatal; is the problem here that
> ENOTSUP shouldn't be treated as fatal? If so, I've already done that in
> 8df8dad
> <https://github.com/the-tcpdump-group/libpcap/commit/8df8dadd730c74ec8cfc3f054a9558cb6c9357a3>.
> Or are there some *other* cases where rte_eal_init() can fail and that
> should be treated as "DPDK isn't available on this machine" rather than
> "DPDK is available but something's broken and needs to be fixed so that it
> *is* usable"?
>
> —
>
Well my thinking here is that if I issue dumpcap -D I expect a list of the
available (Working) interfaces. Without the patch I get no interfaces
listed and an dpdk error.
Regards
Anders

AndersBroman:
Den tors 7 feb. 2019 21:33 skrev Anders Broman <a.broman58@gmail.com>:

>
>
> Den tors 7 feb. 2019 20:04 skrev Guy Harris <notifications@github.com>:
>
>> Most errors from rte_eal_init() *are* fatal; is the problem here that
>> ENOTSUP shouldn't be treated as fatal? If so, I've already done that in
>> 8df8dad
>> <https://github.com/the-tcpdump-group/libpcap/commit/8df8dadd730c74ec8cfc3f054a9558cb6c9357a3>.
>> Or are there some *other* cases where rte_eal_init() can fail and that
>> should be treated as "DPDK isn't available on this machine" rather than
>> "DPDK is available but something's broken and needs to be fixed so that it
>> *is* usable"?
>>
>> —
>>
> Well my thinking here is that if I issue dumpcap -D I expect a list of the
> available (Working) interfaces. Without the patch I get no interfaces
> listed and an dpdk error.
> Regards
> Anders
>
Note to self, read the whole post carefully before answering

I'll make new test with your patch. It looks like it fixes the problems.
Regards
Anders

AndersBroman:
Fixed differently in master

PR #803, opened 2019-02-15
Add DLT_/LINKTYPE_ values for IEEE802_15_4_TAP

jkcko:
Link type value to go with 
https://github.com/the-tcpdump-group/tcpdump/pull/734
and
https://bugs.wireshark.org/bugzilla/show_bug.cgi?id=15429

### DIFF
--- a/pcap-common.c
+++ b/pcap-common.c
@@ -1147,7 +1147,15 @@
 #define LINKTYPE_DSA_TAG_BRCM	281
 #define LINKTYPE_DSA_TAG_BRCM_PREPEND	282
 
-#define LINKTYPE_MATCHING_MAX	282		/* highest value in the "matching" range */
+/*
+ * IEEE 802.15.4 with pseudo-header and optional meta-data TLVs, PHY payload
+ * exactly as it appears in the spec (no padding, no nothing), and FCS if
+ * specified by FCS Type TLV;  requested by James Ko <jck@exegin.com>.
+ * Specification at https://github.com/jkcko/ieee802.15.4-tap
+ */
+#define LINKTYPE_IEEE802_15_4_TAP       283
+
+#define LINKTYPE_MATCHING_MAX	283		/* highest value in the "matching" range */
 
 /*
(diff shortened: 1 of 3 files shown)


pcap-common.c
```diff
@@ -1147,7 +1147,14 @@
 #define LINKTYPE_DSA_TAG_BRCM	281
 #define LINKTYPE_DSA_TAG_BRCM_PREPEND	282
 
-#define LINKTYPE_MATCHING_MAX	282		/* highest value in the "matching" range */
+/*
+ * IEEE 802.15.4 with pseudo-header and optional meta-data TLVs, PHY payload
+ * exactly as it appears in the spec (no padding, no nothing), and optional FCS
+ * at the end of the frame; requested by James Ko <jck@exegin.com>
+ */
+#define DLT_IEEE802_15_4_TAP    283
```
guyharris: That should be `LINKTYPE_IEEE802_15_4_TAP`.

LINKTYPE_ values were introduced because some early DLT_ types had different values on different flavors of BSD, and we wanted to have those link-layer types have the *same* value in capture files, even though the numerical values of the corresponding DLT_ might have to remain different on the different OSes for binary compatibility.

LINKTYPE_ types are defined by libpcap, and, although almost all of them have the same numerical value as the corresponding DLT_ type, there are a few exceptions, corresponding to the cases where the same link-layer type has different numerical values for the DLT_ type.  Our intent is to keep the values in sync for newly-assigned link-layer types, hence the "matching" range notion.

DLT_ values are returned by libpcap APIs; LINKTYPE_ values appear in pcap file headers and pcapng IDBs.  libpcap maps between them as necessary.

jkcko: sorry.  cut and paste error.  thanks for the great explanation though.  will fix when back in the office.

jkcko: Commit updated.
Fixed DLT_ -> LINKTYPE_ in pcap-common.c.
Added link to IEEE 802.15.4 TAP specification on github.

jkcko:
ping

Issue #805, opened 2019-03-08
Allow pcap_dump_fopen to work on Windows even when differing CRTs are used


MadAlexUK:
`pcap_dump_fopen()` is not safe on Windows because the caller may be using a C runtime that differs from that used by libpcap. There is an existing solution for `pcap_fopen_offline()` that is suitable: on Windows export a hopen variant instead that takes a raw OS `HANDLE` and then make the API a wrapper macro around that.

Pull request to follow.

guyharris:
On the one hand, making `pcap_dump_fopen()` a macro rather than a function potentially breaks binary compatibility with WinPcap and Npcap (and, if Npcap picks up the newer version of libpcap, with older versions of Npcap).

On the other hand, if you have a program that depends on WinPcap/Npcap exporting `pcap_dump_fopen()`, you have a program that has to be compiled with a compatible version of the CRT with which the library was built - and if, for example, WinPcap 4.1.3 and the current version of Npcap were built with incompatible CRTs (or, at least, incompatible at the C file-descriptor level), you have a program that depends on running with one of those two libraries, so binary compatibility might not exist *anyway*.

Plus, WinPcap exported `pcap_fopen_offline()` as a function until WinPcap 4..1, so they already broke binary compatibility there.

If some commercial product *really* depends on having those functions exported, they can just ship with the appropriate version of the library.

guyharris:
Next time, just submit the pull request - that way, when I merge it, I don't have to remember to close the issue.

PR #806, opened 2019-03-08
pcap_dump_fopen differing Windows CRTs work-around 

MadAlexUK:
#805
Make `pcap_dump_fopen()` a macro on Windows, wrapping the caller's CRT
functions for getting the raw OS file `HANDLE` and calling the exported
hopen variant. `pcap_dump_hopen()` creates a stream from that using
libpcap's CRT and passes that stream to the real `pcap_dump_fopen()`. This
mirrors what was done for `pcap_fopen_offline()`.

### DIFF
--- a/sf-pcap.c
+++ b/sf-pcap.c
@@ -110,6 +110,20 @@
 
 static int pcap_next_packet(pcap_t *p, struct pcap_pkthdr *hdr, u_char **datap);
 
+#ifdef _WIN32
+/*
+ * This isn't exported on Windows, because it would only work if both
+ * libpcap and the code using it were using the same C runtime; otherwise they
+ * would be using different definitions of a FILE structure.
+ *
+ * Instead we define this as a macro in pcap/pcap.h that wraps the hopen
+ * version that we do export, passing it a raw OS HANDLE, as defined by the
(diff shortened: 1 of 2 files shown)


sf-pcap.c
```diff
@@ -834,9 +848,41 @@ pcap_dump_open(pcap_t *p, const char *fname)
...
+
+	fd = _open_osfhandle(osfd, _O_APPEND);
+	if (fd < 0) {
+		pcap_fmt_errmsg_for_errno(p->errbuf, PCAP_ERRBUF_SIZE,
+		    errno, "_open_osfhandle");
+		return NULL;
+	}
+
+	file = _fdopen(fd, "wb");
+	if (file == NULL) {
```
gvanem: Shouldn't there be a `_close(fd);` here?

guyharris: Yes, there should.

And there should be one in `pcap_hopen_offline_with_tstamp_precision()` as well; I fixed that one in 40fe5f21bae1847a92f18efdcc8ad061db8f6cce.

MadAlexUK: Well spotted! Fixed here as well in 71cb091

PR #807, opened 2019-03-10
Haiku port

mmuman:
This has been getting dust in our ports tree for a while, so I cleaned it up and fixed against latest sources. I believe it's in shape to be upstreamed now.

I didn't include the updated configure (why is this versioned anyway?).

For reference, old patches and recipe [are still in HaikuPorts](https://github.com/haikuports/haikuports/tree/master/net-libs/libpcap).

poke @korli @pulkomandy

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -410,10 +410,18 @@ else(WIN32)
             #
             set(PCAP_LINK_LIBRARIES socket nsl ${PCAP_LINK_LIBRARIES})
         else(LIBSOCKET_HAS_GETADDRINFO)
-            #
-            # We didn't find it.
-            #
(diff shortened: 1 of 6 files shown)


guyharris:
Does CMake run on Haiku?

guyharris:
> Does CMake run on Haiku?

[I'm guessing the answer is "yes"](https://github.com/Kitware/CMake/blob/master/Modules/Platform/Haiku.cmake), so you might want to look at adding CMake support for building libpcap on Haiku as well.

mmuman:
Ah, yes, CMake works fine. I didn't notice it was used as well as, as previous patches didn't mention it.
Will have a look.

mmuman:
There you go.
Oh, I should be sleeping already :-)

PR #809, opened 2019-03-14
Minor Cleanups for pcap-haiku.cpp

agnosticdev:
Proposing a change to pcap-haiku.cpp to remove commented out code.
This PR does not contain any actual code changes, just cleanup.

### DIFF
--- a/pcap-haiku.cpp
+++ b/pcap-haiku.cpp
@@ -84,6 +84,14 @@ pcap_read_haiku(pcap_t* handle, int maxPackets, pcap_handler callback,
 	if (captureLength > handle->snapshot)
 		captureLength = handle->snapshot;
 
+	// run the packet filter
+	if (handle->fcode.bf_insns) {
+		if (pcap_filter(handle->fcode.bf_insns, buffer, bytesReceived,
+				captureLength) == 0) {
+			// packet got rejected
+			return 0;
+		}
+	}
 
 	// fill in pcap_header
 	pcap_pkthdr header;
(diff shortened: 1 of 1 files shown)


guyharris:
The code that runs `bpf_filter()` needs to be *un*commented and changed to use the new name, `pcap_filter()`; if you set a filter, it should filter out packets that don't match, on all platforms, whether the kernel can do the filtering or not.

agnosticdev:
Thank you for the feedback.  Just to confirm, add the BPF filtering code back:
```c
/*
	// run the packet filter
	if (!handle->md.use_bpf && handle->fcode.bf_insns) {
		if (bpf_filter(handle->fcode.bf_insns, buffer, bytesReceived,
				captureLength) == 0) {
			// packet got rejected
			return 0;
		}
	}
*/
```
And change `bpf_filter() to use the new `pcap_filter()` name? 
I am assuming that the `use_bpf` and `bf_insns` properties also have new names, but that is something I can research.

guyharris:
> Thank you for the feedback. Just to confirm, add the BPF filtering code back:

Yes.

> And change `bpf_filter() to use the new `pcap_filter()` name?

Yes.

> I am assuming that the`use_bpf`and`bf_insns` properties also have new names, but that is something I can research.

`bf_insns` has not changed its name.

`use_bpf` no longer exists, but you don't need it, as you're not having Haiku do the filtering.  Don't bother setting it, and don't bother testing it.  *If* Haiku supports attaching a filter to a socket, so that only packets for which a BPF program returns a non-zero value are provided, you'd have a flag in the private structure to indicate whether you need run the filter in libpcap or not.

You can also get rid of `pcap_setfilter_haiku()`, and just set the `setfilter_op` function pointer to point to `install_bpf_program()`.

agnosticdev:
Thank you for the feedback.  I will start making these changes shortly.

agnosticdev:
Changes made, please let me know if this is what you had in mind.

PR #814, opened 2019-04-15
Report the DLT description in error messages

fenner:
Introduce pcap_datalink_val_to_description_or_dlt, and
use that when reporting an error.  This was inspired by
seeing `tcpdump: no VLAN support for data link type 113`.
The new equivalent message is
`tcpdump: no VLAN support for Linux cooked`.

### DIFF
--- a/pcap.c
+++ b/pcap.c
@@ -3152,6 +3152,21 @@ pcap_datalink_val_to_description(int dlt)
 	return (NULL);
 }
 
+const char *
+pcap_datalink_val_to_description_or_dlt(int dlt)
+{
+        static char unkbuf[40];
+        const char *description;
+
+        description = pcap_datalink_val_to_description(dlt);
+        if (description != NULL) {
+                return description;
+        } else {
+                (void)pcap_snprintf(unkbuf, sizeof(unkbuf), "DLT %u", dlt);
+                return unkbuf;
+        }
+}
+
 struct tstamp_type_choice {
 	const char *name;
 	const char *description;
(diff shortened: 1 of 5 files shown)


pcap.c
```diff
@@ -3152,6 +3152,21 @@ pcap_datalink_val_to_description(int dlt)
...
+pcap_datalink_val_to_description_or_dlt(int dlt)
+{
+        static char unkbuf[40];
+        const char *description;
+
+        description = pcap_datalink_val_to_description(dlt);
+        if (description != NULL) {
+                return description;
+        } else {
+                snprintf(unkbuf, sizeof(unkbuf), "DLT %u", dlt);
```
guyharris: Use `pcap_snprintf()` instead; that works on more platforms.

fenner: Thanks!

fxlb: > tcpdump: no VLAN support for Linux cooked.

Not `Linux cooked v1` ?

guyharris: > > tcpdump: no VLAN support for Linux cooked.
> 
> Not `Linux cooked v1` ?

If you apply this change to a version prior to 8cff296dc7c321c76933359d586dbde5b580ce8c, it'll be "Linux cooked".  If you apply it to 8cff296dc7c321c76933359d586dbde5b580ce8c or later, it'll be "Linux cooked v1".

fenner: Sorry, my test environment is "libpcap 1.9.0 plus patches", not master.  Thanks for clarifying, Guy.  (I am looking forward to Linux cooked v2!)

PR #816, opened 2019-04-19
Detect Npcap's version.h in build tree and set HAVE_VERSION_H

dmiller-nmap:
Npcap needs this to be set so that we can report the Npcap version (e.g. 0.992) via `pcap_lib_version()`.

I apparently based this change on the release branch; let me know if you need me to cherry-pick it to a branch off of master and open a new PR instead.

PR #817, opened 2019-04-19
Updated: Detect Npcap's version.h in build tree and set HAVE_VERSION_H

dmiller-nmap:
New PR based on master, replacing #816.


guyharris:
Any idea why this ran build tests on Windows but not Linux/macOS?  It *shouldn't* affect UN\*X builds, but I'm not sure how it could have figure that out automatically.

PR #822, opened 2019-05-31
Remove redundant call to strlen

orgads:


guyharris:
Changed differently in 593769c3637583653bafa924fe348115016618c3.

PR #826, opened 2019-06-15
Fixes double free in opt_cleanup

catenacyber:
Found by oss-fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=15178

When `opt_init` fails after allocating some of `opt_state` fields, they get double freed : once in `opt_init` and once again in `opt_cleanup` after `opt_error` called `longjmp`

Proposed fix is :
- do not free null pointers (`opt_state` is memset to zero in `bpf_optimize`)
- set pointers to null after we free them in `opt_init`

### DIFF
--- a/optimize.c
+++ b/optimize.c
@@ -2051,18 +2051,12 @@ intern_blocks(opt_state_t *opt_state, struct icode *ic)
 static void
 opt_cleanup(opt_state_t *opt_state)
 {
-	if (opt_state->vnode_base)
-		free((void *)opt_state->vnode_base);
-	if (opt_state->vmap)
-		free((void *)opt_state->vmap);
-	if (opt_state->edges)
-		free((void *)opt_state->edges);
-	if (opt_state->space)
-		free((void *)opt_state->space);
-	if (opt_state->levels)
-		free((void *)opt_state->levels);
-	if (opt_state->blocks)
-		free((void *)opt_state->blocks);
+	free((void *)opt_state->vnode_base);
+	free((void *)opt_state->vmap);
+	free((void *)opt_state->edges);
+	free((void *)opt_state->space);
+	free((void *)opt_state->levels);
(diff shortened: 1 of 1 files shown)


optimize.c
```diff
@@ -2051,12 +2051,18 @@ intern_blocks(opt_state_t *opt_state, struct icode *ic)
 static void
 opt_cleanup(opt_state_t *opt_state)
 {
-	free((void *)opt_state->vnode_base);
-	free((void *)opt_state->vmap);
-	free((void *)opt_state->edges);
-	free((void *)opt_state->space);
-	free((void *)opt_state->levels);
-	free((void *)opt_state->blocks);
+	if (opt_state->vnode_base)
```
guyharris: C90 (and subsequent versions - and, I think, C89) explicitly require that, if a null pointer is passed to free(), nothing happens ("If `ptr` is a null pointer, no action occurs."), so the tests aren't a requirement, just a code style choice.  (I'm OK either way.)

optimize.c
```diff
@@ -2183,6 +2189,7 @@ opt_init(opt_state_t *opt_state, struct icode *ic)
 	opt_state->edges = (struct edge **)calloc(opt_state->n_edges, sizeof(*opt_state->edges));
 	if (opt_state->edges == NULL) {
 		free(opt_state->blocks);
+		opt_state->blocks = NULL;
```
guyharris: Would it be cleaner to just remove the `free()` calls from before `opt_error()` calls and let `opt_cleanup()` do the work?

catenacyber: Indeed, removing lines instead of adding them looks cleaner

optimize.c
catenacyber: I did not know it was standardized.
If this is only code style, this should go into another pull request...

PR #829, opened 2019-07-11
A new filter "ifindex" for LINUX_SLL2 and live Linux captures

fenner:
Create a filter for the ifIndex field in the LINUX_SLL2 pcap format, and convert it to the right SKF_AD_ value for live captures.

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -8175,6 +8175,53 @@ gen_multicast(compiler_state_t *cstate, int proto)
 	/*NOTREACHED*/
 }
 
+struct block *
+gen_ifindex(compiler_state_t *cstate, int ifindex)
+{
+	register struct block *b0;
+
+	/*
+	 * Catch errors reported by us and routines below us, and return NULL
+	 * on an error.
+	 */
+	if (setjmp(cstate->top_ctx))
+		return (NULL);
+
+	/*
+	 * Only some data link types support ifindex qualifiers.
+	 */
+	switch (cstate->linktype) {
+	case DLT_LINUX_SLL2:
+		/* match packets on this interface */
+		b0 = gen_cmp(cstate, OR_LINKHDR, 4, BPF_W, ifindex);
+		break;
+        default:
+#if defined(linux)
+		/*
+		 * This is Linux with PF_PACKET support.
+		 * If this is a *live* capture, we can look at
+		 * special meta-data in the filter expression;
+		 * if it's a savefile, we can't.
+		 */
+		if (cstate->bpf_pcap->rfile != NULL) {
+			/* We have a FILE *, so this is a savefile */
+			bpf_error(cstate, "ifindex not supported on %s when reading savefiles",
+			    pcap_datalink_val_to_description_or_dlt(cstate->linktype));
+			b0 = NULL;
+			/*NOTREACHED*/
+		}
+		/* match ifindex */
+		b0 = gen_cmp(cstate, OR_LINKHDR, SKF_AD_OFF + SKF_AD_IFINDEX, BPF_W,
+		             ifindex);
+#else /* defined(linux) */
+		bpf_error(cstate, "ifindex not supported on %s",
+		    pcap_datalink_val_to_description_or_dlt(cstate->linktype));
+		/*NOTREACHED*/
+#endif /* defined(linux) */
+	}
+	return (b0);
+}
+
 /*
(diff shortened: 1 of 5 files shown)


fxlb:
As a user, I’d rather be able to filter by interface name like in some BSD's but in and out.

Something like `ifname eth0`.

Way to have 'eth0' -> 'ifindex N' to build a filter?

See https://www.tcpdump.org/manpages/pcap-filter.7.html:

"ifname interface 
True if the packet was logged as coming from the specified interface (applies only to packets logged by OpenBSD's or FreeBSD's pf(4))."

fenner:
> As a user, I’d rather be able to filter by interface name like in some BSD's but in and out.

You can't do this on a savefile since the interface name is not stored in the savefile.  ifindex works in both live captures and savefile mode, even if it's less friendly.

Reworking the grammar so that "ifname" is not PF-specific is on my to-do list, but that'd be a different pull request.

mcr:
> > As a user, I’d rather be able to filter by interface name like in some BSD's but in and out.
> 
> You can't do this on a savefile since the interface name is not stored in the savefile. ifindex works in both live captures and savefile mode, even if it's less friendly.
> 
> Reworking the grammar so that "ifname" is not PF-specific is on my to-do list, but that'd be a different pull request.

And even if you were on the same machine, the ifindex can change after interfaces come/go, or after a reboot.    This is a case where pcapng would be better.

fxlb:
I suggest to rebase this PR on top of master.

fenner:
Looks like 8c263fd introduced problems with the FreeBSD build.

guyharris:
> Looks like [8c263fd](https://github.com/the-tcpdump-group/libpcap/commit/8c263fda3f8f274359ea47ecf667094443c704c9) introduced problems with the FreeBSD build.

Should be fixed by 00f29bec47f53e747c188f3b23aa7b0c4d45db6e.

gencode.c
```diff
@@ -8175,6 +8175,53 @@ gen_multicast(compiler_state_t *cstate, int proto)
...
+	if (setjmp(cstate->top_ctx))
+		return (NULL);
+
+	/*
+	 * Only some data link types support ifindex qualifiers.
+	 */
+	switch (cstate->linktype) {
+	case DLT_LINUX_SLL2:
+		/* match packets on this interface */
+		b0 = gen_cmp(cstate, OR_LINKHDR, 4, BPF_W, ifindex);
```
guyharris, line 8196: DLT_LINUX_SLL and DLT_LINUX_SLL2 headers are generated by libpcap; they're not visible to the in-kernel filter engine.

This probably needs to be done with something such as

    if (doing a live capture) {
        #ifdef linux
            test SKF_AD_OFF + SKF_AD_IFINDEX
        #else
            report "only supported on Linux"
        #endif
    } else {
        if (linktype is DLT_LINUX_SLL2)
            test the field at an offset of 4
        else
            report "not supported on savefiles except for DLT_LINUX_SLL2
    }

fenner: > DLT_LINUX_SLL and DLT_LINUX_SLL2 headers are generated by libpcap; they're not visible to the in-kernel filter engine.

Right, that's why `pcap-linux.c:fix_offset()` changes the filter on the fly when it's `DLT_LINUX_SLL*`.

> This probably needs to be done with something such as

I think my code is equivalent to what you suggest, but has the advantage of having strong parallels with the existing `gen_inbound()`.  Note that if we're doing a live capture, the code in `pcap-linux.c:fix_offset()` will turn the offset 4 comparison into an offset `SKF_AD_OFF + SKF_AD_IFINDEX` comparison.

guyharris: OK.

gencode.c
```diff
@@ -8175,6 +8175,53 @@ gen_multicast(compiler_state_t *cstate, int proto)
...
+		break;
+        default:
+#if defined(linux) && defined(PF_PACKET) && defined(SO_ATTACH_FILTER)
```
guyharris, line 8199: At this point, if we're building for Linux and neither `PF_PACKET` nor `SO_ATTACH_FILTER` are defined, libpcap won't build - we require a newer kernel - so it's sufficient to test whether `linux` is defined.

fenner: Sure, I first posted this way before bfbe856 . Thanks for pointing this out.

PR #836, opened 2019-08-21
Add support for uClibc < 0.9.31

RedDwarf69:
FWIW the systems where I found this to be a problem define ARPHRD_IEEE80211_PRISM and ARPHRD_IEEE80211_RADIOTAP, but only in `linux/if_arp.h`, not in `net/if_arp.h`.

This file being `pcap-linux.c` not sure whether to argue for changing the #include.


guyharris:
> This file being `pcap-linux.c` not sure whether to argue for changing the #include.

The comment for commit 3b3f7c1ea4ae96a9b5801789e9313ad932f12082 makes that argument, because that's what I did.

PR #839, opened 2019-08-31
Remove deprecated isascii function

neheb:
It is optionally unavailable with uClibc-ng without SUSv4 deprecated APIs.

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -2528,7 +2528,7 @@ add_linux_if(pcap_if_list_t *devlistp, const char *ifname, int fd, char *errbuf)
 	 */
 	p = ifname;
 	q = &name[0];
-	while (*p != '\0' && isascii(*p) && !isspace(*p)) {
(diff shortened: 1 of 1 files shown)


guyharris:
I presume "The isascii() function cannot be used portably in a localized application." in the SUSv4 means "localized applications need to deal with more than ASCII characters.

In our cases, we should deal with all non-ASCII characters by deeming them "not decimal digits", "not hexadecimal digits", "not alphabetical characters", "not upper-case alphabetical characters", "not alphanumeric characters", and "not space characters", so we don't get locale-dependent behavior when parsing Linux sysfs/devfs files, network interface names, and configuration files.  I've gotten rid of all the `<ctype.h>` stuff in 5a418352f3a1ca4f049396d7892d6e74f4011197 and ac0abe39ec6c5110e5a93b78fe80e990f420f05b.

neheb:
Great. Thanks for taking care of this.

PR #840, opened 2019-09-01
[WIP/RFC] Add initial implementation for pcap-openvizsla module

matwey:
Hi,

This module is for supporting OpenVizsla hardware USB sniffer. The implementation is based on [libopenvizsla](https://github.com/matwey/libopenvizsla) library.

DLT was assigned previously: https://github.com/the-tcpdump-group/libpcap/commit/a69f42f7873db778613bbd5c4238acc2d0c31c89


mcr:
Could you rebase?

guyharris:
Please add support to the autoconf script as well.

pcap-openvizsla.c
```diff
@@ -0,0 +1,199 @@
...
+
+	int ret = 0;
+
+	if ((ret = ov_capture_start(handlep->ov, handle->buffer, handle->bufsize, openvizsla_read_cb, &r)) < 0) {
+		snprintf(handle->errbuf, PCAP_ERRBUF_SIZE,
+			"Can't start capture: %s", ov_get_error_string(handlep->ov));
+		return PCAP_ERROR;
+	}
+
+	if ((ret = ov_capture_dispatch(handlep->ov, max_packets)) < 0) {
```
guyharris, line 54: Is there a way in which the callback routine can break out of the loop?  If, for example, a callback routine calls `pcap_breakloop()`, that should cause `pcap_loop()` or `pcap_breakloop()` not to process any more packets.

pcap-openvizsla.c
```diff
@@ -0,0 +1,199 @@
...
+	u_char *user;
+};
+
+static void openvizsla_read_cb(struct ov_packet* packet, void* data) {
+	struct pcap_openvizsla_read* r = (struct pcap_openvizsla_read*)data;
+
+	struct pcap_pkthdr pkth;
+	pkth.caplen = ov_to_host_16(packet->size) + sizeof(struct ov_packet);
+	pkth.len = pkth.caplen;
+	pkth.ts.tv_sec = 0;
```
guyharris, line 32: Can no time stamp be supplied?

matwey: The device provides 24-bit timestamp with precision of `1/60 usec`. Unfortunately, this value wraps around many times during the capture. What would be preferred way to provide best possible timestamp (combining 24-bit timestamp and the host wall clock)?

infrastation: So the 24-bit timestamp goes back to zero about 3.58 times per second, is that correct? (If yes, then just a combination of UNIX time and the 24-bit timestamp does not tell the exact time when a packet was received, especially at less then 3.58 packets per second, it needs to be something else.)

Is there any way to keep the 24-bit timestamp in sync with the host clock, such as sending a pulse to the device every second or a fraction of a second?

fxlb:
Could you rebase on top of the-tcpdump-group:master?

matwey:
Rebased on master

pcap-openvizsla.c
matwey: Correct. I think that the simplest way would be to use UNIX host receive timestamp and keep the counter as is.

pcap-openvizsla.c
matwey: `breakloop_op` implementation has been added.

matwey:
> Please add support to the autoconf script as well.

`autoconf` is supported now

PR #844, opened 2019-09-15
CMAKE_C_LIBRARY_ARCHITECTURE

bscottm:
Set CMAKE_C_LIBRARY_ARCHITECTURE in addition to CMAKE_LIBRARY_ARCHITECTURE
for more recent versions of CMake. CMAKE_C_LIBRARY_ARCHITECTURE is
apparently now preferred over CMAKE_LIBRARY_ARCHITECTURE.

### DIFF
--- a/cmake/Modules/FindPacket.cmake
+++ b/cmake/Modules/FindPacket.cmake
@@ -59,6 +59,7 @@ if(CMAKE_SIZEOF_VOID_P EQUAL 8)
   # without searching in the Lib directory first appears to be to set
   # CMAKE_LIBRARY_ARCHITECTURE to "x64".
   #
+  set(CMAKE_C_LIBRARY_ARCHITECTURE "x64")
   set(CMAKE_LIBRARY_ARCHITECTURE "x64")
 endif()
 
--- a/cmake/Modules/FindPacket.cmake
+++ b/cmake/Modules/FindPacket.cmake
@@ -58,6 +58,10 @@ if(CMAKE_SIZEOF_VOID_P EQUAL 8)
   # without searching in the Lib directory first appears to be to set
   # CMAKE_LIBRARY_ARCHITECTURE to "x64".
   #
+  # In newer versions of CMake, CMAKE_LIBRARY_ARCHITECTURE is set according to
+  # the language, e.g., CMAKE_<LANG>_LIBRARY_ARCHITECTURE. Consequently, it makes
+  # better sense to set CMAKE_C_LIBRARY_ARCHITECTURE as well.
+  #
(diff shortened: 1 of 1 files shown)


cmake/Modules/FindPacket.cmake
```diff
@@ -59,6 +59,7 @@ if(CMAKE_SIZEOF_VOID_P EQUAL 8)
   # without searching in the Lib directory first appears to be to set
   # CMAKE_LIBRARY_ARCHITECTURE to "x64".
```
guyharris: You should probably update the comment to indicate why `CMAKE_C_LIBRARY_ARCHITECTURE` is being set as well.

guyharris:
> Set CMAKE_C_LIBRARY_ARCHITECTURE in addition to CMAKE_LIBRARY_ARCHITECTURE
> for more recent versions of CMake. CMAKE_C_LIBRARY_ARCHITECTURE is
> apparently now preferred over CMAKE_LIBRARY_ARCHITECTURE.

Where is that documented?  I'm curious why it's better to say "look for versions of C libraries for this instruction set architecture" than "look for versions of *all* libraries for this instruction set architecture".  The goal, as indicated, is to for CMake to look in {WinPcap,Npcap}-SDK-to-level-directory/Lib/x64 before looking in {WinPcap,Npcap}-SDK-to-level-directory/Lib.  Are there versions of CMake where setting `CMAKE_LIBRARY_ARCHITECTURE` is not sufficient to do that, and you have to set `CMAKE_C_LIBRARY_ARCHITECTURE`?

bscottm:
From the CMake 3.16 (and was true in 3.15 as well) documentation, CMAKE_LIBRARY_ARCHITECTURE is a meta-variable (ref: [CMake 3.16 doc](https://cmake.org/cmake/help/latest/variable/CMAKE_LIBRARY_ARCHITECTURE.html)):

    CMAKE_LIBRARY_ARCHITECTURE

    Target architecture library directory name, if detected.

    This is the value of CMAKE_<LANG>_LIBRARY_ARCHITECTURE as detected for one of the enabled languages.

Thus, there is a good reason to set the library architecture for Win32 at the appropriate variable. I'll update the comment.

cmake/Modules/FindPacket.cmake
bscottm: Yup. Done.

fxlb:
You should remove the merge (41efdf7) and rebase your branch on master of the-tcpdump-group/libpcap.

bscottm:
Actually, it'll be probably be cleaner if I create a separate branch for a pull request. The merge should (is?) the result of updating to the master the-tcpdump-group/libpcap repo.

Issue #848, opened 2019-09-25
Valgrind reports uninitialized bytes used


ppaul1989:
Call to :

```
TEST(BasicUse, StartApi)
{
    char buffer[1024] = {0};
    pcap_open_live("nflog:30", 262144, 1, 1000, buffer);
}
```
 triggers valgrind error in netfilter_activate :

> 
> ==26589== Syscall param socketcall.sendto(msg) points to uninitialised byte(s)
> ==26589==    at 0x4875252: sendto (sendto.c:27)
> ==26589==    by 0x16EC26: netfilter_send_config_msg (in /XXX/XXX)
> ==26589==    by 0x16F731: netfilter_activate (in /XXX/XXX)
> ==26589==    by 0x164FA9: pcap_activate (in /XXX/XXX)
> ==26589==    by 0x165268: pcap_open_live (in /XXX/XXX)
> ==26589==    by 0x11A472: XXX::XXX::test::BasicUse_StartApi_Test::TestBody() (in /XXX/XXX)
> ==26589==    by 0x145A49: void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (in /XXX/XXX)
> ==26589==    by 0x13B1ED: testing::Test::Run() [clone .part.610] (in /XXX/XXX)
> ==26589==    by 0x13B6AC: testing::TestInfo::Run() [clone .part.611] (in /XXX/XXX)
> ==26589==    by 0x13B8CC: testing::TestCase::Run() [clone .part.612] (in /XXX/XXX)
> ==26589==    by 0x13C7A8: testing::internal::UnitTestImpl::RunAllTests() (in /XXX/XXX)
> ==26589==    by 0x13CAA7: testing::UnitTest::Run() (in /XXX/XXX)
> ==26589==  Address 0x1ffeffef89 is on thread 1's stack
> ==26589==  in frame #1, created by netfilter_send_config_msg (???:)
> ==26589==  Uninitialised value was created by a stack allocation
> ==26589==    at 0x16EB44: netfilter_send_config_msg (in /XXX/XXX)
> ==26589== 

Version used is git commit : 16970a4a176cd262743007bf53aea16321ec060b as submodule build with cmake options for linux:
```
set(BUILD_SHARED_LIBS OFF CACHE BOOL "" FORCE)
set(NO_PROTOCHAIN ON CACHE BOOL "" FORCE)
[...]

PR #849, opened 2019-09-25
#848 Fixed valgrind issue with uninitialized bytes

ppaul1989:
Related to issue #848. Initializes buff array with 0.


mcr:
Could you just squash these two commits together and git push -f ?
Thank you.

ppaul1989:
Squashed.

Issue #828, opened 2019-07-09
Some packets counted both as dropped and as read on Linux


Oppen:
`/proc/net/dev:rx_dropped` reflects **both** drops at L1/2 and drops at L3.
However, raw sockets see everything that make it to L2.
As a consequence, packets dropped because of congestion control or (in a case I stumbled upon on the wild) protocols unknown to the kernel, get counted both as received by the socket and as lost by ps_ifdrop.
There's a statistic more relevant, `rx_missed_errors`, accounting for packets the driver was unable to either read or pass on to the OS, but not all drivers implement it.
Luckily, we can use it through `SIOCETHTOOL` when supported, and fall back to `/proc/net/dev` otherwise.

I have half an implementation in my mind, but I wouldn't like to put the effort if it will be rejected in principle.
This doesn't mean I won't accept much needed criticism, but if the feature is not wanted at all it doesn't make much sense to implement it.

guyharris:
Getting the driver's drop statistics, if the driver keeps them, is the right thing to do.

I'm not sure falling back on getting a drop count that includes network-layer drops is worth the effort; better to just provide 0 for now (and if we later add an API that uses [pcapng ISB option codes](http://xml2rfc.tools.ietf.org/cgi-bin/xml2rfc.cgi?url=https://raw.githubusercontent.com/pcapng/pcapng/master/draft-tuexen-opsawg-pcapng.xml&modeAsFormat=html/ascii&type=ascii#rfc.section.4.6) as keys, so that a program can ask for a set of statistics and may only get back some of them, not provide it at all).

Is `rx_missed_errors` available from sysfs as well?  There's a comment in `pcap-linux.c` that says

     * Grabs the number of dropped packets by the interface from /proc/net/dev.
     *
     * XXX - what about /sys/class/net/{interface name}/rx_*?  There are
     * individual devices giving, in ASCII, various rx_ and tx_ statistics.
     *                  
     * Or can we get them in binary form from netlink?

Oppen:
The `ioctl` is to check for the feature.
You pass the command `ETHTOOL_GSSET_INFO` to get a set of strings, and `ETH_SS_STATS` as an argument to get the supported statistics.
If the string `"rx_missed_errors"` is present, then the statistic is available.
I think we should do this at context initialization time, and then read from `sysfs` if available.
The pseudo-file is always present regardless of support, so we can't rely on that for detection.
[...]

PR #854, opened 2019-10-04
Linux: get ifdrop stats from sysfs.

Oppen:
This change-set trades off possible over-accounting of packet loss by possible under-accounting it.
The previous way, reading the `rx_dropped` field from `/proc/net/dev` would always count not just what the interface loses (which is what we want), but what the link and upper layers decide to drop.
Instead, the sysfs stats are more accurate, and represent only one event each.
We care about those reported as missed and ones coming from FIFO errors, so we only gather that.
However, not all Linux drivers correctly fill these stats, and use `rx_dropped` as a one-size-fits-all stat.
For these cards, we lose the ability to account for interface-level losses.
Linux defaults that value to 0 in that case, so it follows what the manual specifies.

Due to lack of tools to simulate the loss, I tested this by stepping with a debugger to check the right values were read.

This addresses #828.

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -1166,108 +1166,39 @@ pcap_can_set_rfmon_linux(pcap_t *handle)
 }
 
 /*
- * Grabs the number of dropped packets by the interface from /proc/net/dev.
+ * Grabs the number of missed packets by the interface from
+ * /sys/class/net/statistics/{if_name}/rx_{missed,fifo}_errors.
  *
- * XXX - what about /sys/class/net/{interface name}/rx_*?  There are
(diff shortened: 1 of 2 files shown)


sla89:
Will this be in the 1.9 release too?

guyharris:
> Will this be in the 1.9 release too?

Given that there was a libpcap 1.9.0 release and there is currently a 1.9.1 release, there's no such thing as "the" 1.9 release - there are multiple 1.9 releases.  This change is in neither of them, as they were released before the change was submitted.

If there's a 1.9.2 release, the fix will probably be in that release.

PR #857, opened 2019-10-07
Linux: remove support for SOCK_PACKET.

Oppen:
Removes code supporting SOCK_PACKET, as well as any mentions I could find, and makes the code expect AF_PACKET to be present.
Later on I will be updating with whatever we agree to remove, and will add a simple version check that warns the user if their system is unsupported, as well as the last version supporting it.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -697,11 +697,6 @@ else
 		# No prizes for guessing this one.
 		#
 		V_PCAP=linux
-
-		#
-		# XXX - this won't work with older kernels that have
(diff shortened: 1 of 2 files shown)


guyharris:
Having heard no requests to leave SOCK_PACKET support in, away it goes.

guyharris:
And I removed more now-unnecessary stuff in f775748d268885709fe861bce90319fcc6244502 and 26de3578c11f386be25a6858abb7ed1528f03807.

PR #858, opened 2019-10-09
Ignore ENETDOWN

bamartin125:
It was the initial intent of @yuguy back in 2008 with this [commit](https://github.com/the-tcpdump-group/libpcap/commit/34624f128f2e6f937c04317fdd2c0ae6d8883ad5) to ignore `ENETDOWN` and allow libpcap to continue even if the interface goes down.

This functionality was later broken; probably by this [commit](https://github.com/the-tcpdump-group/libpcap/commit/e9de4b862ed18e08b9d1d8f04c804bed30a51431).

I'd like to propose that we attempt a fix with this small change.

The one thing I'm not sure of without really digging in is what the call structure looks like to get to each of `pcap_read_packet()` vs `pcap_wait_for_frames_mmap()`.

I would also like to know if I should restore the change that once existed from @yuguy 's commit in the `pcap_read_packet()` function by switching the do-while conditional back to include a test on `ENETDOWN` and effectively ignoring it there as well.

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -5149,6 +5149,7 @@ static int pcap_wait_for_frames_mmap(pcap_t *handle)
 					snprintf(handle->errbuf,
 						PCAP_ERRBUF_SIZE,
 						"The interface went down");
+					continue;
 				} else {
 					pcap_fmt_errmsg_for_errno(handle->errbuf,
 					    PCAP_ERRBUF_SIZE, errno,


fxlb:
There should be only one commit.
The body of the commit message should be folded to approximately 72 characters.

bamartin125:
Honestly, I'm not sure why there are two commits. I'll work on fixing that and the commit message.

guyharris:
What happens if, for example, you are capturing on a USB-connected interface, and you unplug the interface in the middle of the capture?  Continuing the loop won't work, because the interface is never coming back.

guyharris:
> It was the initial intent of @yuguy

Confusingly, that commit was made by me.  I have no idea why some of my commits were assigned to somebody else; if the commit was done before we moved to GitHub, perhaps GitHub tried to infer who made the commit, but failed to do so correctly.

guyharris:
> What happens if, for example, you are capturing on a USB-connected interface, and you unplug the interface in the middle of the capture?

The capture program keeps capturing.  (And plugging the interface back in doesn't make it start again, as the new interface has different data structures from the old one, and the socket doesn't get rebound to the new instance of the device with that name.)

It might be possible to distinguish between a temporary ENETDOWN from the interface just going down from a permanent ENETDOWN from the interface going away.  I'd been working on code to do that, but I had other projects to work on, so I'd have to page-fault back in the context to remember what more needed to be done.

crass:
Ok, so this isn't the ultimate solution.  But couldn't this be an interim solution?  This is a real problem that it sounds like this PR solves for a lot of cases. So I see a big benefit to the change (I suspect this mostly happens as a result of re-initializing the interface, not removing the device).  What's the downside? Someone is capturing on a USB device, unplugs the device, and tcpdump does not terminate.  The user may not know that after plugging the device back in that the capture will not continue on the new device, but they will certainly find out.  This behavior could be documented until the ultimate fix happens.

So can we get this merged in to provide some needed functionality now?

guyharris:
> I think there's a way to distinguish between "interface went down but is still present" from "interface disappeared", but it's not simple

There is; see a973128a85d7dd75c7ea6fdcf746fc143a987d03.

PR #862, opened 2019-10-14
Linux: no longer report 'without TPACKET'.

Oppen:


guyharris:
That should be done as part of a change that removes all checks for `HAVE_PACKET_RING` and `PCAP_SUPPORT_PACKET_RING`, as well as the tests in autoconf and CMake that set `PCAP_SUPPORT_PACKET_RING`.

Oppen:
OK, that makes sense. I thought it was the only remaining omission.
I'll rebase on this same PR if that's OK.

guyharris:
> OK, that makes sense. I thought it was the only remaining omission.

No.  We now have no support for `SOCK_PACKET`, and thus require at least 2.2, but we still don't require turbopacket (which I think means requiring at least 2.4).

> I'll rebase on this same PR if that's OK.

OK.

guyharris:
Needs to be rebased against the tip of the master branch, to resolve the conflicts.

PR #866, opened 2019-10-16
Fill readme file for Win32 platform

JeromeMigne:
Provide details about rpcapd compilation inside Visual Studio IDE.

These details were missing to compile rpcapd over Npcap for Windows platform.

### DIFF
--- a/doc/README.Win32.md
+++ b/doc/README.Win32.md
@@ -1,3 +1,58 @@
-Win32 used to build with Visual Studio 6, but we now use cmake.
+# Introduction
 
-This file needs to be adopted by a windows expert developer.
+To configure and to build this project entirely in Visual Studio IDE,
+Visual Studio 2017 or later is needed.
+
+Other methods of configuring and building can be adapted from this documentation.
+
+Use the `Open Folder` Visual Studio command to open this CMake project.
+
+
+# Building rpcapd over Npcap
+
+## Prerequisites
+
+The following external dependencies are needed for compilation:
+
+  - [Win flex-bison](https://sourceforge.net/projects/winflexbison/)
+  - [Npcap DLLs](https://nmap.org/npcap/) click on "Npcap _version_ installer" to download
+       and during installation uncheck all options, should not be installed in WinPcap API-compatible Mode.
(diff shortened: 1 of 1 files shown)


doc/README.Win32.md
```diff
@@ -1,3 +1,55 @@
-Win32 used to build with Visual Studio 6, but we now use cmake.
+# Introduction
 
-This file needs to be adopted by a windows expert developer.
+A recent version of Visual Studio is needed to handle CMake projects.
```
guyharris: By this are you referring to [the built-in CMake support introduced in Visual Studio 2017](https://devblogs.microsoft.com/cppblog/cmake-support-in-visual-studio/)?

If so, please note that:

- you should say "Visual Studio 2017 or later", not just "a recent version";
- you don't necessarily need VS 2017 or later to build, but if you have an earlier version, you'll have to run CMake yourself to generate the .prj and .sln files for VS.

Note also that there is no need to run the Visual Studio IDE to build libpcap (or tcpdump or Wireshark or...) - you can just use msbuild.

JeromeMigne: Updated.

JeromeMigne:
When opening the folder with Visual Studio 2019, without any modification:
 - the following CMake error occurs: `Neither flex nor win_flex nor lex was found`,
 - `ENABLE_REMOTE` is set by default to `YES`,
 - `PACKET_TYPE` is not set and after having resolved all other settings, it compiles with `pcap-null`.

guyharris:
I've updated the document to give details about downloading and installing required components and about configuring and building with the VS 2017 GUI, the VS 2019 GUI, and the command line.  I tested the instructions with VS 2017 on Windows 7 and VS 2019 on Windows 10.

daluu:
Thanks for the README update. I would say however, that some changes in this PR that did not make it to the actual README, are more helpful to a newbie to cmake + Visual Studio config, like the cmake variables that point to lex & yacc. It's more clear from this PR about it. And from comparing the two, I guess setting `PCAP_TYPE` to `npf` is unnecessary?

PR #881, opened 2019-11-10
Linux: require PACKET_RING.

Oppen:


pcap-linux.c
```diff
@@ -3530,16 +3462,12 @@ activate_sock(pcap_t *handle, int is_any_device)
 	return status;
 }
 
-#ifdef HAVE_PACKET_RING
 /*
- * Attempt to activate with memory-mapped access.
+ * Activate with memory-mapped access.
```
guyharris: The attempt may fail, which is why I said "Attempt to activate".

pcap-linux.c
```diff
@@ -1456,90 +1441,37 @@ pcap_activate_linux(pcap_t *handle)
...
 	}
 
 	/*
-	 * "handle->fd" is a socket, so "select()" and "poll()"
-	 * should work on it.
+	 * We failed to set up to use it.
+	 * status has been set to the error status to return and, if
+	 * it's PCAP_ERROR, handle->errbuf contains the error message.
 	 */
-	handle->selectable_fd = handle->fd;
```
guyharris: So where does this now set `handle->selectable_fd`?  It has to get set, otherwise applications won't be able to use `select()`/`poll()`/etc. on the socket.  (No, you're not supposed to use `pcap_fileno()` to get the selectable descriptor.)

pcap-linux.c
Oppen: Noted. Sorry.
I was planning to rebase to fix any issues, how should I handle your commit?
Should I keep the mistake in the original commit or just squash the fix with it?

Also, I had no idea you could commit to the PR, I thought that came only from my fork.

pcap-linux.c
Oppen: That was unintented. I'll fix that and take a closer look just in case. I'll try to be more careful next time. Thanks for your patience.

pcap-linux.c
guyharris: Whichever is more convenient; if they're equally committed, squash.

Oppen:
I'm closing this one, it'll be easier to just redo than to fix the conflicts.

PR #882, opened 2019-11-12
Provide APIs to map between DLT_ values and LINKTYPE_ values

rnbdsh:
So it's useable in the .so, not just the .a version of the lib.
This is relevant as the moloch project is using this function in https://github.com/aol/moloch/blob/master/capture/reader-libpcap.c#L94
The full issue is described at https://github.com/aol/moloch/issues/1303

### DIFF
--- a/pcap/pcap.h
+++ b/pcap/pcap.h
@@ -477,6 +477,8 @@ PCAP_API int	pcap_is_swapped(pcap_t *);
 PCAP_API int	pcap_major_version(pcap_t *);
 PCAP_API int	pcap_minor_version(pcap_t *);
 PCAP_API int	pcap_bufsize(pcap_t *);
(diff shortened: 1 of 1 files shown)


guyharris:
1. As pcap APIs, published versions would need to have names beginning with `pcap_`.

1. They would also need to have man pages.

1. Including `pcap/pcap.h`would need to cause all the LINKTYPE_ values returned by `pcap_dlt_to_linktype()`, and taken as arguments by `pcap_linktype_to_dlt()`, to be defined.  Putting them into a `pcap/linktype.h` header, included by `pcap/pcap.h` just as it includes `pcap/dlt.h`, would be appropriate here.

guyharris:
Neither of these are necessary for Moloch, as per my comments on aol/moloch#1303, so I'm closing this.

Issue #843, opened 2019-09-11
Large buffer size leads to packet drops


sla89:
Hello together,

I am capturing packets using libpcap 1.9.0 on linux (kernel 4.19.50-rt22) and I am facing the problem that I have packet drops when I am using buffer sizes >=4MiB.The larger the value the more packets were dropped. The snapshot length was set to 1538 bytes.

With a buffer size below the mentioned value I have no packet drops at all. 

Can you please explain to me why larger buffer sizes lead to drops? In my understanding a larger buffer size should avoid packet drops.

I have the same behavior when I am running tcpdump with the mentioned values on that platform.

It would be great to get an explanation for this behavior.

Thank you very much and with best regards,
Stefan

Oppen:
Maybe buffers this big thrash the TLBs? Just a guess. If it's this, using huge pages should help, but I'm not sure if libpcap provides that optino.
Could you test a dummy read-and-discard on both scenarios?
I think that could give us a lead on whether libpcap is doing something wrong or the application code starts performing worse (thus queuing more packets) on bigger buffers.

sla89:
What do you mean with dummy read-and-discard?

Oppen:
Doing something like:
```
void dummy_handler(u_char *user, const struct pcap_pkthdr *h, const u_char *bytes) {
  return;
}
int main(void) {
  // Initialization stuff...
pcap_loop(pcap_inst, -1, dummy_handler, NULL);
  return 0;
```

Oppen:
Essentially, do no processing, to check whether your processing became slower or libpcap is misbehaving.

guyharris:
> Maybe buffers this big thrash the TLBs? Just a guess. If it's this, using huge pages should help, but I'm not sure if libpcap provides that optino.

It doesn't, and I'm not sure it should; instead, it should, if possible, figure out by itself whether huge pages can be used and, if so, request the appropriate size.

Assuming that the kernel will allow the packet buffer of a PF_PACKET socket to be mapped with huge pages, is there any reason *not* to just have libpcap decide whether to use them?

Oppen:
> It doesn't, and I'm not sure it should; instead, it should, if possible, figure out by itself whether huge pages can be used and, if so, request the appropriate size.
Sounds reasonable.
> Assuming that the kernel will allow the packet buffer of a PF_PACKET socket to be mapped with huge pages, is there any reason _not_ to just have libpcap decide whether to use them?
Sounds doable.
[...]

PR #887, opened 2019-11-19
Set socket protocol only after packet ring configured

mm120:
When using the memory mapped packet receive path on Linux, we find
that if the buffer is "large" we get a number of dropped packets.
This number of packets is roughly proportional to the buffer size.
It turns out that these packets are dropped by the kernel in the time
period between the socket being opened and the ring parameter setup
completing (this is where the kernel has to allocate the buffer memory).

To prevent this, we can open the socket on protocol 0, which the kernel
interprets to mean that we don't want to receive any packets at all (this
is described when opening a memory mapped packet socket for TX only).  We
then set the correct protocol once everything is configured.

Fixes issue #843 
Signed-off-by: Mark Marshall <mark.marshall@omicronenergy.com>

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -1323,17 +1323,21 @@ static void pcap_breakloop_linux(pcap_t *handle)
 static int
 open_pf_packet_socket(pcap_t *handle, int cooked)
 {
-	int	protocol = pcap_protocol(handle);
 	int	sock_fd, ret;
 
 	/*
 	 * Open a socket with protocol family packet. If cooked is true,
 	 * we open a SOCK_DGRAM socket for the cooked interface, otherwise
 	 * we open a SOCK_RAW socket for the raw interface.
+	 *
+	 * The protocol is set to 0.  This means we will receive no
+	 * packets until we "bind" the socket with a non-zero
+	 * protocol.  This allows us to setup the ring buffers without
+	 * dropping any packets.
 	 */
 	sock_fd = cooked ?
-		socket(PF_PACKET, SOCK_DGRAM, protocol) :
-		socket(PF_PACKET, SOCK_RAW, protocol);
+		socket(PF_PACKET, SOCK_DGRAM, 0) :
+		socket(PF_PACKET, SOCK_RAW, 0);
 
 	if (sock_fd == -1) {
 		if (errno == EPERM || errno == EACCES) {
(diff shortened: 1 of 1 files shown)


pcap-linux.c
```diff
@@ -1469,42 +1474,60 @@ pcap_activate_linux(pcap_t *handle)
...
+		 * Kernel doesn't support it (mmap) - just continue
+		 * with non-memory-mapped access.
+		 */
 #endif /* HAVE_PACKET_RING */
 	}
 
+	/*
+	 * We need to set the correct protocol.
+	 */
+	if ((status2 = iface_bind(handle->fd, handlep->ifindex,
```
guyharris: If the "any" device is opened, is `handlep->ifindex` 0?  And does that result in the socket being bound to a protocol but not to an interface (as I suspect it will, from a quick look at af_packet.c), so that packets from all interfaces show up (e.g., your main interface and the loopback interface)?

pcap-linux.c
```diff
@@ -1469,42 +1474,60 @@ pcap_activate_linux(pcap_t *handle)
...
 			 */
-			set_poll_timeout(handlep);
-			return status;
-
-		case 0:
 			/*
-			 * Kernel doesn't support it - just continue
-			 * with non-memory-mapped access.
+			 * Now that we have activated the mmap ring, we can
+			 * set the correct protocol.
```
guyharris: "...we can bind to the specified interface, if one was specified, and set the correct protocol" - that call does both.

pcap-linux.c
mm120: My understanding is that ifindex would be -1, and I map that back to 0 in iface_bind.  The man page for "packet" says about ifindex "0 matches any interface"

pcap-linux.c
mm120: Yes, you are right.

The problem that I had, is that I tried to make my match as minimal as possible, and not to change the overall structure of the code too much.  This means that iface_bind often gets called twice, once with (ifindex, 0) and then again with (ifindex, protocol).

I assumed that binding twice would be best - if the ifindex is bad, or the device has errors, we get those on the first bind call, and the code flow is unchanged by my patch.  The second bind call is less likely to fail, as "less is changing" (I felt, at least).

pcap-linux.c
guyharris: Seems to work in my tests.

PR #895, opened 2019-12-25
Probe CONFIGURATION descriptor of connected USB devices

desowin:
The DEVICE descriptor alone is not enough for full fledged USB class
(eg. Communications, Audio) dissection. Read CONFIGURATION descriptor
twice. First time just with the mandatory 9 bytes, and then using the
wTotalLength to get full CONFIGURATION. Do not attempt requesting full
65535 bytes as many devices do not return the descriptor in such case.

While the DEVICE and CONFIGURATION descriptors are not allowing full
dissection for all cases (eg. HID devices need also report descriptors)
knowing the CONFIGURATION descriptor makes the dissection possible in
many cases.

### DIFF
--- a/pcap-usb-linux.c
+++ b/pcap-usb-linux.c
@@ -489,8 +489,10 @@ int usb_mmap(pcap_t* handle)
 #define USB_REQ_GET_DESCRIPTOR	6
 
 #define USB_DT_DEVICE		1
+#define USB_DT_CONFIG		2
 
 #define USB_DEVICE_DESCRIPTOR_SIZE	18
+#define USB_CONFIG_DESCRIPTOR_SIZE	9
 
 /* probe the descriptors of the devices attached to the bus */
 /* the descriptors will end up in the captured packet stream */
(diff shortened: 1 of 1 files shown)


desowin:
This is inspired by https://bugs.wireshark.org/bugzilla/show_bug.cgi?id=9205

Also, I believe many people might be unfamiliar with what Wireshark can do regarding USB dissection when it has full information about connected device.

desowin:
AppVeyor failed due to issue unrelated to this change:
```
Error retrieving packages from source 'https://chocolatey.org/api/v2/':
 Could not connect to the feed specified at 'https://chocolatey.org/api/v2/'. Please verify that the package source (located in the Package Manager Settings) is valid and ensure your network connectivity.
winflexbison not installed. The package was not found with the source(s) listed.
 Source(s): 'https://chocolatey.org/api/v2/'
 NOTE: When you specify explicit sources, it overrides default sources.
If the package version is a prerelease and you didn't specify `--pre`,
 the package may not be found.
Please see https://chocolatey.org/docs/troubleshooting for more 
 assistance.
Chocolatey installed 0/1 packages. 1 packages failed.
 See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).
```

guyharris:
I got rid of the common buffer for pathnames and USB device descriptors, splitting them into two separate buffers for cleanliness, so you'll need to merge and rebase.

PR #897, opened 2020-01-06
Fix compiler warnings on MSYS2/Cygwin

orgads:
```
/f/Projects/libpcap/fmtutils.c:114:20: warning: initialization of ‘char *’ from ‘int’ makes pointer from integer without a cast [-Wint-conversion]
  114 |  char *errstring = strerror_r(errnum, strerror_buf, PCAP_ERRBUF_SIZE);
      |                    ^~~~~~~~~~
```

CMake executes the test for strerror_r with GNU_SOURCE defined,
but the actual application did not define it.

### DIFF
--- a/ftmacros.h
+++ b/ftmacros.h
@@ -83,7 +83,7 @@
    * least with HP's C compiler; hopefully doing so won't make it
    * *not* work with *un*-threaded code.
    */
-#elif defined(__linux__) || defined(linux) || defined(__linux)
+#elif defined(__linux__) || defined(linux) || defined(__linux) || defined(__CYGWIN__)
   /*
    * Turn on _GNU_SOURCE to get everything GNU libc has to offer,
    * including asprintf().


guyharris:
Do you mean "MSYS" or "MinGW"?  [The MinGW page for MSYS](http://www.mingw.org/wiki/MSYS) says

> MSYS is a collection of GNU utilities such as bash, make, gawk and grep to allow building of applications and programs which depend on traditionally UNIX tools to be present. It is intended to supplement MinGW and the deficiencies of the cmd shell.

so it sounds as if you're building with MinGW + MSYS and that the compiler and library being used here is the MinGW one, not the MSYS one.

Testing for `__GNUC__` is incorrect - that tests for whether the compiler is GCC or a compiler claiming compatibility with GCC, but, just because the compilation is being done with such a compiler, that doesn't mean that the C library is GNU libc, and therefore doesn't mean we should define `GNU_SOURCE`.

orgads:
That's MSYS, not MinGW. I'll replace the test with __MSYS__ then?

guyharris:
> That's MSYS, not MinGW. I'll replace the test with `__MSYS__` then?

What do you mean by "That's MSYS, not MinGW"?  Do you mean you've installed the MSYS commands, but haven't installed the MinGW compiler, and are using some other compiler?

orgads:
MSYS2 has 3 different compilers, each of them has its own "environment". The "native" environment is msys. This env is quite similar to cygwin. It requires msys runtime dlls for executing the product.

The 2 other envs are mingw32 and mingw64. Products that are built in these envs link against the system's C runtime, which is msvcrt.dll, and don't require any msys runtime.

So I referred to the first one.

orgads:
Changed the define to `__MSYS__`.

On MinGW strerror_r doesn't exist at all...

guyharris:
> MSYS2 has 3 different compilers, each of them has its own "environment".

So what you meant to say is "That's [MSYS2](http://www.msys2.org), not MSYS":

> MSYS2 is a software distro and building platform for Windows
At its core is an independent rewrite of MSYS, based on modern Cygwin (POSIX compatibility layer) and MinGW-w64 with the aim of better interoperability with native Windows software. It provides a bash shell, Autotools, revision control systems and the like for building native Windows applications using MinGW-w64 toolchains.

which is different from MSYS.  MSYS2 appears to be a Cygwin alternative, not just a set of UN*X tools ported to Windows, as MSYS is.

So, according to [the MSYS2 wiki's Porting page](https://github.com/msys2/msys2/wiki/Porting), `__CYGWIN__` is defined both if you're building for Cygwin and if you're building for MSYS2, and `__MSYS__` is defined only if you're building for MSYS2.  Does `GNU_SOURCE` need to be defined if you're building for Cygwin?

orgads:
Yes, I thought that was obvious. MSYS1 has been dead for ages.

Regarding cygwin, I didn't test, but I suppose it also defines `__linux__`, so there's no reason to add it explicitly.

orgads:
And the commit message says "MSYS2" :)

guyharris:
> Yes, I thought that was obvious. MSYS1 has been dead for ages.
> 
> Regarding cygwin, I didn't test, but I suppose it also defines `__linux__`

It's not Linux, it's a library for Windows that provides an implementation of a UN\*X-compatible API; the particular UN\*X that it most closely resembles might be Linux+Glibc, but there's stuff that's probably not 100% Linux-compatible, so it might not define `__linux__`.

[The Cygwin FAQ](https://www.cygwin.com/faq.html) doesn't mention `__linux__`.

orgads:
You were right. The warning also appears on Cygwin. Changed the macro to `__CYGWIN__`.

PR #900, opened 2020-02-09
Linux: require turbo packet.

Oppen:


configure.ac
```diff
@@ -1158,11 +1158,6 @@ AC_ARG_ENABLE([packet-ring],
 [AC_HELP_STRING([--enable-packet-ring],[enable packet ring support on Linux @<:@default=yes@:>@])],
```
guyharris: You might as well remove the entire --enable-packet-ring option, not just the test for whether it was enabled or not.

Oppen: Done.

PR #901, opened 2020-02-09
Linux: refactor broken TPACKET_V3 detection.

Oppen:
Extracted to a separate function and using sscanf for version parsing to make it easier to follow.

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -1227,6 +1227,44 @@ static void	pcap_cleanup_linux( pcap_t *handle )
 	pcap_cleanup_live_common(handle);
 }
 
+#ifdef HAVE_TPACKET3
+/*
+ * Some versions of TPACKET_V3 have annoying bugs/misfeatures
+ * around which we have to work.  Determine if we have those
+ * problems or not.
+ * 3.19 is the first release with a fixed version of
+ * TPACKET_V3.  We treat anything before that as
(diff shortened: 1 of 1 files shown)


guyharris:
With which release strings does it fail/

Oppen:
@guyharris I'm not sure why I can't respond to your question inline.
The FIXME was intended for me really, to remind me to check.
Where I think it may fail is for RC kernels, I'm building one now to check this. I think it may have the form "5.3-rc" just as the tar containing it, but I need to confirm that. If that's the case, we would reject that as a valid version because the minor is followed by a '-' instead of either '\0' or '.'.
The patch reproduces the original behavior from the refactored code.

Oppen:
OK, I checked and the default string seems to be of the form "x.yy.zz-rcn", so I'll remove the comment.

mcr:
Why are there any failures/changes on FreeBSD?  It shouldn't have changed. Hmm.

guyharris:
> Why are there any failures/changes on FreeBSD? It shouldn't have changed. Hmm.

Because Cirrus malfunctioned somehow.

I restarted the tests, and it seems to have worked.  (The "restart the tests" UI is impressively bad, at least with Safari - it pops up a window that runs forever, never saying "the tests finished", and I think if you close the window before it's done, I think the tests are stopped.  You have to go to the page in another window to see the status.)

PR #904, opened 2020-02-17
Linux: require TPACKET v2.

Oppen:
This also removes some leftovers from non-mmapped mode.

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -111,11 +111,6 @@
 #ifdef TPACKET3_HDRLEN
 # define HAVE_TPACKET3
 #endif /* TPACKET3_HDRLEN */
-#ifdef TPACKET2_HDRLEN
-# define HAVE_TPACKET2
-#else  /* TPACKET2_HDRLEN */
-# define TPACKET_V1	0    /* Old kernel with only V1, so no TPACKET_Vn defined */
-#endif /* TPACKET2_HDRLEN */
 
 #ifdef SO_ATTACH_FILTER
 #include <linux/types.h>
@@ -187,7 +182,6 @@ typedef int		socklen_t;
  * Private data for capturing on Linux PF_PACKET sockets.
  */
 struct pcap_linux {
-	u_int	packets_read;	/* count of packets read with recvfrom() */
 	long long sysfs_dropped; /* packets reported dropped by /sys/class/net/{if_name}/statistics/rx_{missed,fifo}_errors */
 	struct pcap_stat stat;
 
@@ -234,8 +228,6 @@ static int pcap_activate_linux(pcap_t *);
 static int activate_pf_packet(pcap_t *, int);
 static int setup_mmapped(pcap_t *, int *);
 static int pcap_can_set_rfmon_linux(pcap_t *);
-static int pcap_read_linux(pcap_t *, int, pcap_handler, u_char *);
-static int pcap_read_packet(pcap_t *, pcap_handler, u_char *);
 static int pcap_inject_linux(pcap_t *, const void *, int);
 static int pcap_stats_linux(pcap_t *, struct pcap_stat *);
 static int pcap_setfilter_linux(pcap_t *, struct bpf_program *);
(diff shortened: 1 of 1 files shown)


pcap-linux.c
```diff
@@ -3677,54 +3246,20 @@ init_tpacket(pcap_t *handle, int version, const char *version_str)
...
 /*
  * Attempt to set the socket to version 3 of the memory-mapped header and,
  * if that fails because version 3 isn't supported, attempt to fall
- * back to version 2.  If version 2 isn't supported, just leave it at
- * version 1.
+ * back to version 2.  If version 2 isn't supported, just fail.
  *
- * Return 1 if we succeed or if we fail because neither version 2 nor 3 is
- * supported; return -1 on any other error, and set handle->errbuf.
+ * Return 1 if we succeed or if we fail because is not supported;
```
guyharris: This only returns 1 if it succeeds, at least as I read it; -1 is returned if we fail for any reason.

Also, shouldn't `init_tpacket()` return 0 on success and -1 for any failure?  ENOPROTOOPT from `getsockopt(..., SOL_PACKET, PACKET_HDRLEN, ...)` means "we don't support memory-mapped capture at all, and EINVAL means "we don't support anything other than TPACKET_V1, so `init_tpacket()` should return -1 with an error message saying "your kernel is too old to support TPACKET_V2 memory-mapped capture" (I'm guessing that, by the time TPACKET_V2 was added, you can't un-configure PACKET_MMAP - if not, perhaps the error should distinguish between "kernel is too old" and "kernel isn't too old, but you have to configure PACKET_MMAP).

Given that, `prepare_tpacket_socket()` should treat -1 as an error and anything else as success.

Oppen:
Don't merge, I realized there are still some pieces to be removed, regarding statistics (TPACKET_V2 implies socket stats).
Review is still welcome.

pcap-linux.c
Oppen: 1. Fixed on my local repo.
2. Yes, you're correct, it's used as if it returns 0 on success and -1 on error.
3. I'm not sure about the PACKET_MMAP configuration option, but on kernel version I though of throwing an error on initialization on unsupported kernels instead.
4. Makes sense, I'll take a look.

On next review, could you take special attention in the statistics changes? I think those are harder to properly test, so the extra caution will prove useful.

Oppen: OK, I *think* I addressed all of these.

Oppen:
OK, it can be merged (unless someone point out a defect, of course).

pcap-linux.c
```diff
@@ -2149,51 +1693,7 @@ pcap_stats_linux(pcap_t *handle, struct pcap_stat *stats)
...
-	 * "handlep->packets_read", for reasons described in the comment
-	 * at the end of pcap_read_packet().  We have no idea how many
-	 * packets were dropped by the kernel buffers -- but we know
-	 * how many the interface dropped, so we can return that.
-	 */
-
-	stats->ps_recv = handlep->packets_read;
-	stats->ps_drop = 0;
-	stats->ps_ifdrop = handlep->stat.ps_ifdrop;
+	// MARIO: should indicate failure? If so, is it by returning -1?
```
guyharris: > should indicate failure?

Yes.

> If so, is it by returning -1?

Yes.  Also, set `handle->errbuf` to an appropriate error string, by calling `pcap_fmt_errmsg_for_errno()`.

pcap-linux.c
```diff
@@ -3653,9 +3143,13 @@ init_tpacket(pcap_t *handle, int version, const char *version_str)
 			 * ENOPROTOOPT means the kernel is too old to
 			 * support PACKET_HDRLEN at all, which means
 			 * it either doesn't support TPACKET at all
-			 * or supports  only TPACKET_V1.
+			 * or supports only TPACKET_V1.
 			 */
-			return 1;	/* no */
+			pcap_fmt_errmsg_for_errno(handle->errbuf, PCAP_ERRBUF_SIZE, errno,
```
guyharris: I forgot that this is called with both TPACKET_V3 and TPACKET_V2, to try to get the best version of TPACKET, so it *does* need to have three different return values.

I think EINVAL means "we support TPACKET, we just don't support *that particular version of tpacket*", so that should cause it to return 1, so, if it was called with `TPACKET_V3`, `prepare_tpacket_socket()` can fall back on `TPACKET_V2`.

pcap-linux.c
```diff
@@ -3744,19 +3203,11 @@ prepare_tpacket_socket(pcap_t *handle)
...
 			 */
-			return 1;
-		}
-		if (ret == -1) {
-			/*
-			 * We failed for some reason other than "the
-			 * kernel doesn't support TPACKET_V3".
-			 */
-			return -1;
+			return 0;
```
guyharris: Again, as per the above, I forgot that you have to distinguish between "TPACKET_V3 works", "TPACKET_V3 doesn't work because we only support TPACKET_V3", and "TPACKET_V3 doesn't work for some other reason that probably means TPACKET_V2 will fail as well", so this should treat 0 as success, -1 as a hard error, and 1 as "OK, fall through to the TPACKET_V2 code".

pcap-linux.c
```diff
@@ -3765,65 +3216,13 @@ prepare_tpacket_socket(pcap_t *handle)
...
 		 */
-		return 1;
-	}
-	if (ret == -1) {
-		/*
-		 * We failed for some reason other than "the
-		 * kernel doesn't support TPACKET_V2".
-		 */
-		return -1;
+		return 0;
```
guyharris: This, however, should fail even if `init_tpacket()` returns 1 (that "should not happen").

guyharris:
> OK, it can be merged (unless someone point out a defect, of course).

I think there might be a defect, because I forgot that `init_tpacket()` failing with TPACKET_V3 might not be a fatal error, it might just mean "this kernel doesn't support TPACKET_V3, so try TPACKET_V2".  Sorry about that.

Oppen:
Don't worry. I'll fix it later today.

PR #906, opened 2020-02-17
Linux: proper memory sync for PACKET_MMAP.

### LINKED ISSUE #898: High CPU usage during long captures
Hello,

I have decided to open an issue here because I think there might be a bug in libpcap that causes a thread in my software to use ~100% CPU under certain conditions.
I have searched online for similar issues, but nothing I found seems to apply to my case.
I apologize in advance for the length of the post.

SCENARIO:
I have instance A of a network proxy software that reads TCP/UDP/ICMP packets over a link using libpcap, extract data from those packets, sends those data to some remote instance B of the proxy using different protocols, and finally instance B recreates the original packets and uses libpcap to write those packets to the network. This happens in both directions, so A <--> B.
The software uses 2 pcap handlers for each monitored network interface, one to write packets and the other to read packets. So, in a typical configuration with 2 monitored interfaces, there are 4 different pcap handlers open.
CPU consumption of instances A and B varies depending on the amount of traffic processed, going from ~1-3% when very little traffic is present, to 80-90% when traffic is in the order of hundreds of Mbps.
For my tests, I generate TCP traffic using multiple iperf3 servers and clients running on two hosts (H1 and H2), one in the same network segment of Proxy A (PA) and one in the same network segment of Proxy B (PB):

H1 <-- IPERF3 TCP traffic --> PA <-- other network protocols --> PB <--- IPERF3 TCP traffic --> H2

PROBLEM:
The problem arises when I use iperf3 to generate high traffic for some time. After several minutes to several hours of high network loads, something happens and the CPU usage of the proxies always remain very high (> 95%), even if I turn off all traffic sources. More specifically, using a combination of top -H and gdb, I was able to verify that the thread that shows high CPU consumption is always the one that uses libpcap to read packets from the interface in the network segment on which we are generating traffic with iperf3.
I am able to reproduce the problem consistently, it just takes a variable amount of time each time. The issue does not appear at the same time on the two proxy instances.
Additionally, if I use 2 threads to read packets from the same network interface, using a different pcap handler for each thread, what happens is that the issue appears in one thread first, and then minutes after also the other thread starts consuming high amounts of CPU. The length of the interval between the moments when the problem appears in the two thread varies with each test.

TESTS AND DEBUGGING:
For my tests, I use Ubuntu 16.04 64bit VMs and ESXi 6.7 as host. All software is in C++, compiled with gcc 5.4.
I was able to reproduce the issue using the following versions of libpcap:
1. the version shipped with the libpcap-dev Ubuntu package, v1.7.4-2, linked dynamically;
2. the latest released version here on the official website, v1.9.1, compiled with debug symbols and linked statically;
3. the version compiled from the master branch on github, compiled with debug symbols and linked statically.

The pcap handlers are created with pcap_create() and, before activating them, I make the following calls:
 ```
 pcap_set_snaplen (pCapHandler, 65535);
 pcap_set_promisc (pCapHandler, 1);
 pcap_set_timeout (pCapHandler, 1);
 pcap_set_immediate_mode (pCapHandler, 1);
```
I am fairly positive that the parameters of these calls are not causing the problem, since high CPU usage only starts after the handler processes minutes to hours of high network traffic.

I have tried to destroy and create a new handler every hour, which mitigates the problem.
More specifically, doing this makes the problem disappear if it is currently happening (CPU load goes back to 1-3% under low network activity), but it could still happen again during the next hour and it won't go away until the handler is refreshed again.

I then decided to profile the code, since I have debug symbols available using the libpcap I built.
You can see the results of 3 profiling sessions in the images:

HIGH CPU - LOW Network

Above, profiling data were taken after the issue occurred and I switched off any network traffic. As you can see, the majority of the program is inside pcap_next_ex(). The CPU usage for the thread calling pcap_next_ex() was showing ~99%.

LOW CPU - HIGH Network
LOW CPU - LOW Network

The two images above show profiling data collected before the problem first manifests itself. The first image during high network traffic loads, and the second after I switched off all network traffic sources.
The time during which I collected profiling data to generate Figure 2 and 3 is much shorter (~2/3 minutes vs 20/30 minutes), so please, disregard the difference in the total count of function calls. If you think it could be useful, I will be happy to upload 3 images generated from 3 profiling sessions of the same duration.

Does anybody have any idea on what could be causing this behavior or can think of any other tests I could run to shed some light on the matter?

Thank you very much for your time and kind attention.

Sincerely,
Alessandro

Oppen:
Release and acquire packets using GCC builtins (compatible with
many other common-place compilers).

Fixes #898.

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -117,6 +117,15 @@
 # define HAVE_TPACKET3
 #endif /* TPACKET3_HDRLEN */
 
+#define packet_mmap_acquire(pkt) \
+	(__atomic_load_n(&pkt->tp_status, __ATOMIC_ACQUIRE) != TP_STATUS_KERNEL)
+#define packet_mmap_release(pkt) \
+	(__atomic_store_n(&pkt->tp_status, TP_STATUS_KERNEL, __ATOMIC_RELEASE))
+#define packet_mmap_v3_acquire(pkt) \
+	(__atomic_load_n(&pkt->hdr.bh1.block_status, __ATOMIC_ACQUIRE) != TP_STATUS_KERNEL)
+#define packet_mmap_v3_release(pkt) \
+	(__atomic_store_n(&pkt->hdr.bh1.block_status, TP_STATUS_KERNEL, __ATOMIC_RELEASE))
+
 #include <linux/types.h>
 #include <linux/filter.h>
 
@@ -3680,11 +3689,11 @@ pcap_get_ring_frame_status(pcap_t *handle, int offset)
 	h.raw = RING_GET_FRAME_AT(handle, offset);
 	switch (handlep->tp_version) {
 	case TPACKET_V2:
-		return (h.h2->tp_status);
+		return __atomic_load_n(&h.h2->tp_status, __ATOMIC_ACQUIRE);
 		break;
 #ifdef HAVE_TPACKET3
 	case TPACKET_V3:
-		return (h.h3->hdr.bh1.block_status);
+		return __atomic_load_n(&h.h3->hdr.bh1.block_status, __ATOMIC_ACQUIRE);
 		break;
 #endif
 	}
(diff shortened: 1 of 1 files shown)


mcr:
Can you rebase?

Oppen:
Sorry, I never got nofied D:
I'll probably do it tomorrow, I have to set up a VM to test it (I switched OS because of work).

Oppen:
Sorry, I procrastinated. It's rebased now.

PR #914, opened 2020-03-02
IO plugins for libpcap

mcr:
rebased upon master


pcap-ioplugin.c
```diff
@@ -0,0 +1,259 @@
...
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include "ftmacros.h"
+
+#ifdef _WIN32
+#include <pcap-stdinc.h>
```
guyharris, line 33: There's no such header file; it was removed in 787f37eeffdc3c37beeca8357d667ef35880c62b.

You need to directly include what's needed.

testprogs/selpolltest.c
```diff
@@ -74,7 +74,7 @@ main(int argc, char **argv)
 	struct bpf_program fcode;
 	char ebuf[PCAP_ERRBUF_SIZE];
 	pcap_if_t *devlist;
-	int selectable_fd;
```
guyharris, line 77: This change is already in the main branch from commit fbb5e9eec8bac69c7eb21e97ed5b20751f54290e (and is unrelated to the I/O plugins).

guyharris:
This looks as if it's not up to date with all the main branch changes, so it probably needs to be rebased.

fxlb:
Rebased.

pcap-ioplugin.c
```diff
@@ -0,0 +1,259 @@
...
+ * (on UNIX, a .so shared library)
+ *
+ * NB: fails silently and falls back to existing uncompressed
+ *     stdio-based output if the plugin fails to load
+ */
+
+const pcap_ioplugin_t*
+pcap_ioplugin_init(const char *name)
+{
+	void *lib = NULL;
```
fxlb, line 159:
```
./pcap-ioplugin.c: In function 'pcap_ioplugin_init':
./pcap-ioplugin.c:159:8: warning: unused variable 'lib' [-Wunused-variable]
  159 |  void *lib = NULL;
      |        ^~~
```

guyharris:
If we support plugins for *writing* files, and you're on a platform where packet capture requires elevated privileges, and the user can specify a plugin with an arbitrary path name with an environment variable, every program that uses the pcap dumping APIs needs to be Very Careful to surrender *all* elevated privileges before doing anything that causes the plugins to be loaded.

This requires some more thought.

fxlb:
Rebased.

pcap-ioplugin.c
guyharris: There are several issus here.

1. On UN*X, we need either to assume we have the `dlopen()` APIs or check for it in CMakeLists.txt as well.  It's specified in POSIX and, as far as I know, the only general-purpose UN*Xes that don't have it we're likely to encounter are 1) 32-bit HP-UX (which has its own API) and 2) perhaps older versions of AIX.  However, perhaps some embedded Linux distributions have a cut-back libc that doesn't have it.
2. We're not handling Windows which has its own APIs for this.
3. If we don't have any way to load plugins, we shouldn't define a variable to hold the plugin handle.

infrastation:
If the only functions implemented by a plugin would be a custom read function and a custom write function, would it be simpler to use `popen()` instead? For example, tar allows a number of different compression filters, each of which runs as a separate process.

infrastation:
the-tcpdump-group/tcpdump#458 seems to implement a similar idea, although the code seems a bit more complicated than I thought and uses `fork()` + `execvp()` instead of `popen()`.

infrastation:
the-tcpdump-group/tcpdump#935 is yet another alternative implementation.

PR #921, opened 2020-04-16
dpdk: add rte_ prefix to ether definitions

vivien:
DPDK commits 35b2d13fd6fd ("net: add rte prefix to ether defines")
and 6d13ea8e8e49 ("net: add rte prefix to ether structures") renamed
ETHER_MAX_JUMBO_FRAME_LEN to RTE_ETHER_MAX_JUMBO_FRAME_LEN and struct
ether_addr to struct rte_ether_addr.

Fix the compilation of pcap-dpdk.c, which currently fails with:

    error: ‘ETHER_MAX_JUMBO_FRAME_LEN’ undeclared here (not in a function); did you mean ‘RTE_ETHER_MAX_JUMBO_FRAME_LEN’?
    ...
    error: field ‘eth_addr’ has incomplete type
    ...
    error: dereferencing pointer to incomplete type ‘struct ether_addr’

Signed-off-by: Vivien Didelot <vivien.didelot@gmail.com>
Cc: @jingleyang

### DIFF
--- a/pcap-dpdk.c
+++ b/pcap-dpdk.c
@@ -152,7 +152,7 @@ static char dpdk_cfg_buf[DPDK_CFG_MAX_LEN];
 static uint16_t nb_rxd = RTE_TEST_RX_DESC_DEFAULT;
 static uint16_t nb_txd = RTE_TEST_TX_DESC_DEFAULT;
 
-#define RTE_ETH_PCAP_SNAPLEN ETHER_MAX_JUMBO_FRAME_LEN
+#define RTE_ETH_PCAP_SNAPLEN RTE_ETHER_MAX_JUMBO_FRAME_LEN
 
 static struct rte_eth_dev_tx_buffer *tx_buffer;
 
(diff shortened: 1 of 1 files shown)


guyharris:
> DPDK commits 35b2d13fd6fd ("net: add rte prefix to ether defines")
> and 6d13ea8e8e49 ("net: add rte prefix to ether structures") renamed
> ETHER_MAX_JUMBO_FRAME_LEN to RTE_ETHER_MAX_JUMBO_FRAME_LEN and struct
> ether_addr to struct rte_ether_addr.

So does this mean that, if this pull request is merged, DPDK will now fail to compile with *older* versions of DPDK?

infrastation:
In other words, the change should be to actually fix the broken build, not just to shift it out of sight. Traditionally a block like this does it:

```c
#if <new DPDK>
#define <what libpcap code needs> <what new DPDK has>
#else
#define <what libpcap code neews> <what old DPDK has>
#endif
```

There's a lot of examples of such accommodations in the source code. You can decode whether to check for a specific DPDK version/number or just for the fact a macro is defined.

guyharris:
Fixed in cc375cfa6a73647560346fab3b6a49042d397c47, in a way that works with older versions of DPDK.

PR #925, opened 2020-04-28
pcap-linux.c: use NETLINK socket when INET or UNIX are not available

sh1r4s3:
Since Linux kernel commit 025c68186e07 "netlink: add support for NIC
driver ioctls" it is now possible to fallback to NETLINK socket for
NIC ioctls.

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -90,6 +90,7 @@
 #include <linux/if_packet.h>
 #include <linux/sockios.h>
 #include <linux/ethtool.h>
+#include <linux/netlink.h>
 #include <netinet/in.h>
 #include <linux/if_ether.h>
 #include <linux/if_arp.h>
@@ -1728,6 +1729,24 @@ get_if_flags(const char *name, bpf_u_int32 *flags, char *errbuf)
 	}
 
 	sock = socket(AF_INET, SOCK_DGRAM, 0);
+	if (sock == -1) {
+		int save_errno = errno;
+		sock = socket(AF_NETLINK, SOCK_RAW, NETLINK_GENERIC);
+		/*
+		 * In the case of EOPNOTSUPP we suppose that .ioctl in the Linux
+		 * netlink_ops is set to sock_no_ioctl and AF_NETLINK socket does
+		 * not support ioctls yet. So, provide saved errno to the error
+		 * message.
+		 */
+		if (sock == -1 && errno == EOPNOTSUPP)
+			errno = save_errno;
+		else {
+			fprintf(stderr,
+				"Warning: Can't create socket to get ethtool information for %s: %s\n"
+				"Fallback to NETLINK socket\n",
+				name, pcap_strerror(save_errno));
+		}
+	}
 	if (sock == -1) {
 		pcap_fmt_errmsg_for_errno(errbuf, PCAP_ERRBUF_SIZE, errno,
 		    "Can't create socket to get ethtool information for %s",
@@ -5426,6 +5445,24 @@ iface_ethtool_get_ts_info(const char *device, pcap_t *handle, char *ebuf)
 	 * Create a socket from which to fetch time stamping capabilities.
 	 */
 	fd = socket(PF_UNIX, SOCK_RAW, 0);
+	if (fd < 0) {
+		int save_errno = errno;
+		fd = socket(AF_NETLINK, SOCK_RAW, NETLINK_GENERIC);
+		/*
+		 * In the case of EOPNOTSUPP we suppose that .ioctl in the Linux
+		 * netlink_ops is set to sock_no_ioctl and AF_NETLINK socket does
+		 * not support ioctls yet. So, provide saved errno to the error
+		 * message.
+		 */
+		if (fd < 0 && errno == EOPNOTSUPP)
+			errno = save_errno;
+		else {
+			fprintf(stderr,
+				"Warning: Can't create socket for SIOCETHTOOL(ETHTOOL_GET_TS_INFO)\n"
+				"Fallback to NETLINK socket\n",
+				pcap_strerror(save_errno));
+		}
+	}
 	if (fd < 0) {
 		pcap_fmt_errmsg_for_errno(ebuf, PCAP_ERRBUF_SIZE,
 		    errno, "socket for SIOCETHTOOL(ETHTOOL_GET_TS_INFO)");


guyharris:
Is there any reason to expect that we won't be able to create an `AF_INET` or `AF_UNIX` socket but *will* be able to create an `AF_NETLINK` socket?

sh1r4s3:
I ran into this situation in SELinux hardened system. It has a policy which does not allow to create a `AF_UNIX` socket (even as root) but it allows to create a `AF_NETLINK` socket. With this commit it is becomes possible to use, for example, `tcpdump -i eth0` in this system. I could also imagine another (a little bit weird) situation when Linux kernel was built without `AF_UNIX` or `AF_INET` support for some reason but with 'AF_NETLINK' enabled . The point of this commit is to try to benefit from the Linux kernel commit 025c68186e07 in the case of `AF_UNIX` or `AF_INET` socket creation fail.

guyharris:
> I ran into this situation in SELinux hardened system. It has a policy which does not allow to create a `AF_UNIX` socket (even as root) but it allows to create a `AF_NETLINK` socket.

And an `AF_PACKET` socket, presumably, so that traffic capture can be done.  What is t

So does that mean it can't even create an `AF_UNIX` socket with `socketpair()`?  If so, is there any *other* mechanism by which, for example, a program could run a privileged helper to get an `AF_PACKET` socket (or a USB monitor descriptor) without itself having to run with the requisite privileges?

And why are `AF_UNIX` sockets considered so dangerous?

And are there policies that prevent use of `AF_INET` as well?  If so, does that mean that the policy in question prevents programs from doing address-to-name translations using DNS, or do routines like `gethostbyaddr()` talk to a daemon (by whatever mechanisms SELinux *doesn't* forbid in any useful policy), with the daemon being allowed to send UDP packets to a DNS server?

> With this commit it is becomes possible to use, for example, `tcpdump -i eth0` in this system.

sh1r4s3:
> And an `AF_PACKET` socket, presumably, so that traffic capture can be done. What is t

Yes it is.

> So does that mean it can't even create an `AF_UNIX` socket with `socketpair()`? If so, is there any _other_ mechanism by which, for example, a program could run a privileged helper to get an `AF_PACKET` socket (or a USB monitor descriptor) without itself having to run with the requisite privileges?

It can't create `AF_UNIX` with `socketpair()`. Could it use pipes, instead, for example?

> 
> And why are `AF_UNIX` sockets considered so dangerous?
> 

`AF_UNIX` allows to transfer not labeled data (using socket file) and it is becomes not possible to control the transferred data using iptables, netfilter with extensions to use mandatory access control with additional lables based on RFC 1108 [1].

> And are there policies that prevent use of `AF_INET` as well? If so, does that mean that the policy in question prevents programs from doing address-to-name translations using DNS, or do routines like `gethostbyaddr()` talk to a daemon (by whatever mechanisms SELinux _doesn't_ forbid in any useful policy), with the daemon being allowed to send UDP packets to a DNS server?

No. This policy does not prevent from using `AF_INET`. The reason why I have added an alternative way through `AF_NETLINK` for `AF_INET` is because it is possible to get the required information not only with `AF_INET` but also using `AF_NETLINK`.

[1] https://tools.ietf.org/html/rfc1108

mcr:
> I could also imagine another (a little bit weird) situation when Linux kernel was built without `AF_UNIX` or `AF_INET` support for some reason but with 'AF_NETLINK' enabled . 

Perhaps the kernel has support, but the container in question does not. Not sure that's possible today, but it might be.

sh1r4s3:
Thanks for the answers!

> It seems that if we can AF_NETLINK instead of AF_INET/AF_UNIX then we should just do that, and fallback to the old way if it doesn't work. That way, leads to ripping the old code easier.

I'm not sure that it is worth to do that. The `AF_NETLINK` is just an another way (among `AF_INET`, `AF_UNIX`, `AF_PACKET`, etc) to ask something of NIC driver ioctls, `SIOCETHTOOL`, etc. It is not better nor worse than the other sockets for these purposes, except the cases similar to the described above. The feature of `AF_NETLINK` mentioned in the commit message is available since Linux v4.6 (Linux 025c68186e07), so if an end-user Linux system does not have the ioctls support in `AF_NETLINK` yet, it would create the `AF_NETLINK` socket, but later the ioctl syscall would fail. So in that case the program would never hit a fallback branch with `AF_INET`/`AF_UNIX`.

guyharris:
> > So does that mean it can't even create an `AF_UNIX` socket with `socketpair()`? If so, is there any _other_ mechanism by which, for example, a program could run a privileged helper to get an `AF_PACKET` socket (or a USB monitor descriptor) without itself having to run with the requisite privileges?
> 
> It can't create `AF_UNIX` with `socketpair()`. Could it use pipes, instead, for example?

Only if you can send FDs across a pipe.

> > And are there policies that prevent use of `AF_INET` as well? If so, does that mean that the policy in question prevents programs from doing address-to-name translations using DNS, or do routines like `gethostbyaddr()` talk to a daemon (by whatever mechanisms SELinux _doesn't_ forbid in any useful policy), with the daemon being allowed to send UDP packets to a DNS server?
> 
> No. This policy does not prevent from using `AF_INET`. The reason why I have added an alternative way through `AF_NETLINK` for `AF_INET` is because it is possible to get the required information not only with `AF_INET` but also using `AF_NETLINK`.

Is there any reason *not* to do so using `AF_INET`, if it's no less likely to be supported and if it's *more* likely to be supported in older kernels?

guyharris:
It appears that the ethtool command uses `AF_INET` sockets, rather than `AF_UNIX` sockets, falling back on `AF_NETLINK` if that fails.

Is there any reason to ever use `AF_UNIX` sockets for `SIOCETHTOOL`?

Presumably the only reason why creating an `AF_INET` socket would fail but creating an `AF_NETLINK` socket would succeed is that the kernel was configured without IPv4 support.

Oppen:
> Thanks for the answers!
> 
> > It seems that if we can AF_NETLINK instead of AF_INET/AF_UNIX then we should just do that, and fallback to the old way if it doesn't work. That way, leads to ripping the old code easier.
> 
> I'm not sure that it is worth to do that. The `AF_NETLINK` is just an another way (among `AF_INET`, `AF_UNIX`, `AF_PACKET`, etc) to ask something of NIC driver ioctls, `SIOCETHTOOL`, etc. It is not better nor worse than the other sockets for these purposes, except the cases similar to the described above.

Of course, we couldn't reap it *right now*, but *in the distant future* (when we decide to support 4.6 onwards). If one solution works on all supported systems and does it at least as well as the others for all of them, then that solution *is* better, and it's enough. All else equal, having a single code path to debug is always better.

> Presumably the only reason why creating an `AF_INET` socket would fail but creating an `AF_NETLINK` socket would succeed is that the kernel was configured without IPv4 support.
So it becomes a matter of whether there are useful (and currently supported) use cases where we use a NIC but don't need IPv4 in the kernel. It seems like a stretch, but maybe something that only process in userspace, or some in-house protocol implemented, possibly, with libpcap, such as those who would use immediate mode.

guyharris:
Should be addressed by e65d3cab20910a693a30e18b37576d6649be2e61.

PR #928, opened 2020-05-07
avoid integer multiply overflow

ltx2018:


optimize.c
```diff
@@ -2164,7 +2164,7 @@ static void
 opt_init(opt_state_t *opt_state, struct icode *ic)
 {
 	bpf_u_int32 *p;
-	int i, n, max_stmts;
+	int i, n, max_stmts, block_memsize, edge_memsize;
```
guyharris, line 2167: `block_memsize` and `edge_memsize` should probably be `size_t`, as 1) they shouldn't become negative and 2) on LP64 and LLP64 platforms, you want them to be bigger than 32 bits.

optimize.c
```diff
@@ -2196,9 +2196,17 @@ opt_init(opt_state_t *opt_state, struct icode *ic)
 	opt_state->edgewords = opt_state->n_edges / (8 * sizeof(bpf_u_int32)) + 1;
 	opt_state->nodewords = opt_state->n_blocks / (8 * sizeof(bpf_u_int32)) + 1;
 
+	/*
+	 * Check integer overflow.
```
guyharris, line 2200: ...and "integer" should be replaced by "size_t".

optimize.c
```diff
@@ -2196,9 +2196,17 @@ opt_init(opt_state_t *opt_state, struct icode *ic)
...
+	block_memsize = 2 * opt_state->n_blocks * opt_state->nodewords * sizeof(*opt_state->space);
+	edge_memsize = opt_state->n_edges * opt_state->edgewords * sizeof(*opt_state->space);
+	if (opt_state->n_edges && edge_memsize / sizeof(*opt_state->space) / opt_state->n_edges != opt_state->edgewords) {
+		opt_error(opt_state, "integer overflow");
+	}
+
 	/* XXX */
-	opt_state->space = (bpf_u_int32 *)malloc(2 * opt_state->n_blocks * opt_state->nodewords * sizeof(*opt_state->space)
-				 + opt_state->n_edges * opt_state->edgewords * sizeof(*opt_state->space));
+	opt_state->space = (bpf_u_int32 *)malloc(block_memsize + edge_memsize);
```
guyharris, line 2209: Note that the sum could conceivably overflow even if neither number *itself* overflows.

ltx2018: block and edge both int，malloc() change the sum to size_t，i think it would cause overflows

optimize.c
ltx2018: thanks for your advice

PR #933, opened 2020-05-15
fix pktgen-dpdk cross compilation with libpcap library

rohit15079:
While cross compiling an applicaion with libpcap library,
if the function BPF_STMT and BPF_JUMP already defined in
toolchain, it fails to build.

This pach fixes this issue by checking if the function is
defined already before defining it.

Signed-off-by: Rohit Raj <rohit.raj@nxp.com>

### DIFF
--- a/pcap/bpf.h
+++ b/pcap/bpf.h
@@ -241,8 +241,12 @@ struct bpf_insn {
 /*
  * Macros for insn array initializers.
  */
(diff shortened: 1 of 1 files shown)


guyharris:
0bb0b261a05bd0f396429bf99984343c9bade3b1 should resolve this.

PR #934, opened 2020-05-29
Etw

citronneur:
Hello libpcap Team,

I would like to propose you a pull request that allow the libpcap to capture ETW on windows platform.
This pull request is part of a project named Winshark that will be presented at [SSTIC](https://www.sstic.org/2020/news/), that allow Wireshark to capture and analyze ETW.

Have a nice day,

Sylvain


pcap-etw.c
```diff
@@ -0,0 +1,397 @@
...
+	pcap_t *handle = pcap_create_common(ebuf, sizeof(pcap_etw_t));
+	if (handle == NULL)
+		return NULL;
+
+	handle->activate_op = pcap_activate_op_etw;
+	handle->setfilter_op = pcap_setfilter_op_etw;
+	handle->stats_op = pcap_stats_etw;
+	handle->read_op = pcap_read_op_etw;
+	handle->snapshot = MAXIMUM_SNAPLEN;
+	handle->linktype = DLT_USER0;
```
guyharris, line 315: Don't use a DLT_USER value; send an email to tcpdump-workers@lists.tcpdump.org to request that a DLT_ value be assigned for this purpose.

guyharris:
This should add `pcap-etw.c` to the `EXTRA_DIST` list in Makefile.in, even though it won't be used on UN\*X, so that the release tarball will include it.

pcap-etw.c
citronneur: I'm just sending  the email. Thanks for your feedback.

citronneur:
I've added pcap-etw.c into MANIFEST.in but it seems that the build is broken for network reason...

guyharris:
> I've added pcap-etw.c into MANIFEST.in but it seems that the build is broken for network reason...

Neither Travis nor AppVeyor are reliable CI engines; they both sometimes need to have builds re-tried to get them to finish.

infrastation:
As a generic observation, please use a consistent curly braces style, currently the file alternates between Allman and K&R.

Since this is a case of a "try-fixup-fixup-fixup" sequence of commits, if anyone comes to the point of merging this request, please do that as a single squashed commit.

citronneur:
Thanks a lot!

citronneur:
Rebase commit from PR #978, squash into one commit.
Have a nice day

PR #935, opened 2020-05-29
RFC: Remove undocumented and rather old "ether proto" protocols

fxlb:
They were never documented (at least since 1999).

The only documented protocols are:
```
    ether proto protocol
          True if the packet is of ether type protocol.  Protocol can be a
          number or one of the names ip, ip6, arp, rarp, atalk, aarp, dec-
          net,  sca,  lat,  mopdl, moprc, iso, stp, ipx, or netbeui.
```
(pcap-filter man page)

They are rather old and unlikely to be found in a capture file.

Summary:
```
protocol	value	IANA ieee-802-numbers.xhtml
--------        -----   ---------------------------
decdns		803c	DEC Unassigned
decdts		803e	DEC Unassigned
lanbridge	8038	DEC LANBridge
loopback	9000	Loopback
pup		0200	XEROX PUP
sprite		0500	<UNKWOWN in IANA Registry>
vexp		805b	Stanford V Kernel exp.
vprod		805c	Stanford V Kernel prod.
xns		0600	XEROX NS IDP
```

### DIFF
--- a/pcap-filter.manmisc.in
+++ b/pcap-filter.manmisc.in
@@ -361,9 +361,9 @@ True if the packet is an IPv6 multicast packet.
 .IP  "\fBether proto \fIprotocol\fR"
 True if the packet is of ether type \fIprotocol\fR.
 \fIProtocol\fP can be a number or one of the names
-\fBip\fP, \fBip6\fP, \fBarp\fP, \fBrarp\fP, \fBatalk\fP, \fBaarp\fP,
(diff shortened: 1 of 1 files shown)


fxlb:
You should keep 'loopback' and add it in the man page.
It is used by tcpdump.

fxlb:
I will update for 'loopback'.

fxlb:
Merged a first step, not the PR with the title.
I will open a new PR.

PR #936, opened 2020-05-29
RFC: Remove undocumented and rather old "ether proto" protocols

fxlb:
They were never documented (at least since version 0.4 - 1999).

The only documented protocols are:
```
    ether proto protocol
          True if the packet is of ether type protocol.  Protocol can be a
          number  or  one  of the names aarp, arp, atalk, decnet, ip, ip6,
          ipx, iso, lat, mopdl, moprc, netbeui, rarp, sca  or  stp.
```
(pcap-filter man page)

They are rather old and unlikely to be found in a capture file.

Summary:
```
protocol	value	IANA ieee-802-numbers.xhtml
--------        -----   ---------------------------
decdns		803c	DEC Unassigned
decdts		803e	DEC Unassigned
lanbridge	8038	DEC LANBridge
pup		0200	XEROX PUP
sprite		0500	<UNKWOWN in IANA Registry>
vexp		805b	Stanford V Kernel exp.
vprod		805c	Stanford V Kernel prod.
xns		0600	XEROX NS IDP
```

### DIFF
--- a/nametoaddr.c
+++ b/nametoaddr.c
@@ -593,25 +593,17 @@ PCAP_API_DEF struct eproto eproto_db[] = {
 	{ "arp", ETHERTYPE_ARP },
 	{ "atalk", ETHERTYPE_ATALK },
 	{ "atalkarp", ETHERTYPE_AARP },
-	{ "decdns", ETHERTYPE_DECDNS },
-	{ "decdts", ETHERTYPE_DECDTS },
 	{ "decnet", ETHERTYPE_DN },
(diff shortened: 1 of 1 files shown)


fxlb:
Replace PR #935.

infrastation:
Commit 76bb2ce documents `ether proto loopback`, which already works, in this case `loopback` is an identifier. Support for `loopback` keyword, which would expand to `ether proto \loopback`, still remains to be added as commented above.

PR #946, opened 2020-06-30
mac: Add support for building arm

rknall:
Add support for selecting the appropriate architecture if
building within big sur. This enables a fat binary, which
includes code for arm64 as well as x86_64 targets

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2447,7 +2447,7 @@ if(APPLE AND "${CMAKE_OSX_ARCHITECTURES}" STREQUAL "")
         # captures.)
         #
         set(OSX_LIBRARY_ARCHITECTURES "x86_64;i386;ppc")
-    else()
+    elseif(SYSTEM_VERSION_MAJOR LESS 12)
         #
         # Post-Snow Leopard.  Build for x86-64 and 32-bit x86,
         # with x86-64 first.  (That's what Apple does)
@@ -2496,6 +2496,11 @@ main(void)
               endif()
           endif()
         endif()
+    else()
+        #
+        # Big Sur.  Build for x86-64, arm64, with x86-64 first. 
+        #
+        set(OSX_LIBRARY_ARCHITECTURES "arm64;x86_64")
     endif()
     if(BUILD_SHARED_LIBS)
         set_target_properties(${LIBRARY_NAME} PROPERTIES


guyharris:
Change configure.ac and re-run autoconf as well.

fxlb:
This change gives errors:
https://travis-ci.org/github/the-tcpdump-group/libpcap/jobs/703676181
and
https://travis-ci.org/github/the-tcpdump-group/libpcap/jobs/703676183

rknall:
I will rewrite the change to actively check for arm. But the issue then will be, how to provide for building all necessary plattforms

guyharris:
> This change gives errors:
> https://travis-ci.org/github/the-tcpdump-group/libpcap/jobs/703676181
> and
> https://travis-ci.org/github/the-tcpdump-group/libpcap/jobs/703676183

`/usr/include/sys/cdefs.h`?  That's not supposed to happen on newer versions of macOS; they no longer provide `/usr/include`, they bury the headers under `/Applications/Xcode`.

guyharris:
> I will rewrite the change to actively check for arm. But the issue then will be, how to provide for building all necessary plattforms

What do you mean by "actively check for arm"?  And what defines "all necessary platforms"?  The platforms are version-dependent, as the platforms supported by macOS are version-dependent.

rknall:
> > I will rewrite the change to actively check for arm. But the issue then will be, how to provide for building all necessary plattforms
> 
> What do you mean by "actively check for arm"? And what defines "all necessary platforms"? The platforms are version-dependent, as the platforms supported by macOS are version-dependent.

The idea is to check for arm availibility in the compiler, similar as it is checked for 32bit on certain plattforms. Then the architecture arm64 will only be added if the compiler can implement it. clang on BigSur/arm builds arm by default, I am not so sure if XCode 12 on Catalina could build arm64 atm. My current assumption is as such, but have not seen any proof for it yet.

guyharris:
> > > I will rewrite the change to actively check for arm. But the issue then will be, how to provide for building all necessary plattforms
> > 
> > 
> > What do you mean by "actively check for arm"? And what defines "all necessary platforms"? The platforms are version-dependent, as the platforms supported by macOS are version-dependent.
> 
> The idea is to check for arm availibility in the compiler, similar as it is checked for 32bit on certain plattforms.

My original goal was just to duplicate the way libpcap was built in macOS, rather than to build fat for everything that's supported.

The 32-bit check is to avoid trying to build fat in ways that only work in Apple's internal build environment (which may have, for example, headers that aren't available outside Apple, such as the OpenSSL headers).

> Then the architecture arm64 will only be added if the compiler can implement it. clang on BigSur/arm builds arm by default, I am not so sure if XCode 12 on Catalina could build arm64 atm. My current assumption is as such, but have not seen any proof for it yet.

So you may also have to avoid building for Arm on platforms where that's not supported.  By default, we shouldn't even *try* building for Arm unless we're building on Big Sur, as the default should be to try to build libpcap the same way it's built on for the OS on machine on which the build is being done.

guyharris:
1b6ace5c1a32a5eb526e8e07c518334e4de16468 1) doesn't try to build for 32-bit x86 on Catalina and 2) will try to build 2-way fat on Big Sur and later, if either 1) we're not building with libssl or 2) we can build fat with libssl.

PR #948, opened 2020-07-02
Call PacketSetSnapLen even if non-DAG card. See #947

### LINKED ISSUE #947: Call PacketSetSnapLen() in pcap-npf.c for non-DAG cards
As described in [this bug on Npcap](https://github.com/nmap/npcap/issues/201), Npcap/WinPcap implementation of snaplen is currently dependent on libpcap functions being called in a particular order. We would like to fix this by actually implementing a native "snaplen" config functionality, but we need libpcap to call it in order for it to be useful. Right now, `PacketSetSnapLen()` does nothing on non-DAG cards, so libpcap does not call it. I have a PR incoming that will address this issue.

dmiller-nmap:
Fixes #947

### DIFF
--- a/pcap-npf.c
+++ b/pcap-npf.c
@@ -1218,8 +1218,6 @@ pcap_activate_npf(pcap_t *p)
 		while(FALSE);
 
 
-		p->snapshot = PacketSetSnapLen(pw->adapter, p->snapshot);
-
 		/* Set the length of the FCS associated to any packet. This value
 		 * will be subtracted to the packet length */
 		pw->dag_fcs_bits = pw->adapter->DagFcsLen;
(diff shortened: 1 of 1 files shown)


guyharris:
Or we could just set an "accept all packets, with this snapshot length" filter in the activate routine, as we do in `pcap-bpf.c`.

gvanem:
Maybe we can also get rid of this ugly statement at the same time:
```c
#ifdef HAVE_DAG_API
	}
#endif /* HAVE_DAG_API */
```

dmiller-nmap:
After discussion on the Npcap issue documenting the deficiency, I think that setting an "accept all" filter is the better solution. I'll open a PR for that in a few minutes.

dmiller-nmap:
I was too late! 6c893c13

guyharris:
> Maybe we can also get rid of this ugly statement at the same time:
> 
> ```c
> #ifdef HAVE_DAG_API
> 	}
> #endif /* HAVE_DAG_API */
> ```

Maybe we can just get rid of all the DAG support code in `pcap-npf.c` and in `Packet32.dll`.  I think that was put in there before libpcap had support for multiple modules; now that we have that support, *if* Endace still supports DAG cards (they don't seem to be mentioned as a product on their website), and *if* they support them on Windows, we can modify the existing `pcap-dag.c` as necessary or, if the Windows DAG API is so different that such a modification makes no sense, provide a `pcap-dag-windows.c` file for them.

Is Npcap even built with DAG support?  That requires enabling `HAVE_DAG_API`, which requires having the DAG API libraries and headers installed.  If not, removing it won't make any difference, except to people who build it themselves - they can file a libpcap issue and we can fix that with a libpcap module.

PR #949, opened 2020-07-02
Make packet_version_string const to match PacketGetVersion return type LPCSTR

dmiller-nmap:
Minor change. We (Npcap) made a bunch of things `const` a while back, so the compiler issues a warning now for this.

### DIFF
--- a/pcap-npf.c
+++ b/pcap-npf.c
@@ -2125,7 +2125,7 @@ pcap_lib_version(void)
 		/*
 		 * Generate the version string.
 		 */
-		char *packet_version_string = PacketGetVersion();
+		const char *packet_version_string = PacketGetVersion();
 
 		if (strcmp(WINPCAP_VER_STRING, packet_version_string) == 0) {
 			/*


guyharris:
In which version of the Npcap SDK was that made `const`?  I'm not getting a warning with the existing code on [AppVeyor](https://ci.appveyor.com/project/guyharris/libpcap/builds/33841648) for that particular issue, in any of the builds; it's building with `/Wall` and it should be building with the Npcap 1.05 SDK?

Or is there something else keeping MSVC from warning about that?

guyharris:
> In which version of the Npcap SDK was that made `const`?

If [Microsoft's documentation](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/f8d4fe46-6be8-44c9-8823-615a21d17a61) is to be believed, it's `const` in 1.05:

    #ifdef __cplusplus
    extern "C"
    {
    #endif
    
...

    	//---------------------------------------------------------------------------
    	// EXPORTED FUNCTIONS
    	//---------------------------------------------------------------------------
    
    	LPCSTR PacketGetVersion();

So why no warning?

guyharris:
So [this build](https://ci.appveyor.com/project/guyharris/libpcap/builds/33882439/job/094wduakho1cd9wi) failed pretty quickly when I tested it with

    extern const char *func1(void);
    
    const char *
    func2(void)
    {
    	char *str = func1();
    	return str;
    }

So I'm not sure why it's not failing on every build with the Npcap SDK.

guyharris:
AppVeyor was doing a "build libpcap on its own, using an SDK" build, rather than a "build as part of a WinPcap/Npcap build" build, so it wasn't compiling the code in question.

Issue #955, opened 2020-08-03
rpcapd: Link to shared library rather than statically


netrounds-fredrik:
When compiling the rpcapd application on a UNIX system, libpcap is linked statically rather than dynamically (https://github.com/the-tcpdump-group/libpcap/blob/6fedfcefe327395587fe08ce58a80c0fd1bfc09c/rpcapd/CMakeLists.txt#L119).

In a production environment you might want to use the existing libpcap installed on your system in order to get security updates from your Linux distro and to avoid having multiple binary copies of the library, so it would be good it it was possible to build rpcapd without static linking like you would normally build an application.

guyharris:
> In a production environment you might want to use the existing libpcap installed on your system

Which sounds as if there should be a separate repository for rpcapd, so you can build it separately from libpcap.

netrounds-fredrik:
It would make sense to be able to build rpcapd separate from libpcap as they are not that tightly coupled from an API point of view. They can however still live in the same repo, just like libcurl and the curl command line tool do.

Is there a reason why rpcapd SHOULD link libpcap statically? There might be some low level dependency between them that I'm not aware of.

mcr:
Fredrik Kers <notifications@github.com> wrote:
    > Is there a reason why rpcapd SHOULD link libpcap statically? There
    > might be some low level dependency between them that I'm not aware of.

It is because (i seem recall) most of the support for generated shared
objects is in the distro packaging.  We don't by default, generate a ".so".

--
]               Never tell me the odds!                 | ipv6 mesh networks [
]   Michael Richardson, Sandelman Software Works        |    IoT architect   [
]     mcr@sandelman.ca  http://www.sandelman.ca/        |   ruby on rails    [

infrastation:
Quite the opposite, libpcap does generate a .so by default because that is its main purpose.

Perhaps a good reason to link rpcapd statically for the time being is that rpcapd is still a work in progress and may depend on libpcap features that are still work in progress, so you can try today's rpcapd without introducing a bleeding edge today's version of a shared library into a working system.

Also there is a maintenance effort cost to having one more repository given the software needs to be portable and needs to have two build systems (autotools and CMake).

netrounds-fredrik:
From what I can see without digging too much into the details it doesn't seem to be too much work to make it possible to build rpcapd against a shared library, still in the same repo. Would you be interested in a PR for this once we get time to look into it?

infrastation:
[...]

PR #961, opened 2020-09-02
Add separate ENABLE_RPCAPD option to build rpcapd

netrounds-fredrik:
This PR is related to https://github.com/the-tcpdump-group/libpcap/issues/955 and shows how `rpcapd` can be build without having to enable the new and experimental "remote" `libpcap` API (`ENABLE_REMOTE`).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -205,9 +205,11 @@ set(PCAP_TYPE "" CACHE STRING "Packet capture type")
 # not having it on UN*X.
 #
 if(WIN32)
-    option(ENABLE_REMOTE "Enable remote capture" ON)
+    option(ENABLE_RPCAPD "Enable building of rpcapd" ON)
+    option(ENABLE_REMOTE "Enable libpcap API for connecting to rpcapd" ON)
 else()
-    option(ENABLE_REMOTE "Enable remote capture" OFF)
+    option(ENABLE_RPCAPD "Enable building of rpcapd" OFF)
+    option(ENABLE_REMOTE "Enable libpcap API for connecting to rpcapd" OFF)
 endif(WIN32)
 
 if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
@@ -2202,9 +2204,9 @@ endif(WIN32)
 # Add subdirectories after we've set various variables, so they pick up
 # pick up those variables.
 #
-if(ENABLE_REMOTE)
+if(ENABLE_RPCAPD)
     add_subdirectory(rpcapd)
-endif(ENABLE_REMOTE)
+endif(ENABLE_RPCAPD)
 add_subdirectory(testprogs)
 
 ######################################
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,7 @@ env:
     - REMOTE=disable CMAKE=no
     - ENABLE_REMOTE="" CMAKE=yes
     - REMOTE=enable CMAKE=no
-    - ENABLE_REMOTE="-DENABLE_REMOTE=ON" CMAKE=yes
+    - ENABLE_REMOTE="-DENABLE_REMOTE=ON -DENABLE_RPCAPD=ON" CMAKE=yes
 
 matrix:
   fast_finish: true


guyharris:
So you're building the experimental daemon for a protocol, but not the client code for that protocol? :-)

guyharris:
1. This should be supported both by CMake *and* autotools.

1. There should be a *single* flag that enables both client and server, with a flag that enables just the server, and perhaps *another* flag to enable just the client.  (Either that, or a single flag that can specify "don't enable any remote capture", "enable only the client", "enable only the server", or "enable both".)

netrounds-fredrik:
> So you're building the experimental daemon for a protocol, but not the client code for that protocol? :-)

Yes. A use-case for this is to run `rpcapd` on a Linux machine and Wireshark on Windows which has support for the rpcap protocol (https://qnaplus.com/capture-protocol-packets-from-remote-windows-or-linux-computer-using-wireshark/).

CMakeLists.txt
```diff
@@ -205,9 +205,11 @@ set(PCAP_TYPE "" CACHE STRING "Packet capture type")
 # not having it on UN*X.
 #
 if(WIN32)
-    option(ENABLE_REMOTE "Enable remote capture" ON)
+    option(ENABLE_RPCAPD "Enable building of rpcapd" ON)
+    option(ENABLE_REMOTE "Enable libpcap API for connecting to rpcapd" ON)
 else()
-    option(ENABLE_REMOTE "Enable remote capture" OFF)
+    option(ENABLE_RPCAPD "Enable building of rpcapd" OFF)
```
mcr, line 211:
```suggestion
    option(ENABLE_RPCAPD "Enable building of remote capture daemon (rpcapd)" OFF)
```

CMakeLists.txt
```diff
@@ -205,9 +205,11 @@ set(PCAP_TYPE "" CACHE STRING "Packet capture type")
...
+    option(ENABLE_REMOTE "Enable libpcap API for connecting to rpcapd" OFF)
```
mcr, line 212:
```suggestion
    option(ENABLE_REMOTE "Enable libpcap client API for collecting remote captures via rpcapd" OFF)
```

mcr:
> 1. This should be supported both by CMake _and_ autotools.

Yes.

> 2. There should be a _single_ flag that enables both client and server, with a flag that enables just the server, and perhaps _another_ flag to enable just the client.  (Either that, or a single flag that can specify "don't enable any remote capture", "enable only the client", "enable only the server", or "enable both".)

I think that this is a step forward, and if there is demand for that flag, I guess we can add it.

guyharris:
> Yes. A use-case for this is to run `rpcapd` on a Linux machine and Wireshark on Windows which has support for the rpcap protocol (https://qnaplus.com/capture-protocol-packets-from-remote-windows-or-linux-computer-using-wireshark/).

That's also a use case for an option not to build libpcap *at all*, just rpcapd.  Yes, `./configure --disable-libpcap` sounds a bit weird if you're running libpcap's configure script, but.... :-)

mcr:
Guy Harris <notifications@github.com> wrote:
    > That's also a use case for an option not to build libpcap *at all*,
    > just rpcapd.  Yes, `./configure --disable-libpcap` sounds a bit weird
    > if you're running libpcap's configure script, but.... :-)

Building in libpcap statically rather than risk confusion with "system"
libpcap makes a lot of sense to me.
Even more so on systems that might not have libpcap at all, and one ships
rpcapd as part of some debug enablement.

antoniovazquezblanco:
I have discovered rpcapd recently.

Looking at this PR I wonder if it makes sense to have a separate repos for libpcap and rpcapd. It seems that it makes sense to enable and disable both libpcap building as well as rpcapd server and client.

Please excuse me if this makes little or no sense as I am getting in touch with this tool.

Thanks!

guyharris:
> Looking at this PR I wonder if it makes sense to have a separate repos for libpcap and rpcapd.

The libpcap client for the rpcap protocol, and the rpcapd server for the rpcap protocol, share:

- the sockutils.c code that provides APIs for networking that 1) provide some convenience and 2) handle differences between UN\*X sockets and WinSock and between various UN\*X socket implementations;
- the sslutils.c code that provides convenience routines for doing TLS;
- the rpcap-protocol.c code that provides APIs for handling the rpcap protocol;
- the fmtutils.c code for formatting error messages.

Having separate copies of those in two repositories would complicate maintenance of that code - changes to that code in one repository would have to be propagated to the other repository.  There might be ways to untangle the dependencies, but it'd take more thought to come up with a good way to do it.

PR #962, opened 2020-09-02
Add ENABLE_RPCAPD_DYNAMIC to make rpcapd link to libpcap shared library

netrounds-fredrik:
This PR is related to https://github.com/the-tcpdump-group/libpcap/issues/955 and shows how `rpcapd` can dynamically link to the `libpcap` shared library rather than linking it statically.

It can also be used in combination with https://github.com/the-tcpdump-group/libpcap/pull/961 on order to disable the experimental "remote" API in `libpcap`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -210,6 +210,8 @@ else()
     option(ENABLE_REMOTE "Enable remote capture" OFF)
 endif(WIN32)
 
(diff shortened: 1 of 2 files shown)


infrastation:
The open questions still are "Why would it be right to link rpcapd with libpcap dynamically at this stage of development?" and "What specific use case would this change solve?". You have received plenty of feedback in #955.

PR #965, opened 2020-09-16
Add split pcap dumping functionality to PCAP API

scottfgjordan:
Hello!

I've created a pull request here to add in a functionality to allow for dumping to a pcap dumper in two separate parts. I came across a use case where packet header and payload were held separately and I used this approach to allow for dumping them in unison. I've followed the same approach used in pcap_dump, but with additional writes to handle the two separate parts to dump. It's fairly straightforward code, but I had two questions about what I've done so far:
1. It wasn't clear what kind test setup you have for the pcap_dump code so I haven't added in any test cases. Is this something you would like to see? If so how should I set that up?
2. Wondering about the function name and if there's something else you'd prefer.

### DIFF
--- a/pcap/pcap.h
+++ b/pcap/pcap.h
@@ -596,6 +596,8 @@ PCAP_API int64_t	pcap_dump_ftell64(pcap_dumper_t *);
 PCAP_API int	pcap_dump_flush(pcap_dumper_t *);
 PCAP_API void	pcap_dump_close(pcap_dumper_t *);
 PCAP_API void	pcap_dump(u_char *, const struct pcap_pkthdr *, const u_char *);
+PCAP_API int	pcap_dump_split(u_char *, const struct pcap_pkthdr *, const u_char *, bpf_u_int32,
+	const u_char *, bpf_u_int32);
 
 PCAP_API int	pcap_findalldevs(pcap_if_t **, char *);
 PCAP_API void	pcap_freealldevs(pcap_if_t *);
--- a/sf-pcap.c
+++ b/sf-pcap.c
@@ -761,6 +761,54 @@ pcap_dump(u_char *user, const struct pcap_pkthdr *h, const u_char *sp)
 	(void)fwrite(sp, h->caplen, 1, f);
 }
 
+/*
+ * Output a packet in two parts to the initialized dump file.
+ */
+int
+pcap_dump_split(u_char *user, struct pcap_pkthdr *h, const u_char *sp1, bpf_u_int32 sp1_len,
+	const u_char *sp2, bpf_u_int32 sp2_len)
+{
+	register File *f;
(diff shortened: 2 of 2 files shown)


infrastation:
Would it be simpler to join the parts of the packet outside of libpcap instead?

guyharris:
And if avoiding the copy involved in joining the parts of the packet is important for performance reasons (cf. `writev()` in UN\*X and `WriteFileGather()` on Windows), consider the [zero-one-infinity rule](https://en.wikipedia.org/wiki/Zero_one_infinity_rule).

scottfgjordan:
Hi all, in the use case I described on the pull request description (separate header and payload), this split dumping was used as a way around joining them outside of the pcap to avoid a performance hit. Point taken for the zero-one-infinity rule. I've found this approach useful for my purposes so with this pull request I just wanted see if there was any appetite for making this functionality a part of the libpcap PCAP API.

guyharris:
> Hi all, in the use case I described on the pull request description (separate header and payload), this split dumping was used as a way to avoid joining the parts to avoid a performance hit.

I.e., you measured a significant performance loss due to copying the header and payload into a buffer to hand to `pcap_dump()`?  (Presumably you weren't allocating the buffer and freeing it for each packet.)


> Point taken for the zero-one-infinity rule. I've found this approach useful for my purposes so with this pull request I mainly wanted see if there was any appetite for making this functionality a part of the libpcap PCAP API.

If there's a significant performance hit that can't be overcome in any other way, we might want to offer a *general* gather-write version of `pcap_dump()`, taking a pointer to (the first element of) an array of pointers and sizes (like a `struct iovec`) and a structure element count, such as `pcap_gather_dump()`.

scottfgjordan:
I'd have to get some performance measurements to be certain as it was something that came up in a review stage. A more general gather-write version of pcap_dump() makes sense to me. I can work on putting that together if you're interested.

guyharris:
> I'd have to get some performance measurements to be certain as it was something that came up in a review stage. A more general gather-write version of pcap_dump() makes sense to me. I can work on putting that together if you're interested.

Step 1 is to see whether there's a performance issue that can't be fixed in some other fashion.

scottfgjordan:
Fair enough, would you like me to close down this pull request for now?

mcr:
Yes, at the least, I'd rather have a vector of "const u_char *sp1" rather than two.
Alternatively, leave the writing to the application, and instead have a pcap_dump_header() that will setup the struct pcap_sf_pkthdr sf_hdr into a buffer.  The application can then use writev() or something to send things.  Third possibility, pcap_writev() which will then initiate the writev() in an appropriate OS independent way, but then someone will wand async I/O.

scottfgjordan:
Thanks for all the helpful feedback! I'm going to close this down for now and will reach out again if need be.

PR #979, opened 2020-11-19
fuzz: asserts cap_len is always lesser than length

catenacyber:
It is my understanding that `cap_len` should always be lesser than `len`
cf https://stackoverflow.com/questions/1491660/pcap-struct-pcap-pkthdr-len-vs-caplen
If so, the PR enhances one fuzz target to add this check.

I have a pcap with len=0 and caplen=0x63503d00

### DIFF
--- a/testprogs/fuzz/fuzz_pcap.c
+++ b/testprogs/fuzz/fuzz_pcap.c
@@ -2,6 +2,7 @@
 #include <stdlib.h>
 #include <fcntl.h>
 #include <errno.h>
+#include <assert.h>
 
 #include <pcap/pcap.h>
 
@@ -67,6 +68,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
     r = pcap_next_ex(pkts, &header, &pkt);
     while (r > 0) {
         //TODO pcap_offline_filter
+        assert(header.caplen <= header.len);
         fprintf(outfile, "packet length=%d/%d\n",header->caplen, header->len);
         r = pcap_next_ex(pkts, &header, &pkt);
     }


guyharris:
To quote MSVC:

    C:\projects\libpcap\testprogs\fuzz\fuzz_pcap.c:71:22: error: 'header' is a pointer; did you mean to use '->'?
             assert(header.caplen <= header.len);

catenacyber:
Oops thanks Guy

guyharris:
@fxlb: I think you added checks for caplen > len, but was that in libpcap or tcpdump?

Should we treat caplen > len as an error and fail, should we just increase len to be = caplen, or should we leave that to applications using libpcap?

fxlb:
> @fxlb: I think you added checks for caplen > len, but was that in libpcap or tcpdump?

It was in tcpdump: https://github.com/the-tcpdump-group/tcpdump/blob/a38f788edf0df65f63c84241a4ac5f372c86c7dd/print.c#L348:L355

> Should we treat caplen > len as an error and fail, should we just increase len to be = caplen, or should we leave that to applications using libpcap?

I need a little time to think about that.

catenacyber:
In my case, I think this is rather error and fail (or leave it to the applications using libpcap) than increase len to caplen=0x63503d00 on my 2kb pcap

fxlb:
> In my case, I think this is rather error and fail (or leave it to the applications using libpcap) than increase len to caplen=0x63503d00 on my 2kb pcap

Could you add this pcap file for a test (.gz or .zip) ?

fxlb:
> Could you add this pcap file for a test (.gz or .zip) ?

I mean in this converstion.

catenacyber:
https://catenacyber.fr/fail.pcap

fxlb:
> https://catenacyber.fr/fail.pcap

$ ls -lh fail.pcap 
-rw-r--r-- 1 xxx xxxx 457K Nov 24 15:26 fail.pcap

"my 2kb pcap"?

catenacyber:
Yes it is not 2kb :-p I was thinking about another one

catenacyber:
But it is still less than 0x63503d00 bytes

fxlb:
```
$ ./tcpdump -#nq -r fail.pcap 
reading from file fail.pcap, link-type EN10MB (Ethernet), snapshot length 262144
    1  15:37:47.040000 IP 192.168.202.115.3049 > 192.168.24.100.445: tcp 0
    2  15:37:47.040000 IP 192.168.24.100.445 > 192.168.202.115.3049: tcp 0
    3  15:37:47.040000 IP 192.168.202.115.3049 > 192.168.24.100.445: tcp 0
    4  15:37:47.040000 IP 192.168.202.115.3049 > 192.168.24.100.445: tcp 137
    5  15:37:47.040000 IP 192.168.24.100.445 > 192.168.202.115.3049: tcp 89
    6  15:37:47.210000 IP 192.168.202.115.3049 > 192.168.24.100.445: tcp 0
    7  15:37:49.290000 IP 163.163.163.163 > 163.163.129.64: ip-proto-163
    8  [Invalid header: caplen==0, len(1666202880) > 262144]
    9  [Invalid header: len(3779827712) > 262144]
   10  01:00:00.032768 00:00:00:00:61:00 Unknown SSAP 0x56 > 00:00:00:00:00:02 Null Information, send seq 0, rcv seq 36, Flags [Final], length 92
tcpdump: pcap_loop: invalid packet capture length 2450261632, bigger than snaplen of 262144
```
> I have a pcap with len=0 and caplen=0x63503d00

Ok, in this pcap file there is one pkt with len=1666202880 (=0x63503D00),[ not caplen] and caplen=0 [not len].
And no pkt with `len < caplen`.
Have you another pcap file for this case?

catenacyber:
Sorry my mistake...

No there pcaps.
But I guess we should let the fuzzer try to find one with this PR

PR #982, opened 2020-12-19
Raybellis gzip v2

mcr:
This is a rebase of Ray Bellis' gzip interface for pcap.

### DIFF
--- a/configure
+++ b/configure
@@ -719,6 +719,7 @@ infodir
 docdir
 oldincludedir
 includedir
+runstatedir
 localstatedir
 sharedstatedir
 sysconfdir
@@ -747,6 +748,7 @@ enable_protochain
 with_pcap
 with_libnl
 enable_ipv6
+enable_ioplugin
 with_dag
 with_dag_includes
 with_dag_libraries
@@ -819,6 +821,7 @@ datadir='${datarootdir}'
 sysconfdir='${prefix}/etc'
 sharedstatedir='${prefix}/com'
 localstatedir='${prefix}/var'
+runstatedir='${localstatedir}/run'
 includedir='${prefix}/include'
 oldincludedir='/usr/include'
 docdir='${datarootdir}/doc/${PACKAGE_TARNAME}'
@@ -1071,6 +1074,15 @@ do
   | -silent | --silent | --silen | --sile | --sil)
     silent=yes ;;
 
+  -runstatedir | --runstatedir | --runstatedi | --runstated \
+  | --runstate | --runstat | --runsta | --runst | --runs \
+  | --run | --ru | --r)
+    ac_prev=runstatedir ;;
+  -runstatedir=* | --runstatedir=* | --runstatedi=* | --runstated=* \
+  | --runstate=* | --runstat=* | --runsta=* | --runst=* | --runs=* \
+  | --run=* | --ru=* | --r=*)
+    runstatedir=$ac_optarg ;;
+
   -sbindir | --sbindir | --sbindi | --sbind | --sbin | --sbi | --sb)
     ac_prev=sbindir ;;
   -sbindir=* | --sbindir=* | --sbindi=* | --sbind=* | --sbin=* \
@@ -1208,7 +1220,7 @@ fi
 for ac_var in	exec_prefix prefix bindir sbindir libexecdir datarootdir \
 		datadir sysconfdir sharedstatedir localstatedir includedir \
 		oldincludedir docdir infodir htmldir dvidir pdfdir psdir \
-		libdir localedir mandir
+		libdir localedir mandir runstatedir
 do
   eval ac_val=\$$ac_var
   # Remove trailing slashes.
@@ -1361,6 +1373,7 @@ Fine tuning of the installation directories:
   --sysconfdir=DIR        read-only single-machine data [PREFIX/etc]
   --sharedstatedir=DIR    modifiable architecture-independent data [PREFIX/com]
   --localstatedir=DIR     modifiable single-machine data [PREFIX/var]
+  --runstatedir=DIR       modifiable per-process data [LOCALSTATEDIR/run]
   --libdir=DIR            object code libraries [EPREFIX/lib]
   --includedir=DIR        C header files [PREFIX/include]
   --oldincludedir=DIR     C header files for non-gcc [/usr/include]
@@ -1398,6 +1411,8 @@ Optional Features:
   --disable-largefile     omit support for large files
   --disable-protochain    disable \"protochain\" insn
   --enable-ipv6           build IPv6-capable version [default=yes]
+  --enable-ioplugin       build support for loading I/O plugins that can do
+                          compression, etc. [default=no]
   --enable-remote         enable remote packet capture [default=no]
   --disable-remote        disable remote packet capture
   --enable-optimizer-dbg  build optimizer debugging code
@@ -4528,7 +4543,7 @@ else
     We can't simply define LARGE_OFF_T to be 9223372036854775807,
     since some C++ compilers masquerading as C compilers
     incorrectly reject 9223372036854775807.  */
-#define LARGE_OFF_T (((off_t) 1 << 62) - 1 + ((off_t) 1 << 62))
+#define LARGE_OFF_T ((((off_t) 1 << 31) << 31) - 1 + (((off_t) 1 << 31) << 31))
   int off_t_is_large[(LARGE_OFF_T % 2147483629 == 721
 		       && LARGE_OFF_T % 2147483647 == 1)
 		      ? 1 : -1];
@@ -4574,7 +4589,7 @@ else
     We can't simply define LARGE_OFF_T to be 9223372036854775807,
     since some C++ compilers masquerading as C compilers
     incorrectly reject 9223372036854775807.  */
-#define LARGE_OFF_T (((off_t) 1 << 62) - 1 + ((off_t) 1 << 62))
+#define LARGE_OFF_T ((((off_t) 1 << 31) << 31) - 1 + (((off_t) 1 << 31) << 31))
   int off_t_is_large[(LARGE_OFF_T % 2147483629 == 721
 		       && LARGE_OFF_T % 2147483647 == 1)
 		      ? 1 : -1];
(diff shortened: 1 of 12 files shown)


pcap-ioplugin.c
```diff
@@ -0,0 +1,259 @@
...
+		    "A null pointer was supplied as the file name");
+		return (NULL);
+	}
+
+        if (fname[0] == '-' && fname[1] == '\0')
+        {
+                fp = stdin;
+                if (stdin == NULL) {
+                        snprintf(errbuf, PCAP_ERRBUF_SIZE,
+                                 "The standard input is not open");
```
gvanem, line 78: This is not a good test for a safe use of `stdin` since it could be a fixed address (like on djgpp).
Perhaps `if (!stdin || ferror(stdin))` is better.

pcap-ioplugin.c
```diff
@@ -0,0 +1,259 @@
...
+}
+
+static FILE*
+stdio_open_write(const char *fname, char *errbuf)
+{
+	FILE *fp = NULL;
+
+	if (strcmp(fname, "-") == 0) {
+		fp = stdout;
+                if (stdout == NULL) {
```
gvanem, line 113: Same issue here.

pcap-ioplugin.c
mcr: Is djgpp still a build target that matters?  (I'm seriously asking: I didn't think so).
I'm happy to do something different here, but I want to make sure that it's really better.
I don't understand what !stdin would do?  That's the same as stdin != NULL, isn't it?

mcr: changed based upon suggestion.

pcap-ioplugin.c
mcr: also updated.

configure
```diff
@@ -4643,7 +4658,7 @@ else
     We can't simply define LARGE_OFF_T to be 9223372036854775807,
     since some C++ compilers masquerading as C compilers
     incorrectly reject 9223372036854775807.  */
-#define LARGE_OFF_T (((off_t) 1 << 62) - 1 + ((off_t) 1 << 62))
+#define LARGE_OFF_T ((((off_t) 1 << 31) << 31) - 1 + (((off_t) 1 << 31) << 31))
```
mcr, line 4661: @guyharris is this an error, or something we should be doing?  
This seems to have jumped into this patch, and I don't know why.

pcap-ioplugin.c
gvanem: > Is djgpp still a build target that matters?

Not really. But there could be other exotic targets were `#define stdin (&__internal_stdin)`.

configure
guyharris: This might just be a case of different versions of autoconf generating different code.

pcap-ioplugin.c
guyharris: On my Solaris 11 VM:

/usr/include/stdio.h includes

    #include <iso/stdio_iso.h>

and /usr/include/iso/stdio_ios.h includes

```
#if defined(__STDC__)
extern __FILE	__iob[_NFILE];
#define	stdin	(&__iob[0])
#define	stdout	(&__iob[1])
#define	stderr	(&__iob[2])
#else
extern __FILE	_iob[_NFILE];
#define	stdin	(&_iob[0])
#define	stdout	(&_iob[1])
#define	stderr	(&_iob[2])
#endif	/* __STDC__ */
```

That goes all the way back to System N, for some Roman-numeral version of N, so other commercial UN\*Xes may do that, too.

(GNU libc doesn't, so few Linux distributions do that; as far as I know, Chris Torek's stdio code doesn't, either, so I don't think any of the BSD-favored UN\*Xes, including macOS, do that.)

testprogs/selpolltest.c
```diff
@@ -74,7 +74,7 @@ main(int argc, char **argv)
 	struct bpf_program fcode;
 	char ebuf[PCAP_ERRBUF_SIZE];
 	pcap_if_t *devlist;
-	int selectable_fd;
+	int selectable_fd = -1;
```
guyharris, line 77: This is presumably done to work around compilers that don't figure out that if neither `doselect` nor `dopoll` are set, `selectable_fd` isn't used.

It belongs in a separate change (which shouldn't require any review, so you might as well just commit it).

sf-pcap.c
```diff
@@ -821,23 +805,19 @@ pcap_dump_open(pcap_t *p, const char *fname)
 		    "A null pointer was supplied as the file name");
 		return NULL;
 	}
-	if (fname[0] == '-' && fname[1] == '\0') {
```
guyharris, line 824: If *all* plugins are supposed to treat "-" as meaning "write to the standard output", `pcap_setup_dump()` should probably do so as well, and report that a write error occurred to "(the) standard output" rather than to "-", as the code currently does.

sf-pcap.c
```diff
@@ -794,6 +777,7 @@ pcap_setup_dump(pcap_t *p, int linktype, FILE *f, const char *fname)
 pcap_dumper_t *
 pcap_dump_open(pcap_t *p, const char *fname)
 {
+	const pcap_ioplugin_t *plugin = pcap_ioplugin_init(getenv("PCAP_IOPLUGIN_WRITE"));
```
guyharris, line 780: Presumably your other pull request will eventually allow a program using libpcap to choose to use different I/O routines, so that, for example, a program could provide a command-line flag, or a GUI checkbox, or something such as that to do compressed output, rather than requiring the user to set an environment variable.

pcap-ioplugin.c
guyharris: > I don't understand what !stdin would do? That's the same as stdin != NULL, isn't it?

Yes.

guyharris:
The use of space indentation vs. tab indentation isn't consistent.

guyharris:
@mcr: Speaking of I/O plugins, would that be something that could be used to do asynchronous I/O, especially when writing a capture file?  I think you have some ideas about doing async I/O when writing.

pcap-ioplugin.c
```diff
@@ -0,0 +1,259 @@
...
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include "ftmacros.h"
+
+#ifdef _WIN32
+#include <pcap-stdinc.h>
```
guyharris, line 33: That file no longer exists; to quote the message in the commit that removed it:

```
commit 787f37eeffdc3c37beeca8357d667ef35880c62b
Author: Guy Harris <gharris@sonic.net>
Date:   Tue Sep 5 13:51:37 2017 -0700
    
    Get rid of pcap-stdinc.h.
    
    On Windows, in each file, include whatever that particular file needs,
    just as we do on UN*X and MS-DOS.
```

guyharris:
I presume there's no guarantee that, for an arbitrary plugin, there will be a `FILE *` corresponding to the input or output stream - or that, even if there is, it will contain the full state of the stream, given that the compression/decompression code might require its own state.

In addition, I'm not sure how control gets handed to the plugin when the sf-pcap.c code attempts to read from or write to the file - or how this handles reading from pcapng files at all (we don't *yet* support writing pcapng files, so that's not an issue, *yet*).

Thus, on the read side, we might want to:

- add a flag to a `struct pcap` to indicate whether it's for a live capture or a savefile;
- put some members of `struct pcap` into a union, with one element of the union being a structure for live captures and another one being a structure for savefiles;
- put the `rfile` member into the union for savefiles, and change it to be a `void *`;
- have `pcap_file()` return NULL for a live capture and return the `rfile` member, cast to `FILE *` for a savefile;
- deprecate `pcap_file()`, as, in the general case, all we can guarantee is that it'll return NULL for a live capture and something non-NULL, but *not* guaranteed to be usable with standard I/O routines, for a savefile;
- add `pcap_is_savefile()`, which returns 0 for live captures and 1 for savefiles (meaning "return the "this is a savefile" flag mentioned above;
- have read plugins have a read routine as well as an open-for-reading routine, and write plugins have a write routine as well as an open-for-writing routine, with the read and write routines taking `fread()` and `fwrite()`-style arguments, but with the `void *` passed, rather than a `FILE *`, as the handle argument, and with the stdio read and write routines being tiny wrappers around `fread()` and `fwrite()`;
- have a pointer to the read routine, and a pointer to the handle returned by the open-for-reading routine, be members of the savefile structure mentioned above;
- have the sf-pcap.c and sf-pcapng.c code call the read routine, handing it the pointer to the handle, whenever they nee to read data from the savefile.

guyharris:
It looks as if his code depended on the standard I/O routines supporting the caller being able to supply I/O methods to use on the `FILE *`; that works with GNU libc and with BSD-flavored libc's, but not with all UN\*Xes and not on Windows.

I guess we could do step 1, using a `FILE *`, and then later redo it with a more general mechanism that allows arbitrary handles to be associated with the `pcap_t`; that would allow, for example, an asynchronous I/O backend to be provided, but it'd mean that *existing* plugins would have to be changed.

guyharris:
This also needs CMake changes and a `cmakeconfig.h.in` file.

guyharris:
> Thus, on the read side, we might want to:

Unfortunately, there's no data structure on the write side, so we're stuck, for now, with relying on the standard I/O library allowing write routines to be added, unless we make a `pcap_dumper_t` not be a `FILE *`, which might break some code using libpcap that "knows" that it's a `FILE *`.

We should provide new writing routines that don't work that way.

testprogs/selpolltest.c
mcr: fixed on main in fbb5e9ee

mcr:
> @mcr: Speaking of I/O plugins, would that be something that could be used to do asynchronous I/O, especially when writing a capture file? I think you have some ideas about doing async I/O when writing.

Maybe. But, as currently structured in libpcap, it would require making a copy of the buffer.
I have some ideas and some code from a failed project earlier in 2020 on doing it without a copy. 
It will only work if pcap asks the kernel for rather big ring size, but on machines that want to capture everything, that seems a reasonable thing to do.

mcr:
example code at https://github.com/raybellis/libpcap-gzip and 
https://github.com/DNS-OARC/dnscap/blob/develop/src/dumper.c

infrastation:
These changes are a strict subset of pull request #914.

PR #983, opened 2020-12-20
create standard "options" structure to keep ABI options in wraps

mcr:
This would affect:
* pcap_init()
* pcap_activate()
* pcap_open_offline(),
* pcap_dump_open*()

This could also replace some pcap_set_* functions in the future, or obviate the need for new ones.

### DIFF
--- a/pcap-options.c
+++ b/pcap-options.c
@@ -1,0 +1,141 @@
+/*
+ * Copyright (c) 1993, 1994, 1995, 1996, 1998
+ *	The Regents of the University of California.  All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that: (1) source code distributions
+ * retain the above copyright notice and this paragraph in its entirety, (2)
+ * distributions including binary code include the above copyright notice and
+ * this paragraph in its entirety in the documentation or other materials
+ * provided with the distribution, and (3) all advertising materials mentioning
+ * features or use of this software display the following acknowledgement:
+ * ``This product includes software developed by the University of California,
+ * Lawrence Berkeley Laboratory and its contributors.'' Neither the name of
+ * the University nor the names of its contributors may be used to endorse
+ * or promote products derived from this software without specific prior
+ * written permission.
+ * THIS SOFTWARE IS PROVIDED ``AS IS'' AND WITHOUT ANY EXPRESS OR IMPLIED
+ * WARRANTIES, INCLUDING, WITHOUT LIMITATION, THE IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif
+
+#include <sys/types.h>
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "pcap-int.h"
+
+/*
+ * Private data for storing sets of options.
+ * This is to be used by functions which seem to grow additional options
+ * This is avoid ABI explosition of do_thing, and do_thing_with_opt_bar(),
+ * and then do_thing_with_opt_bar_baz(...). Instead a "pcap_option" should be
+ * created to which get/set shall be done.
+ *
+ * each option shall have an element in enum pcap_option_name {}.
+ */
+struct pcap_options {
+        int   tstamp_precision;
+        const char *io_read_plugin;
+        const char *io_write_plugin;
+};
+
+pcap_options *pcap_alloc_option(void)
+{
+        pcap_options *po = malloc(sizeof(struct pcap_options));
+        memset(po, 0, sizeof(struct pcap_options));
+        return po;  // caller has to check for NULL anyway.
+}
+
+void pcap_free_option(pcap_options *po)
+{
+        if(po != NULL) {
+                if(po->io_read_plugin) free((void *)po->io_read_plugin);
+                if(po->io_write_plugin) free((void *)po->io_write_plugin);
+                free((void *)po);
+        }
+}
+
+/* Return 0 on success, -1 on failure invalid option, -2 on type mis-match */
+int pcap_set_option_string(pcap_options *po,
+                           enum   pcap_option_name pon,
+                           const char *value)
+{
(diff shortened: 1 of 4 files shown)


guyharris:
"Would affect" as in "those would be changed in separate commits"?

This is similar to some stuff I've been thinking about, based on some long-ago requests for device-specific options.

The stuff I've been working on gave a name and description for each option; the name is for use on the command line, e.g. with a `--capture-option` flag, e.g. `--capture-option channel=XXX`, and the description would be for use in 1) GUI applications such as Wireshark and 2) a `--show-capture-options` flag that would give a list, for a given device, of the available options, giving the name and the description.

Presumably:

- this would affect `pcap_init()` by having a set of global pcap options to apply (in an argument to `pcap_init()`?;
- this would affect `pcap_activate()` by having a per-device list of options, with calls to set the options being made after `pcap_create()` and before `pcap_activate()`;
- this would affect `pcap_open_offline()` by having something such as a `pcap_open_offline_with_options()` call that takes a list of options as an argument;
- this would affect `pcap_dump_open*()` by having similar `pcap_dump_open*_with_options()` routines.

There would presumably be calls to indicate which options are available for a particular context.

mcr:
> "Would affect" as in "those would be changed in separate commits"?
> 
> This is similar to some stuff I've been thinking about, based on some long-ago requests for device-specific options.

Cool.
Let's take it to the list, if you agree.

> The stuff I've been working on gave a name and description for each option; the name is for use on the command line, e.g. with a `--capture-option` flag, e.g. `--capture-option channel=XXX`, and the description would be for use in 1) GUI applications such as Wireshark and 2) a `--show-capture-options` flag that would give a list, for a given device, of the available options, giving the name and the description.

I think that it could support such use case with some additional calls.
There is perhaps a chicken and egg issue with not knowing which options apply until a device has been selected.
But at least, compile time options in libpcap would become apparent.

> Presumably:
> 
> * this would affect `pcap_init()` by having a set of global pcap options to apply (in an argument to `pcap_init()`?;
> * this would affect 'pcap_activate()`by having a per-device list of options, with calls to set the options being made after`pcap_create()`and before`pcap_activate()`;
> * this would affect `pcap_open_offline()` by having something such as a `pcap_open_offline_with_options()` call that takes a list of options as an argument;
> * this would affect `pcap_dump_open*()` by having similar `pcap_dump_open*_with_options()` routines.

Yes, to all of these.

> There would presumably be calls to indicate which options are available for a particular context.

yes, I think so.
See what I roughed out in this pull request, which is the minimum I think I need to do the ioplugin.
Strings and integers and an ENUM.   Maybe we want to have a call to find out if an enum is valid (i.e. not #ifdef out).

guyharris:
> > "Would affect" as in "those would be changed in separate commits"?
> > This is similar to some stuff I've been thinking about, based on some long-ago requests for device-specific options.
> 
> Cool.
> Let's take it to the list, if you agree.

Sounds good to me.

Does "take it to the list" mean "mention this discussion on the list, and have people comment on GitHub" or "mention the pull request on the list, and have the discussion take place on the list"?

> > The stuff I've been working on gave a name and description for each option; the name is for use on the command line, e.g. with a `--capture-option` flag, e.g. `--capture-option channel=XXX`, and the description would be for use in 1) GUI applications such as Wireshark and 2) a `--show-capture-options` flag that would give a list, for a given device, of the available options, giving the name and the description.

> I think that it could support such use case with some additional calls.
> There is perhaps a chicken and egg issue with not knowing which options apply until a device has been selected.
> But at least, compile time options in libpcap would become apparent.

The options that aren't device-type-dependent would be known at compile time, with known option numbers.  They might not apply to *all* devices - a "monitor mode" option would apply only to Wi-Fi devices - but there would be a way of finding out which options are available.

> See what I roughed out in this pull request, which is the minimum I think I need to do the ioplugin.
> Strings and integers and an ENUM. Maybe we want to have a call to find out if an enum is valid (i.e. not #ifdef out).

We might also want Booleans, and perhaps both signed and unsigned integers.

pcap-options.c
```diff
@@ -0,0 +1,141 @@
...
+        case PON_IO_WRITE_PLUGIN:
+                return po->io_write_plugin;
+        default:
+                return NULL;
+        }
+        return NULL;
+}
+
+
+/* return int value, or zero for not-found, mis-type */
```
guyharris, line 126: You might want to limit the value to 0-to-INT_MAX, and have -1 mean "not found", or have a Boolean success/failure return and return the value through a pointer.

(Many's the time when I wished various programming languages had the notion of a function failing as an explicit concept - "failure *is* an option". :-)  Perhaps that's what exceptions are for, but C doesn't have them.)

mcr: Yes, I was thinking about whether to return the value as a result parameter.
RUST and GO both seem to do failure is an option well.  Rust seems very good at it.
I've been immersed in Rust for 6 weeks.  

I will revise things in that way.  I just wasn't sure if it was worth it.

mcr:
from:https://ci.appveyor.com/project/guyharris/libpcap/builds/36924953/job/lmv1onulm0br98w3
   c:\projects\libpcap\pcap-options.c(122): warning C4702: unreachable code [C:\projects\libpcap\build\pcap_static.vcxproj]

I'm guessing that I need somethinig to call this code?  Or this just about the return at the bottom?

guyharris:
> from:https://ci.appveyor.com/project/guyharris/libpcap/builds/36924953/job/lmv1onulm0br98w3
> c:\projects\libpcap\pcap-options.c(122): warning C4702: unreachable code [C:\projects\libpcap\build\pcap_static.vcxproj]
> 
> I'm guessing that I need somethinig to call this code? Or this just about the return at the bottom?

The default case in the switch statement returns NULL, and all the other cases also return something, so any code after the switch statement will never be executed.

guyharris:
So would there be one namespace of options, with some applying to capture sources, some applying to capture files being read, some applying to capture files being written, and some applying to some or all of those?  In that case, presumably `pcap_init_with_options()` would set default values of those options, which can be overridden with calls to open capture sources, open capture files for input, and capture files for output.

Or would there be separate namespaces of options?

guyharris:
I've fixed the warnings with MSVC and committed this.

PR #995, opened 2021-02-15
Added new DLT_AUERSWALD_LOG and LINKTYPE_AUERSWALD_LOG 

fgw-auerswald:
Here the Pull Request as requested on mailing list.

### DIFF
--- a/pcap-common.c
+++ b/pcap-common.c
@@ -1230,7 +1230,16 @@
 #define LINKTYPE_USB_2_0_FULL_SPEED	294
 #define LINKTYPE_USB_2_0_HIGH_SPEED	295
 
-#define LINKTYPE_MATCHING_MAX	295		/* highest value in the "matching" range */
+/*
+ * Auerswald Logger Protocol
+ * description is provided on
+ * https://github.com/Auerswald-GmbH/auerlog/blob/master/auerlog.txt
+ * requested by Auerswald Developer Team <developer(at)auerswald.de>
+ */
+#define LINKTYPE_AUERSWALD_LOG	296
+
+#define LINKTYPE_MATCHING_MAX	296		/* highest value in the "matching" range */
+
 
 /*
  * The DLT_ and LINKTYPE_ values in the "matching" range should be the
--- a/pcap/dlt.h
+++ b/pcap/dlt.h
@@ -1588,6 +1588,13 @@
 #define DLT_USB_2_0_HIGH_SPEED	295
 
 /*
+ * Auerswald Logger Protocol
+ * description is provided on
+ * https://github.com/Auerswald-GmbH/auerlog/blob/master/auerlog.txt
+ */
+#define DLT_AUERSWALD_LOG	296
+
+/*
  * In case the code that includes this file (directly or indirectly)
  * has also included OS files that happen to define DLT_MATCHING_MAX,
  * with a different value (perhaps because that OS hasn't picked up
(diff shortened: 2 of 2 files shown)


fgw-auerswald:
Which DLT number should we choose, I see we conflicted with another submission.

guyharris:
In what byte order are the numeric fields?  Big-endian, little-endian, or the byte order of the host writing the file?

What are names for message types 0 and 1?

Is `category` just a string?  I assume `procname`, `threadname`, `class_name or source file`, and `method or function name ` are just strings.  To what do those strings correspond?

Is "process ID" a UN*X/Windows process ID?  If so, for which process?

What is `tgid`?

What is `line number`?

What is `sequence number`?

Does `debugtext` run to the end of a type 0 message, with `Null` being a terminating null character?

In a type 1 message, is `sip package` a Session Initiation Protocol message, with `src_ip`, `src_port`, `dst_ip`, and `dst_port` being the IP address and port from which it was sent and the IP address and port to which it was sent?

fgw-auerswald:
I updated the documentation linked in the merge request.
https://github.com/Auerswald-GmbH/auerlog/blob/master/auerlog.txt

We have another release coming up soon. I would appreciate if we could switch to using the proper ID by then.

guyharris:
The type should be 296, not 297, as that's the next available type.

fgw-auerswald:
@guyharris 296 was applied for by FIRA_UCI
https://github.com/the-tcpdump-group/libpcap/pull/1107/files
I was trying to avoid another collision.
So shall I take 296?

guyharris:
> @guyharris 296 was applied for by FIRA_UCI

People can give a number in a pull request, but that doesn't mean they'll get that number.  In the case of that request, they don't have a publicly-available specification for the format, which is delaying the acceptance of the request.

(I've thought about providing a mechanism similar to pcapng's custom block/custom option mechanism to allow an organization with a [Private Enterprise Number](https://en.wikipedia.org/wiki/Private_Enterprise_Number) to manage their own provide space of link-layer header type values, so that 1) the number is permanently assigned and can be used in pcap and pcapng files without fear of collision with other vendors' numbers and 2) they don't need a public spec, because we don't "own" the value.  That would be a way to handle "we want a LINKTYPE_ but we don't want to tell people how it works".)

> So shall I take 296?

Yes.  If, as, and when the people who want a FIRA_UCI tpe make it possible for developers to write code to read or write files with their link-layer header type without having to join their organization etc., they will get assigned whatever the next free type value is - it won't be 296, but they'll just have to deal with that.

fxlb:
And please rebase you PR on top of the-tcpdump-group/libpcap:master.

fxlb:
The simpler way to update the PR to have a readable  history:
```
git reset --hard @^
git reset --hard @^
git rebase master
git apply AUERSWALD_LOG.patch.txt
git push -f
```
[AUERSWALD_LOG.patch.txt](https://github.com/the-tcpdump-group/libpcap/files/9488667/AUERSWALD_LOG.patch.txt)

infrastation:
@fgw-auerswald, now that there is a code point, could you document it [here](https://www.tcpdump.org/linktypes.html) ([source file](https://github.com/the-tcpdump-group/tcpdump-htdocs/blob/master/htmlsrc/linktypes.html), [instructions](https://github.com/the-tcpdump-group/tcpdump-htdocs/blob/master/README.md))?

fgw-auerswald:
@infrastation  Will update documentation tomorrow or the day after, which will also include the link to a preliminary lua dissector for the protocol and hopefully a sample file.

infrastation:
There is now a basic reference for the new type in the table of link-layer types, you are welcome to convert it to a better documented page as discussed.

infrastation:
For posterity, `LINKTYPE_AUERSWALD_LOG` became available in libpcap 1.10.6.

PR #1000, opened 2021-03-05
.cirrus.yml: quote strings in only_if expression

edigaryev:
Hi!

We're going to make some changes to how Cirrus CI parses configurations in the next couple of weeks (there's no press-release ATM) and found some potential issues with your `.cirrus.yml` while doing backwards compatibility checks.

Conditional task execution using`only_if` script should quote strings, similarly to the [examples in the documentation](https://cirrus-ci.org/guide/writing-tasks/#conditional-task-execution), otherwise the expression engine might treat these characters as operators.

This change makes sure the transition goes smoothly.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -4,7 +4,7 @@ env:
 
 freebsd_task:
   name: FBSD+BM
-  only_if: $CIRRUS_BRANCH != coverity_scan
+  only_if: $CIRRUS_BRANCH != 'coverity_scan'
   freebsd_instance:
     matrix:
       - image_family: freebsd-11-4
@@ -28,7 +28,7 @@ freebsd_task:
 
 linux_task:
   name: LNX+BM
-  only_if: $CIRRUS_BRANCH != coverity_scan
+  only_if: $CIRRUS_BRANCH != 'coverity_scan'
   container:
     image: ubuntu:20.04
     cpu: 2
@@ -46,7 +46,7 @@ linux_task:
 
 macos_task:
   name: MAC+BM
-  only_if: $CIRRUS_BRANCH != coverity_scan
+  only_if: $CIRRUS_BRANCH != 'coverity_scan'
   macos_instance:
     image: big-sur-xcode
   env:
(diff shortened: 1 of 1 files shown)


fxlb:
Thank you!

infrastation:
One thousand pull requests, isn't that a milestone?

@edigaryev, it is not quite clean what problem you are trying to solve in Cirrus CI, but if it is the same problem shell syntax has had for a long time (expansions of all kinds), perhaps it would be better to consider the same solution:
```yaml
only_if: "$CIRRUS_BRANCH" != coverity_scan
```

Beware that rolling a lexical/syntax parser implementation out before the specification is fully sound and polished usually creates more problems than it solves. On a related note, it might be a good time to introduce versions into the Cirrus specification language.

edigaryev:
> One thousand pull requests, isn't that a milestone?

Haha, that was a lucky draw.

> @edigaryev, it is not quite clear what problem you are trying to solve in Cirrus CI, but if it is the same problem shell syntax has had for a long time (expansions of all kinds), perhaps it would be better to consider the same solution:

Thanks for giving this a good thought 👍

We're moving to our recently open-sourced [configuration parser](https://github.com/cirruslabs/cirrus-cli/tree/master/pkg/parser) used in the [Cirrus CLI](https://github.com/cirruslabs/cirrus-cli), which is written in Go and uses [gval](https://github.com/PaesslerAG/gval) to drive `only_if` expressions.

> 
> ```yaml
> only_if: "$CIRRUS_BRANCH" != coverity_scan
> ```

I'm probably totally missing the advantages of this solution, but one immediate problem I see is that the `coverity_scan` should not be allowed to contains spaces, because operator-like symbols (e.g. `-`) in it might be treated as operators, and this will introduce ambiguity.

Why not `$CIRRUS_BRANCH != "coverity_scan"` instead?

> Beware that rolling a lexical/syntax parser implementation out before the specification is fully sound and polished usually creates more problems than it solves. On a related note, it might be a good time to introduce versions into the Cirrus specification language.

Since the syntax was already informally defined a long time ago, we're merely trying to keep things backwards compatible and this PR is one of the cases where it was easier to fix the potential errors in the "consumers", since there are not many cases observed in the wild.

PR #1005, opened 2021-04-12
Fix symlinking with DESTDIR

orgads:
`make install DESTDIR=/path/inst` failed on symlink.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2504,14 +2504,14 @@ function(install_manpage_symlink SOURCE TARGET MANDIR)
     endif(MINGW)
 
     install(CODE
-        "message(STATUS \"Symlinking: ${CMAKE_INSTALL_PREFIX}/${MANDIR}/${SOURCE} to ${TARGET}\")
+        "message(STATUS \"Symlinking: \$ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/${MANDIR}/${SOURCE} to ${TARGET}\")
          execute_process(
             COMMAND \"${CMAKE_COMMAND}\" \"-E\" \"remove\" \"${TARGET}\"
-            WORKING_DIRECTORY ${CMAKE_INSTALL_PREFIX}/${MANDIR}
+            WORKING_DIRECTORY \$ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/${MANDIR}
           )
          execute_process(
             COMMAND ${LINK_COMMAND}
-            WORKING_DIRECTORY ${CMAKE_INSTALL_PREFIX}/${MANDIR}
+            WORKING_DIRECTORY \$ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/${MANDIR}
             RESULT_VARIABLE EXIT_STATUS
           )
           if(NOT EXIT_STATUS EQUAL 0)


guyharris:
1. In which version of CMake was `ENV()` introduced?  We'd need to make the maximum of that and the current required versions the required versions.
2. Does this work if DESTDIR *isn't* in the environment?

orgads:
1. The earliest version I found is 2.4.0. Commit Kitware/CMake@0fa30e1a6.
2. Yes. If you pass it in env var or command-line, make passes it on as an env var (with a trailing slash). If it's empty, it's just empty, and install works as intended.

guyharris:
> es. If you pass it in env var or command-line, make passes it on as an env var (with a trailing slash).

Presumably CMake appends the trailing slash, because Make most definitely doesn't:

```
$ cat Makefile
test:
	echo ${DESTDIR}
$ DESTDIR=foo make
echo foo
foo
```

orgads:
Ah, sorry. My mistake. `DESTDIR` doesn't have a trailing slash. `CMAKE_INSTALL_PREFIX` has a *leading* slash. So if prefix is `/usr` and `DESTDIR` is `/my/path`, the result is `/my/path/usr`.

guyharris:
But the first argument to `install_manpage_symlink()`, `SOURCE`, is just the file name of the man page, not a path name, so it has no leading slash.

Furthermore, the symlinks created by `install_manpage_symlink()` should *not* contain a slash - they just exist so that name A in directory D refers to the same file as name B, so it's just a symlink from A to B in D, not a symlink to D/A in D - so the destination directory should *not* be prepended to `SOURCE`, it should be prepended to `MANDIR` - which is what the later changes, to the `execute_process()` calls, do.

orgads:
Ok, I'll check it tomorrow.

guyharris:
> Ok, I'll check it tomorrow.

I checked it today; it seems to work if the only place where `DESTDIR` is prepended is in the `execute_process()` calls, not in `LINK_COMMAND`.

Note that uninstalling requires that you set `DESTDIR`.

CMakeLists.txt
```diff
@@ -2498,20 +2498,20 @@ function(install_manpage_symlink SOURCE TARGET MANDIR)
...
         # we've found that executable.
-        set(LINK_COMMAND "\"${LINK_EXECUTABLE}\" \"-s\" \"${SOURCE}\" \"${TARGET}\"")
+        set(LINK_COMMAND "\"${LINK_EXECUTABLE}\" \"-s\" \$ENV{DESTDIR}\"${SOURCE}\" \"${TARGET}\"")
     else(MINGW)
-        set(LINK_COMMAND "\"${CMAKE_COMMAND}\" \"-E\" \"create_symlink\" \"${SOURCE}\" \"${TARGET}\"")
+        set(LINK_COMMAND "\"${CMAKE_COMMAND}\" \"-E\" \"create_symlink\" \$ENV{DESTDIR}\"${SOURCE}\" \"${TARGET}\"")
     endif(MINGW)
 
     install(CODE
         "message(STATUS \"Symlinking: ${CMAKE_INSTALL_PREFIX}/${MANDIR}/${SOURCE} to ${TARGET}\")
```
guyharris, line 2507: You should prepend it to `CMAKE_INSTALL_PREFIX` here as well, so that the progress messages for adding symlinks are correct.

orgads: Done

PR #1010, opened 2021-04-29
Fix for Npcap Issue #303

joelg989:
https://github.com/nmap/npcap/issues/303

MS Surface LTE Mobile Network Adaptors generate a nonstandard error code when attempting to set non-promiscuous mode.  Given that the interface can only ever operate in non-promiscuous mode, this operation can't fail, so the error should be ignored.  Have tested on the affected hardware and the fix results in packet capture functioning as normal.

Submitting to mainline libpcap as recommended in that issue.


guyharris:
So what happens if you *do* try to capture in promiscuous mode?  Does it report the same error?

joelg989:
You get the same error, but I decided not to suppress it as in that case it's a real error; the device won't capture in promiscuous mode.

PR #1018, opened 2021-05-30
cmake: fix Finddpkd comparison between strings

clementperon:
This is not a real issue, just proper to not have the PATH included 2 times.


guyharris:
> This is not a real issue

Well, it's not a real issue in the sense that I guess it doesn't *break* anything, but using numeric comparison on non-numbers is Just Wrong, so it should be fixed.

PR #1020, opened 2021-06-04
add loongarch ISA_64_BIT

loongson-zjl:
add new arch loongarch .

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -3881,6 +3881,8 @@ init_tpacket(pcap_t *handle, int version, const char *version_str)
 #define ISA_64_BIT	"s390x"
 #elif defined(__mips__)
 #define ISA_64_BIT	"mips64"
+#elif defined(__loongarch__)
+#define ISA_64_BIT     "loongarch64"
 #elif defined(__hppa__)
 #define ISA_64_BIT	"parisc64"
 #endif


guyharris:
That code was removed in 1.10.0, as it applies only if you're running on a kernel that only supports TPACKET_V1; TPACKET_V2 was added in 2.6.27, so this change would apply only to libpcap 1.8.1 or 1.9.1 running on a really old kernel.

We won't be releasing any more 1.8.x or 1.9.x - releases, so this change will *not* end up in any downstream distributions, because it won't show up as a 1.8.2 or a 1.9.2 release from tcpdump.org.  You'd have to request that the downstream distributions apply it to any releases that use libpcap 1.8.1 or 1.9.1.  (And they may not be interested if they use a 2.6.27 or later kernel, as that code won't be used on such a kernel.)

infrastation:
@guyharris, does this mean the current code should work on Loongson without any additional provisions?
@loongson-zjl, do you have a virtual machine to host a Buildbot worker for continuous integration?

guyharris:
> @guyharris, does this mean the current code should work on Loongson without any additional provisions?

Yes.

That whole pile of "get the architecture and see whether it matches the name for any 64-bit flavors of any architecture that has both 32-bit and 64-bit versions" is to handle a botch in TPACKET_V1, wherein the header in the memory-mapped buffer shared between the kernel and userland has an `unsigned long` element in it.

That works fine as long as both kernel and userland are ILP32, because that element is 32 bits in both kernel and userland, or as long as both kernel and userland are LP64, because that element is 64 bits in both kernel and userland.

It does *not* work if the kernel is LP64 and userland is ILP32, as that element is 64 bits in the kernel and *32* bits in userland, so userland *can't* use the normal structure for it - it needs a structure where that field is an unsigned long long, and thus 64 bits, just as it is in the kernel.

(As far as I know, no Linux platform supports an ILP32 kernel and an LP64 userland.  macOS did, for a while, but that's a separate story.)

Unfortunately, the only way I found to allow userland code to determine whether the kernel was ILP32 or LP64 was to query the `machine` member of a structure filled in by a `uname()` call to see whether it corresponded to the name for the 64-bit version of the instruction set for which the code was compiled.

We dropped support for TPACKET_V1 in libpcap 1.10, so that's no longer needed - TPACKET_V2 and later share a data structure with all `__u32` and `__u16` members, so the members are the same sizes for ILP32 and LP64.  Thus, there is no need to add more #defines for any architectures, as we don't bother calling `uname()` to get the architecture.

infrastation:
@guyharris, thank you for the detailed explanation!
@loongson-zjl, I meant a Loongson VM, of course.

PR #1026, opened 2021-06-17
Fix some build issues with MinGW

Biswa96:


rpcapd/CMakeLists.txt
```diff
@@ -136,7 +136,12 @@ if(WIN32 OR ((CMAKE_USE_PTHREADS_INIT OR PTHREADS_FOUND) AND HAVE_CRYPT))
   set(MANFILE_EXPAND rpcapd-config.manfile.in)
 
   if(NOT MSVC)
-    install(TARGETS rpcapd DESTINATION sbin)
+    # DLL and EXE need to be in same bin path in Win32 platform
+    if(MINGW)
+      install(TARGETS rpcapd DESTINATION ${CMAKE_INSTALL_BINDIR})
+    else(MINGW)
+      install(TARGETS rpcapd DESTINATION sbin)
+    endif(MINGW)
```
gvanem, line 144: It would be nice if this old and ugly syntax would be dropped in favour of:
```cmake
if(MINGW)      
  install(TARGETS rpcapd DESTINATION ${CMAKE_INSTALL_BINDIR})    
else()
  install(TARGETS rpcapd DESTINATION sbin)    
endif()
```

But what is `sbin`? I have MinGW (TDM-gcc), but no `sbin` among it's sub-dirs.

Biswa96: I choose that way not to create issue in other environment.

guyharris:
> But what is `sbin`? I have MinGW (TDM-gcc), but no sbin among it's sub-dirs.

It's a UN\*X directory...

...but it's not necessarily the right one for daemons.

I no longer remember the exact rationale for the name; it dates back to a proposal by some people at Sun in the mid-1980's.  That proposal was originally done for SunOS 4.0, with at least part of the rationale being to better support diskless workstations without using Sun's ND protocol for accessing a file as a network disk.  [The Wikipedia page on the (mostly Linux) Filesystem Hierarchy Standard](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) (FHS), a spec influenced by predecessors dating back to the aforementioned proposal, says /sbin and /usr/sbin are for "system binaries", so I guess that's what "sbin" stands for.  (See [somebody's 1992 USENET posting](https://groups.google.com/g/comp.unix.questions/c/g9DsvKQx8h8/m/QNs0F-mHpR4J) for a bit more history.)

The FHS says /usr/sbin is for daemons, but some other systems use /usr/libexec instead.

In any case, with this change, the CMakeLists.txt file uses it for *non*-MSVC, *non*-MinGW environments.

rpcapd/CMakeLists.txt
guyharris: > But what is `sbin`? I have MinGW (TDM-gcc), but no sbin among it's sub-dirs.

It's a UN\*X directory...

...but it's not necessarily the right one for daemons.

I no longer remember the exact rationale for the name; it dates back to a proposal by some people at Sun in the mid-1980's.  That proposal was originally done for SunOS 4.0, with at least part of the rationale being to better support diskless workstations without using Sun's ND protocol for accessing a file as a network disk.  [The Wikipedia page on the (mostly Linux) Filesystem Hierarchy Standard](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) (FHS), a spec influenced by predecessors dating back to the aforementioned proposal, says /sbin and /usr/sbin are for "system binaries", so I guess that's what "sbin" stands for.  (See [somebody's 1992 USENET posting](https://groups.google.com/g/comp.unix.questions/c/g9DsvKQx8h8/m/QNs0F-mHpR4J) for a bit more history.)

The FHS says /usr/sbin is for daemons, but some other systems use /usr/libexec instead.

In any case, with this change, the CMakeLists.txt file uses it for *non*-MSVC, *non*-MinGW environments.

Biswa96:
Done.

Biswa96:
Any progress on this topic?

rpcapd/CMakeLists.txt
kayoub5: You may want to use `${CMAKE_INSTALL_SBINDIR}` instead of `${CMAKE_INSTALL_BINDIR}` and `sbin`

See https://cmake.org/cmake/help/latest/command/install.html#installing-files

Biswa96: As I said in the commit message, DLL and exe need to be in same folder I.e. bin.

mcr (approved):

rpcapd/CMakeLists.txt
guyharris: I've changed that part of rpcapd/CMakeLists.txt to make the sbin-or-not decision based on whether the target is Windows rather than on whether the compiler is MSVC.  You'll need to rebase and update the change.

Biswa96:
Done.

infrastation:
As far as Windows people see it, does the problem stand and does the proposed change solve it?

infrastation:
The proposed change changes a type cast after `DIAG_OFF_NARROWING` to address a type cast warning, which means the diagnostics control simply does not work. There is a comment just before the type cast with the rationale of. So if everything worked as intended, simply there would be no warning, no diagnostics control and no comment. On the other hand, if the type cast solves the problem, it should remove the diagnostics control and the comment. So something is not right.

What exactly are the build system, the steps to reproduce this issue and the warning message?

infrastation:
Without the comments this request will be closed in 7 days.

Biswa96:
My apology 😔 I missed the previous notification. I am using gcc with cmake and ninja. Here is the warning:

```c
libpcap/pcap.c:3488:30: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
 3488 |                 return ((int)(DWORD)p->handle);
      |                              ^
```

guyharris:
Is the "cast from pointer to integer of different size" warning still happening with the current tip of the main branch?

guyharris:
The underlying problem is that `pcap_fileno()` should never have existed on Windows, because the Windows equivalent of a UN\*X file handle is a `HANDLE`, and a `HANDLE` is, apparently, a pointer type, for reasons probably having to do with backwards compatibility with 16-bit Windows or something such as that.

Instead, either there should have been no equivalent, or there should have been an equivalent that returned a `HANDLE` (and possibly returns `INVALID_HANDLE_VALUE` if the `pcap_t` doesn't have a handle corresponding to the device).

[The WinPcap documentation](https://www.winpcap.org/docs/docs_412/html/group__wpcapfunc.html) doesn't mention `pcap_fileno()`, so *perhaps* we could have removed it; however, it's in the WinPcap 4.1.3 .def files and in its pcap/pcap.h, so it's exported by the library, and programs might have used it, so perhaps some programs might have broken if we did.

In any case, it's here, and it's somewhat broken by design.  Perhaps we should deprecate it for Windows; on UN\*X, it's available for programs that only open *known* device types to do ioctls etc. on, but I don't think the `HANDLE` is useful for any purpose on Windows.

guyharris:
> Perhaps we should deprecate it for Windows

We already do.

We might as well just merge this change, and get rid of the DIAG_OFF/DIAG_ON stuff if it's no longer necessary.

Biswa96:
> Is the "cast from pointer to integer of different size" warning still happening with the current tip of the main branch?

Yes, I have verified that also.

PR #1027, opened 2021-06-23
pcap-npf.c: handle NdisMediumIP

fghzxm:
When encountering the NDIS medium type `NdisMediumIP`, treat the interface as an IP interface.

Related to nmap/npcap#173.

### DIFF
--- a/pcap-npf.c
+++ b/pcap-npf.c
@@ -1127,6 +1127,10 @@ pcap_activate_npf(pcap_t *p)
 		p->linktype = DLT_RAW;
 		break;
 
+	case NdisMediumIP:
+		p->linktype = DLT_RAW;
+		break;
+
 	default:
 		/*
 		 * An unknown medium type is assumed to supply Ethernet


guyharris:
I'm not sure all versions of Windows *have* `NdisMediumIP`, and thus it might not be the case that the Windows SDK that comes with older versions of Visual Studio have `NdisMediumIP`.

I'll ask the Npcap folks what the oldest version of Visual Studio that can build Npcap is, and, if it's not VS 2015, adjust the list of VS versions with which we test.

zx2c4:
Or just #define the constant by hand if the SDK doesn't have it? It's not going to change between windows versions or anything.

guyharris:
> Or just #define the constant by hand if the SDK doesn't have it?

Note that it's not a #define, it's a member of an enum, so "the SDK doesn't have it" would have to be determined by a test in CMake.

If we don't have to do that, because we don't have to support older versions of Visual Studio with SDKs that don't have it, I'd rather not complicate things by doing that.

guyharris:
Checked in, with changes to handle older Windows SDKs by checking for the presence of `NdisMediumIP`, as 74be794fe7b3540999149f976e4a829289ad43ef.

guyharris:
OK, the problem is that:

- MinGW defaults to building for Windows Server 2003 (!), which is too old to have even proper IPv6 support;
- we work around that by changing it to build for Windows Vista by setting `_WIN32_WINNT` to 0x0600, to get the IPv6 support;
- that's not sufficient to get us `NdisMediumIP`.

So I've change it to set `_WIN32_WINNT` to 0x0601, instead, and backed out my change (74be794fe7b3540999149f976e4a829289ad43ef).

I'll reopen this and merge it atop my other changes.

fghzxm:
Sorry for the disruption, I've changed the commit message to hopefully be clearer and more coherent in style with existing commits.

guyharris:
> Sorry for the disruption

Not a problem, the preparatory changes are still running through AppVeyor, so updating the pull request doesn't disrupt anything.

guyharris:
Thanks.  Sorry about the mess with MinGW-w64 builds.

PR #1028, opened 2021-06-23
ONLY TEST FOR NOW - build.sh: Build with CFLAGS=-Werror

fxlb:


infrastation:
Last time freebsd-amd64, netbsd-* and openbsd-* had warnings and the rest was clean. Do you think `LIBPCAP_TAINTED` (as in tcpslice now) would be a good way to start guarding known-good builds and to describe the known-bad ones before everything is warning-free?

fxlb:
Closing.

PR #1035, opened 2021-07-19
WIP: Set CANFD_FDF when CAN-FD frame received

kayoub5:
# Context

Linux recently introduced the CANFD_FDF flag, it is a flag not used by Linux kernel itself, however it allows application to use `canfd_frame` struct for both CAN and CAN-FD.

Libpcap is one use case of such applications, since it merges CAN and CAN-FD traffic into one format.

## CAN-FD
* CAN & CAN-FD frames could be physically transmitted on the same CAN-FD bus, on the wire the FDF bit is used to distinguish them.
* Linux socket CAN API assume that application check frame size (MTU), to distinguish between them.
* The tcpdump documentation does not mention how the application can differentiate between CAN and CAN-FD when receiving traffic, and does not specify a fixed frame size, original documentation assume that frame ends when payload ends.
* Linux socket CAN introduced the CANFD_FDF bit recently, that allows applications to mark frames manually, the bit is not used by the kernel itself.
* Keeping unused trailing bytes when storing CAN-FD frames on disk present a waste ((72  -16) / 72 = 77% of frame size), removing them makes it difficult to tell if frame was originally was transmitted as CAN or CAN-FD frame.
* When injecting, Linux only accepts frame sizes of 16 and 72.
#### When receiving (This PR)
See https://github.com/the-tcpdump-group/libpcap/pull/1035
* Library must set the CANFD_FDF for CAN-FD frames received from Linux kernel when receiving bytes (Library checks MTU)
* Library will keep the original frame size as received by the Linux kernel, for backward compatibility, application may strip trailing padding bytes when storing frame to disk.
#### When injecting
Currently Libpcap does not support injection, this will be added as reference for future implementation 
Library must upsize the frame to 72 bytes, when injecting a CAN frame if any of the below is true:
  - Any FD flag is set
  - Payload length is > 8
  - Frame size is > 16

Library must upsize the frame to 16 bytes otherwise.

### DIFF
--- a/pcap/can_socketcan.h
+++ b/pcap/can_socketcan.h
@@ -44,6 +44,8 @@
 #define CAN_MTU     16
 #define CANFD_MTU   72
 
+#define CANFD_BRS   0x01 /* bit rate switch (second bitrate for payload data) */
+#define CANFD_ESI   0x02 /* error state indicator of the transmitting node */
 #define CANFD_FDF   0x04 /* mark CAN FD for dual use of CAN format */
 
 /*
--- a/pcap/can_socketcan.h
+++ b/pcap/can_socketcan.h
@@ -41,6 +41,11 @@
 
 #include <pcap/pcap-inttypes.h>
 
+#define CAN_MTU     16
+#define CANFD_MTU   72
+
+#define CANFD_FDF   0x04 /* mark CAN FD for dual use of CAN format */
+
 /*
  * SocketCAN header, as per Documentation/networking/can.txt in the
  * Linux source.
@@ -48,7 +53,7 @@
 typedef struct {
 	uint32_t can_id;
 	uint8_t payload_length;
-	uint8_t pad;
+	uint8_t fd_flags;
 	uint8_t reserved1;
 	uint8_t reserved2;
 } pcap_can_socketcan_hdr;
--- a/pcap-common.c
+++ b/pcap-common.c
@@ -1409,14 +1409,12 @@ max_snaplen_for_dlt(int dlt)
 static void
 swap_linux_sll_header(const struct pcap_pkthdr *hdr, u_char *buf)
 {
-	u_int caplen = hdr->caplen;
-	u_int length = hdr->len;
+	u_int caplen = min(hdr->len, hdr->caplen);
 	struct sll_header *shdr = (struct sll_header *)buf;
 	uint16_t protocol;
 	pcap_can_socketcan_hdr *chdr;
 
-	if (caplen < (u_int) sizeof(struct sll_header) ||
-	    length < (u_int) sizeof(struct sll_header)) {
+	if (caplen < (u_int) sizeof(struct sll_header)) {
 		/* Not enough data to have the protocol field */
 		return;
 	}
@@ -1428,13 +1426,21 @@ swap_linux_sll_header(const struct pcap_pkthdr *hdr, u_char *buf)
 	/*
 	 * SocketCAN packet; fix up the packet's header.
 	 */
+	caplen -= sizeof(struct sll_header);
 	chdr = (pcap_can_socketcan_hdr *)(buf + sizeof(struct sll_header));
-	if (caplen < (u_int) sizeof(struct sll_header) + sizeof(chdr->can_id) ||
-	    length < (u_int) sizeof(struct sll_header) + sizeof(chdr->can_id)) {
+	if (caplen < sizeof(chdr->can_id)) {
 		/* Not enough data to have the CAN ID */
 		return;
 	}
 	chdr->can_id = SWAPLONG(chdr->can_id);
+	if (caplen < sizeof(chdr->can_id) + sizeof(chdr->payload_length) + sizeof(chdr->fd_flags)) {
+		/* Not enough data to have the CAN-FD flags */
+		return;
+	}
+	// We intentionally test on caplen (MTU) and NOT payload length
+	if (caplen > CAN_MTU || chdr->fd_flags) {
+		chdr->fd_flags |= CANFD_FDF;
+	}
 }
 
 /*


guyharris:
Note that libpcap itself no longer produces DLT_CAN_SOCKETCAN packets; support for them was introduced in

commit f3edbb599b8cbcc7e4560000dcba8e992dc11a31
Author: Akos Vandra <axos88@gmail.com>
Date:   Fri Dec 23 03:29:52 2011 +0200

    Added initial support for canusb devices.

    CanUSB devices are not directly supported by the kernel.
    LibUSB-1.0 is used to communicate them, with a special protocol.

    Reading 2 bytes from interrupt endpoint 1 returns the number of packets in $
    rx and the tx buffer,

    Reading 20 bytes from bulk endpoint 5 returns a 4byte timestamp in 10us
    resolution since powerup, and a can fram in socket_can frame format, so
    DLT_SOCKETCAN may be used.

    TODO: use async libusb communication

and was removed in

commit 93ca5ff7030aaf1219e1de05ec89a68384bfc50b
Author: Guy Harris <gharris@sonic.net>
Date:   Thu Aug 25 13:26:43 2016 -0700
    
    On Linux, handle all CAN captures with pcap-linux.c, in cooked mode.
    
    There's no need to support capturing on PF_CAN sockets; CAN interfaces
    show up as regular interfaces on which you can capture with PF_PACKET
    sockets, so just let pcap-linux.c handle them, and get rid of
    pcap-can-linux.c.

    Capture on them in cooked mode, so we get the protocol field and can
    distinguish between "classic" CAN and CAN FD.
    
    The hardware for which pcap-canusb-linux.c was intended never reached
    production:
    
        https://github.com/axos88/libpcap/commit/f3edbb599b8cbcc7e4560000dcba8e992dc11a31#commitcomment-18716617
    
    so we don't need pcap-canusb-linux.c, either.
    
    This all removes the need for the "host-endian" link-layer header type;
    it never made it into a libpcap release, so we just remove it.  The
    "big-endian" link-layer header type is kept for the benefit of packets
    captured with the old pcap-can-linux.c code; we revert it to its old
    name.

See [this comment on that commit](https://github.com/axos88/libpcap/commit/f3edbb599b8cbcc7e4560000dcba8e992dc11a31#commitcomment-18716617) and the two following comments.

Is there any other packet source that produces `LINKTYPE_CAN_SOCKETCAN` packets?  If not, is there any reason to expect that there will be any old `LINKTYPE_CAN_SOCKETCAN` files out there that have CAN-FD packets?

kayoub5:
> Is there any other packet source that produces `LINKTYPE_CAN_SOCKETCAN` packets?

Yes, the following tools produce PCAPNG files with `LINKTYPE_CAN_SOCKETCAN`
### Open Source Converters that convert files from automotive specific formats to PCAPNG
* https://github.com/Technica-Engineering/Technica.Traces.Tecmp.Converter
* https://github.com/Technica-Engineering/Technica.Traces.Blf.Converter
* https://github.com/Technica-Engineering/Technica.Traces.Asc.Converter
### Commercial product that produce PCAPNG files with `LINKTYPE_CAN_SOCKETCAN` from Vector Informatik devices
* https://technica-engineering.de/en/produkt/andi-automotive-network-diagnoser-ethernet-rest-bus-simulation/

pcap/can_socketcan.h
```diff
@@ -41,14 +41,19 @@
 
 #include <pcap/pcap-inttypes.h>
 
+#define CAN_MTU     16
+#define CANFD_MTU   72
+
+#define CANFD_FDF   0x04 /* mark CAN FD for dual use of CAN format */
```
guyharris, line 47: This file is also for the benefit of programs reading frames, so it should also define `CANFD_BRS` and `CANFD_ESI`.

kayoub5:
@guyharris just to clarify

`LINKTYPE_CAN_SOCKETCAN` was removed from libpcap, and Linux cooked packets can only be used for receiving as far as I know.

So how could a user use libpcap for injecting CAN frames (with `pcap_sendpacket`) ?

kayoub5:
I agree that Linux cooked packets made LINKTYPE_CAN_SOCKETCAN use cases a bit limited on Linux, however there are plenty of devices that still do not support socket can due to platform limitations (windows) or vendor limitations and for those having a link type for CAN do make sense.

One example I can think of is CAN over Serial / SLCAN protocol.

guyharris:
> I agree that Linux cooked packets made LINKTYPE_CAN_SOCKETCAN use cases a bit limited on Linux, however there are plenty of devices that still do not support socket can due to platform limitations (windows) or vendor limitations and for those having a link type for CAN do make sense.

"LINKTYPE_CAN_SOCKETCAN was removed from libpcap" means "the two pcap-XXX.c files that used it, both of which are Linux-only, were removed from libpcap". It doesn't mean "LINKTYPE_CAN_SOCKETCAN and DLT_CAN_SOCKETCAN were completely removed from libpcap, so that they're not even #defined".

For Windows, we never had any CAN support, so it's not as if we've removed the ability to inject CAN packets, or even removed the ability to sniff CAN packets.

For Linux, we never had CAN support for devices that don't support SocketCAN, so, again, for those devices, we never removed the ability to inject CAN packets, or even sniff them, because there was nothing *to* remove.

guyharris:
> "LINKTYPE_CAN_SOCKETCAN was removed from libpcap" means "the two pcap-XXX.c files that used it, both of which are Linux-only, were removed from libpcap". It doesn't mean "LINKTYPE_CAN_SOCKETCAN and DLT_CAN_SOCKETCAN were completely removed from libpcap, so that they're not even #defined".

And, in fact, the current pcap-linux.c in the main and 1.10 branches uses `DLT_CAN_SOCKETCAN` for devices with `ARPHRD_CAN`; the current versions of Wireshark 3.4 and 3.6 support that, complete with the FDF flag, and tcpdump doesn't currently parse CAN packets (whether they're in `DLT_CAN_SOCKETCAN`, `DLT_LINUX_SLL ` in an "any" capture, or `DLT_LINUX_SLL2` in an "any" capture) and, if it ever does, it'll handle them in `DLT_CAN_SOCKETCAN` and rely on the FDF flag to distinguish between CAN and CAN FD.

For `DLT_LINUX_SLL` and `DLT_LINUX_SLL2`, the protocol is available in the SLL/SLL2 header to indicate whether the frame is a CAN or CAN FD frame.

So:

1. Are there any cases where an existing capture file will use `DLT_CAN_SOCKETCAN` and contain CAN FD frames that *don't* have the FDF flag set, so that code reading the file will have to detect CAN FD frames by looking at the length?
2. If so, should libpcap - *and* Wireshark's libwiretap - set the FDF flag when reading `DLT_CAN_SOCKETCAN` files, based on the length, or should that be left up to dissectors?

kayoub5:
> Are there any cases where an existing capture file will use DLT_CAN_SOCKETCAN and contain CAN FD frames that don't have the FDF flag set, so that code reading the file will have to detect CAN FD frames by looking at the length?

I don't know, I can't speak for the entire community, but I do know that the community of DLT_CAN_SOCKETCAN users is not that large.

> If so, should libpcap - and Wireshark's libwiretap - set the FDF flag when reading DLT_CAN_SOCKETCAN files, based on the length, or should that be left up to dissectors?

Dissectors could implement logic as below "can.fd = can.fd_flags & 0x5 > 0 or len(frame) > 16"

kayoub5:
Closing in favor of #1052 and https://github.com/the-tcpdump-group/libpcap/commit/e390976cddf94f659b86cf4830ce593199379aad

PR #1037, opened 2021-07-21
fix unaligned access to structs in rpcapd

lordrasmus:
fix unaligned access to structs in daemon_msg_findallif_req() and daemon_seraddr()

if an interface has, for example a 3 char name ( unm ), the pointer for sockaddrout_ipv4
gets an uneven address like 0x1003

an processors with alignment restrictions like arm and mips
the write to sockaddrout_ipv4->family does not write to 0x1003 but to 0x1002

same problem for findalldevs_if

so an local variable is defined which will have the correct alignment and then
the variable is copied to the destination address with memcpy

### DIFF
--- a/rpcapd/daemon.c
+++ b/rpcapd/daemon.c
@@ -1703,7 +1703,10 @@ daemon_msg_findallif_req(uint8 ver, struct daemon_slpars *pars, uint32 plen)
 	{
 		uint16 lname, ldescr;
 
-		findalldevs_if = (struct rpcap_findalldevs_if *) &sendbuf[sendbufidx];
+		struct rpcap_findalldevs_if dev_if_tmp;
+		char*  dev_if_p = &sendbuf[sendbufidx];
+		
+		findalldevs_if = (struct rpcap_findalldevs_if *) &dev_if_tmp;
 
 		if (sock_bufferize(NULL, sizeof(struct rpcap_findalldevs_if), NULL,
 		    &sendbufidx, RPCAP_NETBUF_SIZE, SOCKBUF_CHECKONLY, errmsgbuf, PCAP_ERRBUF_SIZE) == -1)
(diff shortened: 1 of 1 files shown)


lordrasmus:
maybe the warning for that should activated by default when using gcc 

here are the options

-Wcast-align

> Warn whenever a pointer is cast such that the required alignment of the target is increased. For example, warn if a char * is cast to an int * on machines where integers can only be accessed at two- or four-byte boundaries.

-Wcast-align=strict

> Warn whenever a pointer is cast such that the required alignment of the target is increased. For example, warn if a char * is cast to an int * regardless of the target machine.

guyharris:
> if an interface has, for example a 3 char name ( unm ), the pointer for sockaddrout_ipv4
gets an uneven address like 0x1003

Sigh. Maybe that's something to clean up in a future protocol revision.

> an processors with alignment restrictions like arm and mips
the write to sockaddrout_ipv4->family does not write to 0x1003 but to 0x1002

And on processors with alignment restrictions like SPARC, the write to sockaddrout_ipv4->family does not write to 0x1003 but crashes.

infrastation:
I presumed all OSes running on MIPS would automatically abort processes on unaligned access, but it turns out that's not the case. I just ran a little C program on SPARC to confirm it gets a `SIGILL` there, and then tested it on the three MIPS Buildbot workers. On Linux and NetBSD the program does not get a signal, although I didn't check if memory access resulted in consistent data. On OpenBSD it gets a `SIGBUS`, so if you want to add a test to guard some particular code, you can use that already.

On ARM, as far as I gather it, unaligned access only applies to 32-bit and the only 32-bit ARM worker (Linux) seems to have a knob to enable `SIGBUS` on unaligned access, but the knob does not seem to have any effect on the behaviour (the program does not receive the signal and there is no log message). This might have to do with the fact that the hardware is actually AArch64.

guyharris:
The client side (pcap-rpcap.c) doesn't appear to have this problem, as it doesn't read the entire response into one big buffer, it reads it an item at a time, so that the addresses are copied into an on-stack structure that's properly aligned.

lordrasmus:
i see you changed the  code a bit and removed the extra pointer and added some comments

i thougt about doing that too but i didn't wanted to change to much

PR #1039, opened 2021-07-28
IEEE 802.11: add support to filter Block Ack Req & Block Ack control frames

gokulkumar792:
As per the IEEE 802.11-2016 std, section 9.2.4.1.3, Table 9-1, the control frame Block Ack Request has the subtype 0b1000 and frame Block Ack has the subtype 0b1001. Add this two control frame subtypes in ieee80211.h so that the 802.11 frames can be filtered using "bar" or "ba" in addition to the other ctl frame subtypes.


gokulkumar792:
Hi, please let me know, if there is anything that I could do to help review & merge this PR. Thanks.

Issue #1059, opened 2021-11-02
Description of pcap_next_ex return value is not complete


PeiweiHu:
The document says pcap_next_ex returns 1 if the packet was read without problems. However, it seems should be > 0. It's better to correct the document if it does.

guyharris:
> it seems should be > 0

Under what circumstances have you found that it returns a value > 1?

PeiweiHu:
> > it seems should be > 0
> 
> Under what circumstances have you found that it returns a value > 1?

Sorry, I looked through pcap_offline_read and thought it will return a number representing packet count, but forgot that only cnt=1 is passed in pcap_next_ex. I will be more careful next time. Sorry again.

PR #1060, opened 2021-11-02
Correct return value comments in pcap_next_ex

ThomasHabets:
The comments in the code say that it's `>0`, not `1`.

### DIFF
--- a/pcap.c
+++ b/pcap.c
@@ -622,7 +622,7 @@ pcap_next_ex(pcap_t *p, struct pcap_pkthdr **pkt_header,
 		 * Return codes for pcap_offline_read() are:
 		 *   -  0: EOF
 		 *   - -1: error
-		 *   - >1: OK
+		 *   - >0: OK (can only be 1 since we pass in 1)
 		 * The first one ('0') conflicts with the return code of
 		 * 0 from pcap_read() meaning "no packets arrived before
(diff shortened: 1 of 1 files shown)


ThomasHabets:
I see there's bug #1059

guyharris:
Under what circumstances have you found that it returns a value > 1?

ThomasHabets:
Hmm... maybe it doesn't. Maybe it's actually the comment in the code that's wrong:

```
/*
	 * Return codes for pcap_read() are:
	 *   -  0: timeout
	 *   - -1: error
	 *   - -2: loop was broken out of with pcap_breakloop()
	 *   - >1: OK
	 * The first one ('0') conflicts with the return code of 0 from
	 * pcap_offline_read() meaning "end of file".
	*/
```

Twice in that function comments says `>1` is `OK`, but actually `>0 and in this case always 1`. Still, I misread it.

The comments should be fixed though, to at least say `>0`, not `>1`. I can update this PR when I get to my coding laptop.

guyharris:
2aaaef7d3a802d4379d7e994d16ac428e5faefe5 fixes the comments *and* explains, in the fixed comments and the commit message, why, in *this particular* case, "> 0" means "= 1".

ThomasHabets:
Awesome. Thank you.

And apoligies for my misunderstanding.

PR #1067, opened 2021-11-16
NPF: improve error handling in pcap_create_interface()

dmiller-nmap:
pcap_close() cannot be used at this point (in pcap_create()) since it
calls cleanup_op, which is not set until pcap_activate(). Additionally,
return value of strdup was never checked, and several error conditions
resulted in failing to call PacketCloseAdapter on the temporary adapter.

### DIFF
--- a/pcap-npf.c
+++ b/pcap-npf.c
@@ -1516,12 +1516,12 @@ pcap_create_interface(const char *device _U_, char *ebuf)
 {
 	pcap_t *p;
 #ifdef HAVE_PACKET_GET_TIMESTAMP_MODES
-	char *device_copy;
-	ADAPTER *adapter;
+	char *device_copy = NULL;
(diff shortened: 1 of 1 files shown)


dmiller-nmap:
My analysis regarding cleanup_op was incorrect; it is set to `pcap_cleanup_live_common` in `pcap_create_common`, which is called in the `PCAP_CREATE_COMMON` macro at the beginning of `pcap_create_interface()`. The changes are still good, but my comments might be misleading.

guyharris:
> The changes are still good, but my comments might be misleading.

I'll look at them and fix them if they need it.

guyharris:
I've cleaned the code a bit to move the fetch-the-timestamp-types stuff to a routine of its own, matching the way we do it in pcap-linux.c.  I also put back the `pcap_close()` call; as you indicate, it's safe in this context.

PR #1068, opened 2021-11-22
Haiku: implement pcap_lib_version

davidkaroly:
Since the introduction of commit #f290068 each pcap-xx file provides its own pcap_lib_version() implementation.

This pull request adds pcap_lib_version() to pcap-haiku as it was missing before, leading to library loading issues on Haiku.


infrastation:
Thank you for suggesting this fix. Would you like to host a Haiku VM for Buildbot worker to plug into ci.tcpdump.org? This way it would be easier to notice when something breaks.

davidkaroly:
> Would you like to host a Haiku VM for Buildbot worker to plug into ci.tcpdump.org? This way it would be easier to notice when something breaks.

The suggestion makes sense, but I just recently started contributing to Haiku so probably I'm not the right person for this.

PR #1075, opened 2021-12-14
Add support for Realtek (Ethertype) DSA data link types

luizluca:


infrastation:
Please see [this page](https://www.tcpdump.org/linktypes.html) and produce a sufficiently detailed description of the new header types.

guyharris:
See the entries for ``LINKTYPE_DSA_TAG_BRCM`, `LINKTYPE_DSA_TAG_BRCM_PREPEND`, `LINKTYPE_DSA_TAG_DSA`, and `LINKTYPE_DSA_TAG_EDSA` on the page mentioned by @infrastation as examples of how to give a description of DSA data link types.

Note that those descriptions are all very similar on that page, *but* have a link to another page describing the switch tag.  You should provide such a page in addition to the added entries for the page.

luizluca:
> 

OK. I'm already working on it. Those protocols are much simpler than LINKTYPE_DSA_TAG_EDSA. I also added another similar protocol (version 0x9). However, there is no public information about protocol 0xa and even linux dsa code was reversed engineered. I'll make it clear about the unknown fields.

luizluca:
> Please see [this page](https://www.tcpdump.org/linktypes.html) and produce a sufficiently detailed description of the new header types.

https://github.com/the-tcpdump-group/tcpdump-htdocs/pull/26

luizluca:
I'm just curious about how setting a different data link type will make the capture of ethernet with dsa easier. With 1.9.x, it was listing EN10MB as the default data link (as a normal ethernet device). Now it lists things like:

```
Data link types for eth0 (use option -y to set):
  DSA_TAG_EDSA (Marvell EDSA) (printing not supported)
```

And now you will need to patch tcpdump, wireshark and any other tool that consumes that pcap to undestand a new link type.
Wouldn't it be better to simply use EN10MB for all Ethertype DSA? (or, at least, set it as an alternative)

infrastation:
Whatever the right solution is, if it takes time to figure out, it is best to have the time to figure it out.

guyharris:
> Wouldn't it be better to simply use EN10MB for all Ethertype DSA?

No.  It would be worse.

[Broadcom DSA packets](https://www.tcpdump.org/linktypes/LINKTYPE_DSA_TAG_BRCM.html) do *not* have a header with a 6-octet destination address, a 6-octet source address, and a 2-octet field containing the Ethernet packet's type/length value, followed immediately by the Ethernet payload.  Instead, they either have a switch tag between the source address and the type/length field, or have a switch tag before the destination address, so they are *not* `LINKTYPE_ETHERNET`/`DLT_EN10MB` packets, and any code that dissects `LINKTYPE_ETHERNET`/`DLT_EN10MB` packets will *not* dissect them correctly.  Therefore, it would be an error to use `LINKTYPE_ETHERNET`/`DLT_EN10MB` for them.

[Marvell DSA packets](https://www.tcpdump.org/linktypes/LINKTYPE_DSA_TAG_DSA.html) do *not* have a header with a 6-octet destination address, a 6-octet source address, and a 2-octet field containing the Ethernet packet's type/length value, followed immediately by the Ethernet payload.  Instead, they have a switch tag between the source address and the type/length field, so they are *not* `LINKTYPE_ETHERNET`/`DLT_EN10MB` packets, and any code that dissects `LINKTYPE_ETHERNET`/`DLT_EN10MB` packets will *not* dissect them correctly.  Therefore, it would be an error to use `LINKTYPE_ETHERNET`/`DLT_EN10MB` for them.

The only way to use `LINKTYPE_ETHERNET`/`DLT_EN10MB` would be to strip out the switch tag, leaving only the Ethernet header.  This would be useful only if nobody wants the information from the switch tag.

If you preserve the switch tag, yes, you would have to modify tcpdump and Wireshark and any other program that reads pcap or pcapng files.  If people are interested in the switch tag information, you have to do that *anyway*.

> Doesn't pcap have a header field to add a hint for the pcap consumer which tag is in use?

No.

[pcap files](https://pcapng.github.io/pcapng/draft-ietf-opsawg-pcap.html) have a file header that contains a magic number, two must-be-zero fields (formerly semi-defined as containing values that would *not* be hints of that sort, but it's not clear they were ever used), a field containing the snapshot length, and a field containing the link-layer type and some other information, and per-packet headers containing a packet time stamp, the packet length, and the amount of captured data in the packet.  Adding such a hint to any field in the file header would mean that:

1. libpcap would have to be modified to have an API to read that hint;
2. programs *not* modified to get and understand that hint would misdissect those packets, showing a garbage dissection with no obvious indication of what the problem was, confusing users and, probably, causing a lot of bug reports for tcpdump, Wireshark, etc. with which the developers would have to deal.

The *current* scheme, with a different `LINKTYPE_`/`DLT_` value, at least means that, for programs that don't understand those types and the corresponding switch tag, you get an "I don't understand this link-layer type" error rather than a weird misdissection.

> And for other registered DSA tags, they could be treated simply as a normal proprietary Ethertype.

Not if they don't put that proprietary Ethertype in the type/length field of the Ethernet header, with the switch tag following the type/length field, and the *real* type/length field either in the switch tag or immediately after the switch tag, and the regular Ethernet payload after that.

And they don't.

guyharris:
> I noticed that DSA_TAG_EDSA was removed because the ethertype used was not registered and it was in conflict.

No, `LINTKYPE_DSA_TAG_EDSA`/`DLT_DSA_TAG_EDSA` was *not* removed.

What was removed from tcpdump, in this commit:

```
commit d5c16b1668fef7011a239c216d07024136cc49e7
Author: Vivien Didelot <vivien.didelot@gmail.com>
Date:   Sat Apr 6 16:49:10 2019 -0400
    
    Remove MEDSA ethertype printer
    
    MEDSA (which unofficially stands for Marvell EDSA) relies on the
    unregistered 0xDADA ethertype value that was randomly choosen and
    currently used by the kernel. But with EDSA this value is programmable,
    hence can be changed anytime.

    Now that a reliable DLT parser is implemented for both DSA and EDSA
    tagged master interfaces, let's get rid for the MEDSA ethertype parser.
```

was code to, well, "[treat 0xDADA] simply as a normal proprietary Ethertype", because it is *not* guaranteed that the  programmable DSA Ethernet type will be 0xDADA, and that's not a registered Ethernet type.  Instead, the *current* mechanism, in which the link-layer header type indicates the switch tag, is used.

*If* Realtek packets *always* have that registered Ethernet type, and *if* it *always* indicates that what follows the 0x8899 is that switch tag - *never* any other type of packet (such as remote control or loop detection packets), unless the "Protocol" field can be used to distinguish between DSA and other packets with an Ethertype of 0x8899 - *then* we could have tcpdump/Wireshark/etc. use the Ethertype to dissect all of those packet types, in which case this change isn't necessary (and neither is the corresponding link-layer header type list change), and, of course,  tcpdump/Wireshark/etc. would have to be modified to understand that Ethertype.

Otherwise, it can't be trusted.

luizluca:
> The only way to use `LINKTYPE_ETHERNET`/`DLT_EN10MB` would be to strip out the switch tag, leaving only the Ethernet header. This would be useful only if nobody wants the information from the switch tag.

I'm only considering Ethertype-like DSA tags. At least for Realtek and Marvell EDSA, the tag is similar to a 802.1Q. Marvell does support other tag formats but I'm not talking about them. I don't see why it should be treated differently from any other proprietary Ethernet protocol. Isn't this an Ethernet frame? Here are two examples for Realtek and Marvell EDSA:

image
image

I'm not an expert with pcap but it does look like an Ethernet proprietary protocol. Or am I missing something stupid?

A libpcap program might not know how to digest that specific Ethertype but, at least, it would be able to open a pcap file, list the frames, read the dst/src mac addresses. It would not be different than what already happens with any new proprietary ethernet protocol.

With libcap-git, an unknown tag will generate a capture using the NULL link type. Inside wireshark, for example, each frame is shown as a blob but I cannot force it to read them as Ethernet. I cannot, for example, write a dissector because there is no "anchor/condition" that would call my dissector. Wireshark needs to know at least the first layer. That's why the first layer is important.

With a registered DSA tag, it will use a custom link type (like in the case of this PR). Now, any program compiled with an older libpcap will not even open it. I'm not even able to read it as a blob (it is worse than NULL link type). All programs that I tested just fail and abort.

With libpcap-1.9, that same capture would use DLT_EN10MB as it ignored the dsa tag. Wireshark, for example, will dissect the ethernet header, show src/dst mac addresses and the say that the ethernet type is unknown. Payload is a blob but I can already use that info as is or, just like I did, implement a dissector script and read that data without patching and rebuilding wireshark.

> If you preserve the switch tag, yes, you would have to modify tcpdump and Wireshark and any other program that reads pcap or pcapng files. If people are interested in the switch tag information, you have to do that _anyway_.

I know tcpdump, wireshark and other will need to know the tag format to be able to digest the next layer. But being DLT_EN10MB will already allow me to use editcap, do some statistics, and even some raw filters. It is not ideal but it is much better than "I don't know this link type... bye"

> > And for other registered DSA tags, they could be treated simply as a normal proprietary Ethertype.
> 
> Not if they don't put that proprietary Ethertype in the type/length field of the Ethernet header, with the switch tag following the type/length field, and the _real_ type/length field either in the switch tag or immediately after the switch tag, and the regular Ethernet payload after that.

Well, for some DSA tags, they do. I think that for those cases, EN10 could be the default one. For unknown tags, EN10 could be an option that the user could select (tcpdump -y), even if the default one would be NULL.

guyharris:
> I'm only considering Ethertype-like DSA tags.

So you're *not* considering any of the Broadcom tags or the Marvell DSA tag.

And the Marvell EDSA tag can be considered only if the programmed Ethertype is set to an Ethertype registered specifically for this purpose.  Is there any guarantee that it is? If not, then *none* of the currently registered DSA tags can be considered here.

luizluca:
> was code to, well, "[treat 0xDADA] simply as a normal proprietary Ethertype", because it is _not_ guaranteed that the programmable DSA Ethernet type will be 0xDADA, and that's not a registered Ethernet type. Instead, the _current_ mechanism, in which the link-layer header type indicates the switch tag, is used.

Even in this case, why not offer EN10 as an alternative link type? It might not be the default one but it is better than nothing.

> _If_ Realtek packets _always_ have that registered Ethernet type, and _if_ it _always_ indicates that what follows the 0x8899 is that switch tag - _never_ any other type of packet (such as remote control or loop detection packets), unless the "Protocol" field can be used to distinguish between DSA and other packets with an Ethertype of 0x8899 - _then_ we could have tcpdump/Wireshark/etc. use the Ethertype to dissect all of those packet types, in which case this change isn't necessary (and neither is the corresponding link-layer header type list change), and, of course, tcpdump/Wireshark/etc. would have to be modified to understand that Ethertype.

0x8899 is registered for Realtek. There is a protocol version in the next 4 or 8 bytes after the 0x8899, even for control or loop test. However, a change is still necessary (but not this one) because libpcap will use linktype NULL if there is an unknown DSA tag in use.

Should I update this PR to simply register the realtek DSA tag as an DLT_EN10MB?

luizluca:
> > I'm only considering Ethertype-like DSA tags.
> 
> So you're _not_ considering any of the Broadcom tags or the Marvell DSA tag.
> 
> And the Marvell EDSA tag can be considered only if the programmed Ethertype is set to an Ethertype registered specifically for this purpose. Is there any guarantee that it is? If not, then _none_ of the currently registered DSA tags can be considered here.

Wouldn't it be better to offer the user the option (those listed with "tcpdump -L") to capture that data as EN10? Would it do any harm?

infrastation:
@luizluca, what Guy tells here makes a lot of sense: it would be wrong to use `DLT_EN10MB` for something that is not (by definition of the packet structure) Ethernet. Although it is acceptable (and often necessary) to break the rules in a controlled lab environment to facilitate an experiment, the release code must get things right for everyone.

As just one example, if a `DLT_EN10MB` pcap file contains DSA packets, a filter such as `src host 1.2.3.4` will produce random results because it needs to look into the IPv4 header after the Ethernet header, and in packet data that is not Ethernet (even though it contains, among other data, all the original Ethernet fields) all offsets would be wrong. Hopefully you understand why it does not look right to try to fit this kind of DSA into `DLT_EN10MB`.

Could you clarify a few points please?

* The different versions of Realtek DSA all have in common that they appear only on the internal "CPU" port of the switch chip, is this correct?
* From the DSA header alone it is difficult to tell the version, because the Version field does not have the same size and position in every version, is this correct?
* How does the capturing host know which DSA version the CPU port is supposed to have?
* What is the simplest practical way to see DSA packets on the wire?

luizluca:
> @luizluca, what Guy tells here makes a lot of sense: it would be wrong to use `DLT_EN10MB` for something that is not (by definition of the packet structure) Ethernet. Although it is acceptable (and often necessary) to break the rules in a controlled lab environment to facilitate an experiment, the release code must get things right for everyone.

That is the point that I might have some knowledge limitation. I cannot see why frames like these are not valid Ethernet II:

dst mac | src mac | 0x8899 | payload | ...
dst mac | src mac | proprietary ethtype | payload | ...

While these are:

dst mac | src mac | 0x8100 | payload | ...
dst mac | src mac | 0x0806 | payload | ...
dst mac | src mac | 0x0800 | payload | ...
dst mac | src mac | 0x86DD | payload | ...

Notice that this arrangement applies to any realtek proprietary Ethernet protocol (DSA or not)

dst mac | src mac | 0x8899 | payload | ...

If libpcap does not know how to deal with the payload for a specific ethtype, it is fine. I understand that, for Marvell that uses a non standard nor constant ethtype, link type is a trusted way to make sure the user will not have a false Marvel EDSA. But even when the user knows it is a Marvell tag, libpcap behavior is simply locking those packages, either in a NULL DLT (but readable) or in an unreadable unknown (for a time) DLT. And the worse part is that any other ethype dsa tag will pay because Marvell didn't registered a proper ethtype.

If link type was something easily overwritten by a user, it will be a minor issue. However, the libpcap behavior (for 1.9) is to abort as soon as an unknown link type appears. Maybe that is the real issue.

> As just one example, if a `DLT_EN10MB` pcap file contains DSA packets, a filter such as `src host 1.2.3.4` will produce random results because it needs to look into the IPv4 header after the Ethernet header, and in packet data that is not Ethernet (even though it contains, among other data, all the original Ethernet fields) all offsets would be wrong. Hopefully you understand why it does not look right to try to fit this kind of DSA into `DLT_EN10MB`.

I'm pretty sure it will generate lots of random results if libpcap does not apply that rules only when ethtype is 0x0800 or it is another ethtype that encapsulate IPv4 and libpcap can already parse it. From my experience, it is much easier to have a false negative match than a false positive with libpcap. 

> Could you clarify a few points please?

Sure.

> 
> * The different versions of Realtek DSA all have in common that they appear only on the internal "CPU" port of the switch chip, is this correct?
As D in DSA is distributed, they might also appear in a connection between two switches. You can configure a switch user port as a CPU port, but it is low level hacking.

> * From the DSA header alone it is difficult to tell the version, because the Version field does not have the same size and position in every version, is this correct?
I don't think so. AFAIK, it is not dubious. if ethtype is 0x8899, check the next 4 bits. I think they can be 0x1, 0x2, 0x3, 0x9 (dsa), 0xa (dsa), but only two of these are DSA tags (the others are other Realtek stuff). If it is 0x0 (I guess Realtek run out of numbers), get 4 more bits (just like UTF). If it is 0x4, it is another dsa protocol. So, the first known values of a Realtek DSA tag are 0x9, 0xa, or 0x04.

> * How does the capturing host know which DSA version the CPU port is supposed to have?
It is the dsa switch driver that configures it. Normally, when the switch supports only one type of tag, it is hardcoded. If it support multiple tags, it is normally configured as a device-tree property. Until the network device is configured, it is a normal Ethernet interface. After DSA configures it, all traffic arriving or leaving will be intercepted and pass through a "tag processor" that adds or consumes the tag (kernelsrc/net/dsa/tag*.c). It is also exposed to userland at /sys/class/net/<eth...>/dsa/tag (I think). libpcap gets the tag from that path.

> * What is the simplest practical way to see DSA packets on the wire?
From a DSA-supported router, 'tcpdump -i eth0'. Traffic without the tag will be exposed at slave devices (wan, lan1, lan2...), quite similar to 802.1Q virtual interfaces (eth0.2).

infrastation:
Thank you for the comments. You are right: my example was incorrect (unless . So the point you are making is that DSA could use the same approach as 802.1Q to remain in the encoding space of `DLT_EN10MB`. Let me note that in such case the version heuristics would be more complicated, and would need to be done per packet, as opposed to be defined once per file.

This can be rather convoluted, not the least because nobody knows what specifications and conventions the vendor is following.

guyharris:
> Even in this case, why not offer EN10 as an alternative link type? It might not be the default one but it is better than nothing.

I.e., if the user selects Ethernet as the link type, we strip out the switch tag?

guyharris:
> 0x8899 is registered for Realtek. There is a protocol version in the next 4 or 8 bytes after the 0x8899, even for control or loop test.

So net/dsa/tag_rtl4_a.c in the 5.15.10 kernel says:

```
/*
 * 0x1 = Realtek Remote Control protocol (RRCP)
 * 0x2/0x3 seems to be used for loopback testing
 * 0x9 = RTL8306 DSA protocol
 * 0xa = RTL8366RB DSA protocol
 */
#define RTL4_A_PROTOCOL_RTL8366RB       0xa
```

It doesn't mention a protocol of 0x04, other than *not* mentioning it - i.e., it doesn't say it's used for purposes other than switch tags.

[A dissector submitted for Wireshark](https://www.wireshark.org/lists/ethereal-dev/200409/msg00090.html) back in 2004 (back when it was still called Ethereal) for the Realtek Remote Control Protocol (RRCP), Realtek Echo Protocol (REP), and Realtek Loop Detection Protocol (RLDP) appears to show the first byte of the Ethernet payload containing a one-byte protocol field, with the value 0x01 for RRCP, 0x02 for REP, and 0x03 for RLDP.

So it appears that, in an Ethernet packet with an Ethertype of 0x8899, the upper 4 bits of the first byte of the Ethernet payload are:

- 0x0: something indicated by the lower 4 bits of that byte;
- 0x9: switch tag, with the lower 4 bits of that byte and the byte after that containing the tag, and with the Ethertype following the tag;
- 0xA: switch tag, with the lower 4 bits of that byte and the byte after that containing the tag, and with the Ethertype following the tag.

For 0x0, the lower 4 bits of the first byte are:

- 0x1: RRCP;
- 0x2: REP;
- 0x3: RLDP;
- 0x4: switch tag, with the 5 bytes after that containing the tag, and with the Ethertype following the tag.

So it would appear that a protocol dissector for Ethertype 0x8899 could handle RRCP, REP, RLDP, and those three switch tag types, without any heuristics (*gag*) being required.

> However, a change is still necessary (but not this one) because libpcap will use linktype NULL if there is an unknown DSA tag in use.

No, pcap-linux.c does not contain `DLT_NULL` anywhere, so that won't happen.

If an device XXX has a /sys/class/net/XXX/dsa/tagging file, and it contains an unknown DSA tag, the attempt to activate the device will fail, with a "unsupported DSA tag: {tag}" message.

> Should I update this PR to simply register the realtek DSA tag as an DLT_EN10MB?

Yes, like "none".

guyharris:
> I don't think so. AFAIK, it is not dubious. if ethtype is 0x8899, check the next 4 bits. I think they can be 0x1, 0x2, 0x3, 0x9 (dsa), 0xa (dsa), but only two of these are DSA tags (the others are other Realtek stuff). If it is 0x0 (I guess Realtek run out of numbers), get 4 more bits (just like UTF). If it is 0x4, it is another dsa protocol. So, the first known values of a Realtek DSA tag are 0x9, 0xa, or 0x04.

Not according to the RRCP/REP/RLDP dissector - and not according to the-tcpdump-group/tcpdump-htdocs#26.  The next 4 bits can be 0x0, 0x9, or 0xa; if they're 0x0, the 4 bits after that (the lower 4 bits of the first byte) indicate what follows.  See [my previous comment](https://github.com/the-tcpdump-group/libpcap/pull/1075#issuecomment-997041724).

guyharris:
> I understand that, for Marvell that uses a non standard nor constant ethtype, link type is a trusted way to make sure the user will not have a false Marvel EDSA. But even when the user knows it is a Marvell tag, libpcap behavior is simply locking those packages, either in a NULL DLT (but readable) or in an unreadable unknown (for a time) DLT.

So it goes.

> And the worse part is that any other ethype dsa tag will pay because Marvell didn't registered a proper ethtype.

They only "pay" by getting errors opening the CPU port device; once libpcap is adds support for that device, the open works.  Non-Ethertype DSA tags "pay" in the same fashion.

And there does not appear to be a way to ask the kernel whether the DSA tags are always indicated by an *official* Ethertype (as opposed to an Ethertype that might get assigned to somebody else for a different purpose, at which point, if anybody submits a tcpdump/Wireshark/etc. dissector for that Ethertype, it'll be added and that DSA tag will no longer be recognized.  (No, no heuristics.  Every time somebody does something that requires a heuristic because information is being discarded when it doesn't have to be, God kills a kitten.  Please, think of the kittens.)

luizluca:
> Thank you for the comments. You are right: my example was incorrect (unless you short-cut EtherType multiplexing to treat DSA as IPv4) . So the point you are making is that DSA could use the same approach as 802.1Q to remain in the encoding space of `DLT_EN10MB`. Let me note that in such case the version heuristics would be more complicated, and would need to be done per packet, as opposed to be defined once per file.

If a DSA tag preserves the Ethernet structure as in Ethertype DSA, yes. Ethertype DSA can even be stacked when two different Ethernet switches are in use. So, the link type trick might not play too long anyway.

> This can be rather convoluted, not the least because nobody knows what specifications and conventions the vendor is following.

My suggestion is to offer a list of link types for each DSA case, with a default one being the bold one. I think that's the function of handle->dlt_list and handle->linktype, right?

1. Known DSA tags
1.1 Ethertype-like with standard number: [**EN10**]
1.1 Ethertype-like with non-standard, possibly conflicting number: [**custom_DLT**, EN10]
2. Unknown DSA tags: [**Null_DLT**, EN10]

It is not too much different from what pcap is currently doing. The difference is to offer the user way out from "unsupported link layer type errors".

Maybe the Null DLT could be an option for any kind of link type.

Should I use EN10 for rtl* dsa tag family? How should I handle the documentation as it will not have a custom link type?

guyharris:
> 1. Known DSA tags
>   1.1 Ethertype-like with standard number: [EN10]
>   1.2 Ethertype-like with non-standard, possibly conflicting number: [custom_DLT, EN10]

You forgot

  1.3 non-Ethertype-like [custom_DLT, EN10 which strips the tags so that you actually get *Ethernet* frames rather than random weird frames]

> Unknown DSA tags: [Null_DLT, EN10]

How about

Unknown DSA tags: return an error, so we can fix libpcap to provide the appropriate behavior (Ethernet for registered-Ethertype DSA tags, custom link type for DSA tags not using a registered Ethertype), given that, for an "unknown" DSA tag, the format of the tag is, well, *unknown*, so we don't know how to modify the packet to look like an Ethernet packet.

Conveniently, for that behavior, the patch to libpcap is zero bytes long....

> Maybe the Null DLT could be an option for any kind of link type.

Or DLT_LINUX_SLL, which handles protocols *other* than IPv4 and IPv6 (remember, the so-called "null" link-layer type uses AF_ values for the protocol - no AF_ value, no support).

> Should I use EN10 for rtl* dsa tag family?

Yes.

> How should I handle the documentation as it will not have a custom link type?

No custom link type, nothing to document on the tcpdump.org Web site, so you would handle the documentation by closing the tcpdump-htdocs pull request, and *possibly* setting up a site to document the format of Ethertype-0x8899 frames, based on whatever can be dug out of Realtek data sheets and by observing Realtek network traffic and switch chip behavior.  The source code to 0x8899 dissectors than can then include the URL of that documentation.

infrastation:
The latest description possibly explains the origins of `print-rrcp.c` in tcpdump.

guyharris:
So Wikipedia *wasn't* mistaken when [they said tcpdump had a dissector for it](https://en.wikipedia.org/wiki/Realtek_Remote_Control_Protocol).

They *were* mistaken when they said Wireshark did (the one that was submitted was never added, perhaps because the author said they'd be submitting some updates soon, so we waited, but they never did).  I'll look at cleaning that one up and modernizing it, and then adding it.

infrastation:
Is there a way to reduce this complexity and to make the next step, whatever it is?

luizluca:
I'm closing this PR as I opened #1082 only adding the respective DSA tags as known Ethertype DSA tags (and using Ethernet Link Type).

I think the same change should happen with protocol "edsa".

PR #1082, opened 2022-01-06
Add support for Realtek (Ethertype) DSA data

luizluca:
Realtek switchtag rtl4a (4 bytes long, protocol 0xA) and rtl8_4 (8 bytes
long, protocol 0x04) are Ethertype DSA tags, inserted in the Ethernet
header similar to an 802.1Q tag. Both shares the same Ethertype 0x8899
as other Realtek proprietary protocols.

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -5048,6 +5048,9 @@ static struct dsa_proto {
 	{ "brcm-prepend", DLT_DSA_TAG_BRCM_PREPEND },
 	{ "dsa", DLT_DSA_TAG_DSA },
 	{ "edsa", DLT_DSA_TAG_EDSA },
+	{ "rtl4a", DLT_EN10MB },
+	{ "rtl8_4", DLT_EN10MB },
+	{ "rtl8_4t", DLT_EN10MB },
 };
 
 static int


luizluca:
The doc about this tag was submitted to wikipedia https://en.wikipedia.org/wiki/Draft:Realtek_Layer_2_Protocols

infrastation:
It looks like this project's difficulty lies mostly in the documentation and specification aspects.

luizluca:
>
> It looks like this project's difficulty lies mostly in the documentation
> and specification aspec
>
And wikipedia seems to be a no-go. They'll not accept an article with a raw
description of a protocol. They need context, history, facts, not a table
with descriptions.

infrastation:
That's a good policy, though sadly not entirely connected with their existing practice.

Anyway, the root of the problem is that the protocol designer has not published it completely, so whoever works on a decoder will face the dilemma of either doing a poor job, or publishing what was intended to be kept proprietary.

How much exactly is the encoding specified and in which documents?

guyharris:
I'd say at this point that documenting the DSA protocols is probably best done in some place such as [the Linux kernel documentation](https://github.com/torvalds/linux/tree/master/Documentation/networking/dsa), if Realtek isn't going to put anything about them in their datasheets.

I don't think documenting them is an requirement for *this* change.  It might be a good idea, even if not a requirement, for changes to tcpdump (print-realtek.c) or Wireshark ([upcoming packet-realtek.c](https://gitlab.com/wireshark/wireshark/-/merge_requests/6004)) to dissect those protocols.

luizluca:
> DSA protocols is probably best done in some place such as [the Linux kernel documentation](https://github.com/torvalds/linux/tree/master/Documentation/networking/dsa)

We would need to start a new trend as there is no tag format described there. I only saw some tag doc in the header of each tag_*.c implementation.

luizluca:
I've just pushed an update that also adds the rtl8_4t tag format. It is equals to rtl8_4 but positioned before the CRC. tcpdump does not bother the extra bytes after the payload and filters work as expected.

With wireshark, I'll assume by default the frame has a FCS and it will show part of the tag as a trailer and part of if as a failed frame sequence (although filters seem to work as expected). The user can disable the FCS warning changing the "Assuming packets have FCS" from "heuristic" to "never". This way, tag will be shown as a trailer.

guyharris:
So does this mean that the frames delivered to `PF_PACKET` sockets by the kernel for those DSA protocols will have *no* switch tag header, just an Ethernet packet?

guyharris:
> So does this mean that the frames delivered to PF_PACKET sockets by the kernel for those DSA protocols will have no switch tag header, just an Ethernet packet?

I.e., the first 14 bytes of the packet are a 6-byte destination MAC address, a 6-byte source MAC address, and a 2-byte Ethernet tie value of 0x8899, with a tag switch at the beginning of the Ethernet *payload*, rather than one of the DSA formats with a switch tag inserted before, or in the middle of, the Ethernet *header*?

luizluca:
> So does this mean that the frames delivered to `PF_PACKET` sockets by the kernel for those DSA protocols will have _no_ switch tag header, just an Ethernet packet?

If you are asking about rtl8_4t, yes, the packet will look like an ordinary ethernet frame with some extra bytes after payload and before FCS (frame check sequence)

luizluca:
Thanks, @guyharris

guyharris:
> There is no reason to treat realtek DSA tags as non-ethernet.

At least, no reason not to treat the DSA tagged frames with 0x8899 as non-Ethernet, as the Ethertype indicates that it has some Realtek protocol in it and the protocol can be determined from the header following the Ethernet header.

Marking rtl8_4t frames with a separate link-layer type would allow the dissector for that type to know for certain that there's a tag at the end of the frame, without needing to use any heuristics.

PR #1084, opened 2022-01-17
Add support for GNU/kFreeBSD, from Aurelien Jarno and others

LocutusOfBorg:


infrastation:
Here FreeBSD and illumos fail for reasons unrelated to Valgrind (missing header, the same as NetBSD and OpenBSD). These changes require a bit more work to be correct.

guyharris:
> Here FreeBSD and illumos fail for reasons unrelated to Valgrind (missing header, the same as NetBSD and OpenBSD). These changes require a bit more work to be correct.

Not all system that use BPF as a capture mechanism have <net/if_pflog.h> or librt.  All \*BSDs, macOS, and Solaris (and Illumos), use BPF as a capture mechanism.

```
$ cat foo.c
#include <net/if_pflog.h>

int
main(void)
{
	return 0;
}
$ cc -o foo foo.c -lrt
foo.c:1:10: fatal error: 'net/if_pflog.h' file not found
#include <net/if_pflog.h>
         ^~~~~~~~~~~~~~~~
1 error generated.
$ ed foo.c
57
/if_pflog/d
w
31
q
$ cat foo.c

int
main(void)
{
	return 0;
}
$ cc -o foo foo.c -lrt
ld: library not found for -lrt
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

LocutusOfBorg:
ok so maybe this can be keep as Debian specific patch

guyharris:
> ok so maybe this can be keep as Debian specific patch

Or make the configure script and CMake files test whether the header files in question and the library in question exist and only use them if they are.

LocutusOfBorg:
Sorry but I find difficult to do it in a smart way... :/

guyharris:
OK, closed.

PR #1085, opened 2022-01-17
From Debian, tcpdump manpage is section 8 not section 1

LocutusOfBorg:


guyharris:
Why should it be in section 8?  It's not a system administration command.  (Not everybody who captures network traffic or reads capture files is a system administrator.)

LocutusOfBorg:
Hello, ok you have a good point :)
I'll close it thanks for the quick answer!

PR #1088, opened 2022-01-26
Libpcap 1.10

xfliu2665:
a early exit while pcap_loop is required to run forever : i found a case in read_op when **"h.h3->hdr.bh1.block_status == TP_STATUS_KERNEL**" is true, pcap_loop may exit in early


guyharris:
If you have a fix, pleases submit it as a pull request for the mater branch, not for the 1.10 branch.  Once that works, we can backport it to the 1.10 branch with `git cherry-pick -x`.

guyharris:
It's a large collection of changes that appear to have nothing to do with the problem in question, and it's still applied to the 1.10 branch.

If this is fixed in the current main branch, but not fixed in the 1.10 branch, please see what change in the main branch fixed it, and indicate what change that is, so we can back port it.

If it's *not* fixed in the current main branch, please make it a pull request for the *main* branch, with a meaningful commit message ("Libpcap 1.10" isn't meaningful), so that we can merge it and then backport it to the 1.10 branch.

PR #1092, opened 2022-02-21
autoconf: Use AS_HELP_STRING macro in configure.ac for '--enable-remote'

fxlb:
There is no need to have:
  --enable-remote         enable remote packet capture [default=no]
  --disable-remote        disable remote packet capture

Use only:
  --enable-remote         enable remote packet capture [default=no]

### DIFF
--- a/configure
+++ b/configure
@@ -1396,7 +1396,6 @@ Optional Features:
   --disable-protochain    disable \"protochain\" insn
   --enable-ipv6           build IPv6-capable version [default=yes]
   --enable-remote         enable remote packet capture [default=no]
-  --disable-remote        disable remote packet capture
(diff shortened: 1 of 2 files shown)


fxlb:
This proposal is for master and libpcap-1.10 branches.
Any reason to object?

infrastation:
I wonder why some similar flags use `[default=no]` and other use `@<:@default=no@:>@`. Other than that, the change looks consistent.

guyharris:
> Not sure they are needed.

Their main purpose is to deal with `[` and `]` being quoting characters in autoconf's flavor of M4; I've had to use them in at least one case due to that, as I remember.

PR #1096, opened 2022-03-01
cmake: fix when using libpcap Cmake out of this directory

clementperon:
Hi,

I'm using FetchContent_Declare() and FetchContent_MakeAvailable() to build locally the libpcap.

But i got an issue when I link with pcap_static target the compiler can't find pcap.h.

This is because no include_directories are defined for the target.

Let's add a proper define.

### DIFF
--- a/pcap-bpf.c
+++ b/pcap-bpf.c
@@ -661,7 +661,11 @@ bpf_bind(int fd, const char *name, char *errbuf)
 		case ENXIO:
 			/*
 			 * There's no such device.
+			 *
+			 * There's nothing more to say, so clear out the
+			 * error message.
 			 */
+			errbuf[0] = '\0';
 			return (PCAP_ERROR_NO_SUCH_DEVICE);
 
 		case ENETDOWN:
@@ -3052,7 +3056,11 @@ monitor_mode(pcap_t *p, int set)
 		case ENXIO:
 			/*
 			 * There's no such device.
+			 *
+			 * There's nothing more to say, so clear the
+			 * error message.
 			 */
+			p->errbuf[0] = '\0';
 			close(sock);
 			return (PCAP_ERROR_NO_SUCH_DEVICE);
 
--- a/pcap-dag.c
+++ b/pcap-dag.c
@@ -793,9 +793,14 @@ static int dag_activate(pcap_t* p)
 		/*
 		 * XXX - does this reliably set errno?
 		 */
-		if (errno == ENOENT)
+		if (errno == ENOENT) {
+			/*
+			 * There's nothing more to say, so clear
+			 * the error mesage.
+			 */
 			ret = PCAP_ERROR_NO_SUCH_DEVICE;
-		else if (errno == EPERM || errno == EACCES) {
+			p->errbuf[0] = '\0';
+		} else if (errno == EPERM || errno == EACCES) {
 			ret = PCAP_ERROR_PERM_DENIED;
 			snprintf(p->errbuf, PCAP_ERRBUF_SIZE,
 			    "Attempt to open %s failed with %s - additional privileges may be required",
--- a/pcap-dlpi.c
+++ b/pcap-dlpi.c
@@ -366,8 +366,12 @@ open_dlpi_device(const char *name, u_int *ppa, char *errbuf)
 	 * chop off the unit number, so "dname" is just a device type name.
 	 */
 	cp = split_dname(dname, &unit, errbuf);
-	if (cp == NULL)
+	if (cp == NULL) {
+		/*
+		 * split_dname() has filled in the error message.
+		 */
 		return (PCAP_ERROR_NO_SUCH_DEVICE);
+	}
 	*cp = '\0';
 
 	/*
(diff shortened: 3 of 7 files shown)


guyharris:
This uses generator expressions, which appear to have been introduced in CMake 2.8.11.  As such, the `cmake_minimum_required()` value for UN*X should be increased to at least 2.8.11 from the 2.8.6 in the current CMakeLists.txt.

CMakeLists.txt
```diff
@@ -2723,7 +2731,7 @@ if(WIN32 OR CYGWIN OR MSYS)
         endif(MSVC)
     endif(MSVC AND CMAKE_SIZEOF_VOID_P EQUAL 8)
 else(WIN32 OR CYGWIN OR MSYS)
-    install(TARGETS ${LIBRARIES_TO_INSTALL} DESTINATION ${CMAKE_INSTALL_FULL_LIBDIR})
```
guyharris, line 2726: So what does this change do?  [The latest version of the documentation for the GNUInstallDirs module](https://cmake.org/cmake/help/v3.23/module/GNUInstallDirs.html) says that `CMAKE_INSTALL_<dir>` can be a relative path but `CMAKE_INSTALL_FULL_<dir>` is an absolute path, usually produced by prepending `[CMAKE_INSTALL_PREFIX](https://cmake.org/cmake/help/v3.23/variable/CMAKE_INSTALL_PREFIX.html#variable:CMAKE_INSTALL_PREFIX)`, but with some [special cases](https://cmake.org/cmake/help/v3.23/module/GNUInstallDirs.html#special-cases) handled.

Are the special cases relevant here?

clementperon: Using the absolute path will make the cmake_install.cmake generated install path with absolute path:
```
if("x${CMAKE_INSTALL_COMPONENT}x" STREQUAL "xUnspecifiedx" OR NOT CMAKE_INSTALL_COMPONENT)
  list(APPEND CMAKE_ABSOLUTE_DESTINATION_FILES
   "/usr/local/lib/libpcap.a")
  if(CMAKE_WARN_ON_ABSOLUTE_INSTALL_DESTINATION)
    message(WARNING "ABSOLUTE path INSTALL DESTINATION : ${CMAKE_ABSOLUTE_DESTINATION_FILES}")
  endif()
  if(CMAKE_ERROR_ON_ABSOLUTE_INSTALL_DESTINATION)
    message(FATAL_ERROR "ABSOLUTE path INSTALL DESTINATION forbidden (by caller): ${CMAKE_ABSOLUTE_DESTINATION_FILES}")
  endif()
  file(INSTALL DESTINATION "/usr/local/lib" TYPE STATIC_LIBRARY FILES "/home/clement/work/PcapPlusPlus/Dist/_deps/pcap-build/libpcap.a")
endif()

```

Using relative path will generate a cmake_install.cmake like this where the PREFIX can still be defined at install time:
```
if("x${CMAKE_INSTALL_COMPONENT}x" STREQUAL "xUnspecifiedx" OR NOT CMAKE_INSTALL_COMPONENT)
  file(INSTALL DESTINATION "${CMAKE_INSTALL_PREFIX}/lib" TYPE STATIC_LIBRARY FILES "/home/clement/work/PcapPlusPlus/Dist/_deps/pcap-build/libpcap.a")
endif()
```
   
   
   
   Using the

clementperon:
@guyharris what do you think about the new revision pushed ?

clementperon:
@guyharris coud you make another review on the new version? Thanks

aloisklink:
Hi, @clementperon, I'm not 100% why you closed this, but this is a feature I'm still interested in!

Our CMake workaround to get `pcap_static` working using `FetchContent` is doing the following and making a dummy `PCAP::pcap` target and adding `include_directories` to that, but merging this would fix it:

```cmake
  FetchContent_Declare(
    libpcap
    URL https://github.com/the-tcpdump-group/libpcap/archive/refs/tags/libpcap-1.10.1.tar.gz
    URL_HASH SHA3_256=9aedcbec09b7b3b01c78cc80822c505846d73928a72ae96eb907b1f467eee649
  )
  FetchContent_MakeAvailable(libpcap)

  # pcap_static does not declare include_directories, so we need to manually add them
  add_library(PCAP::pcap INTERFACE IMPORTED)
  target_link_libraries(PCAP::pcap INTERFACE pcap_static)
  target_include_directories(PCAP::pcap INTERFACE "${libpcap_SOURCE_DIR}")
  target_compile_definitions(
    # pcap requires BSD types to be defined, e.g. u_int/u_short/u_char
    # the following syntax works for uClibc, glibc, musl libc
    PCAP::pcap INTERFACE
      "_BSD_SOURCE" # deprecated in glibc >2.20
      "_DEFAULT_SOURCE" # only added in glibc >2.19, musl >=1.1.5
  )
```

clementperon:
@aloisklink I have push -f in my master branch so this automatically close this MR.

You can add you're test to this new MR https://github.com/the-tcpdump-group/libpcap/pull/1135

PR #1099, opened 2022-03-11
libpcap: Avoid openssl full library path

rprasanth:
OPENSSL_LIBRARIES provide the entire path
When I used the pcap-config to grep the LIBS in ppp build,
it fails with below error,
cannot find -l/path-to-/lib/libssl.so
cannot find -l/path-to-/toolchain/lib/libcrypto.so


guyharris:
This doesn't work with MinGW on Windows; see the failing AppVeyor build.

PR #1109, opened 2022-05-15
Add speed specific USB 2.0/1.1/1.0 linktypes

desowin:
USB 2.0/1.1/1.0 devices (or 3.x and newer when connected to hosts that
are not Super-Speed capable) operate at one of three speeds:
  * Low-Speed (1.5 Mbps)
  * Full-Speed (12 Mbps)
  * High-Speed (480 Mbps)

While the packets are generally common to all three speeds, there are
some differences that span across different layers. The capture speed
should be available to analyzer and it makes sense to expose it via
speed specific linktypes. Existing USB 2.0/1.1/1.0 captures can be
manually, retroactively updated to speed specific linktype.

Low-Speed device will always operate at Low-Speed, regardless of what
host it is connected to. USB cable connected to Low-Speed device only
ever carries packets sent at Low-Speed.

High-Speed device will operate at High-Speed when connected to High-Speed
capable host. If High-Speed Detection Handshake succeeds, the cable will
only carry High-Speed packets.

Full-Speed device will always operate at Full-Speed, regardless of what
host it is connected to. High-Speed device connected to Full-Speed host
or hub, will operate at Full-Speed. The cable connected to device
operating at Full-Speed can only carry Full-Speed or Low-Speed packets
preceded by Low-Speed preamble (PRE packet) sent at Full-Speed. Non-hub
device operating at Full-Speed always ignores Low-Speed packets. The hub
forwards the Low-Speed packets to downstream ports but does not act upon
the actual Low-Speed packet contents. The packets intended for device
operating at Full-Speed are always sent at Full-Speed.

### DIFF
--- a/pcap-common.c
+++ b/pcap-common.c
@@ -1194,6 +1194,8 @@
 
 /*
  * USB 2.0, 1.1, and 1.0 packets as transmitted over the cable.
+ * Deprecated in favor of speed specific LINKTYPEs: LINKTYPE_USB_2_0_LOW_SPEED,
+ * LINKTYPE_USB_2_0_FULL_SPEED and LINKTYPE_USB_2_0_HIGH_SPEED.
  */
 #define LINKTYPE_USB_2_0	288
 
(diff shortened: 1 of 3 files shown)


mcr (approved):

infrastation:
For posterity, `LINKTYPE_USB_2_0_LOW_SPEED`, `LINKTYPE_USB_2_0_FULL_SPEED` and `LINKTYPE_USB_2_0_HIGH_SPEED` became available in libpcap 1.10.6.

PR #1111, opened 2022-06-04
cmake: suppress CMP0042 OLD deprecated warning

### LINKED ISSUE #1094: CMake (version 3.22.1) Deprecation Warning
```
CMake Deprecation Warning at CMakeLists.txt:16 (cmake_policy):
  The OLD behavior for policy CMP0042 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.
```

aloisklink:
Suppresses the CMake CMP0042 OLD deprecation warning (closes #1094)

The CMP0042 NEW behaviour sets the target property `MACOSX_RPATH` to TRUE by default.
However, this behaviour can also be controlled by `CMAKE_MACOSX_RPATH`, which `MAXOSX_RPATH` will default to, if set.

For more information, see CMake docs on [CMP0042](https://cmake.org/cmake/help/latest/policy/CMP0042.html) and [MACOSX_RPATH](https://cmake.org/cmake/help/latest/prop_tgt/MACOSX_RPATH.html#prop_tgt:MACOSX_RPATH).

`CMAKE_MACOSX_RPATH` is only set if undefined, allowing users to change this option.

According to the CMake docs, it should be exactly the same behaviour, but it might be worth somebody with a MacOS device testing this, since I don't have a MacOS device.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,8 +12,15 @@ endif(WIN32)
 # neither do we with autotools; don't do so with CMake, either, and
 # suppress warnings about that.
 #
+# Setting CMAKE_MACOSX_RPATH to FALSE uses the old behavior,
+# but removes the POLICY CMP0042 OLD deprecated warning.
+# See https://cmake.org/cmake/help/latest/policy/CMP0042.html
+#
+if (NOT DEFINED CMAKE_MACOSX_RPATH)
+    set(CMAKE_MACOSX_RPATH FALSE)
+endif()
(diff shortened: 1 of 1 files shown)


guyharris:
See also 16e1e243a9e9f03d8e8f4ee1067ba46795439a8d.

PR #1113, opened 2022-07-01
Close the eventfd if we are non-blocking

fenner:
The eventfd is used to break out of a poll() before it
times out, used by pcap_breakloop().  If we are non-blocking,
then the eventfd is never needed, so we close it.  (And
open a new eventfd if we switch to blocking).

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -842,7 +842,10 @@ static void	pcap_cleanup_linux( pcap_t *handle )
 		handlep->device = NULL;
 	}
 
-	close(handlep->poll_breakloop_fd);
+	if (handlep->poll_breakloop_fd != -1) {
+		close(handlep->poll_breakloop_fd);
+		handlep->poll_breakloop_fd = -1;
+	}
 	pcap_cleanup_live_common(handle);
 }
 
@@ -941,7 +944,8 @@ static void pcap_breakloop_linux(pcap_t *handle)
 
 	uint64_t value = 1;
 	/* XXX - what if this fails? */
-	(void)write(handlep->poll_breakloop_fd, &value, sizeof(value));
+	if (handlep->poll_breakloop_fd != -1)
+		(void)write(handlep->poll_breakloop_fd, &value, sizeof(value));
 }
 
 /*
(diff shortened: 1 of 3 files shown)


pcap-linux.c
```diff
@@ -3375,10 +3378,19 @@ pcap_setnonblock_linux(pcap_t *handle, int nonblock)
...
+			close(handlep->poll_breakloop_fd);
+			handlep->poll_breakloop_fd = -1;
+		}
 	} else {
 		if (handlep->timeout < 0) {
 			handlep->timeout = ~handlep->timeout;
 		}
+		if (handlep->poll_breakloop_fd == -1) {
+			/* If we did not have an eventfd, open one now that we are blocking. */
+			handlep->poll_breakloop_fd = eventfd(0, EFD_NONBLOCK);
```
guyharris, line 3392: This should probably check whether the attempt to (re)open the eventfd fails and, if so, return an error.  (That means the FD should be reopened before changing the timeout, so that, on an error return, non-blocking mode is left on.)

pcap-linux.c
```diff
@@ -3419,10 +3431,16 @@ static int pcap_wait_for_frames_mmap(pcap_t *handle)
 	struct ifreq ifr;
 	int ret;
 	struct pollfd pollinfo[2];
+	int numpollinfo;
 	pollinfo[0].fd = handle->fd;
 	pollinfo[0].events = POLLIN;
-	pollinfo[1].fd = handlep->poll_breakloop_fd;
-	pollinfo[1].events = POLLIN;
+	if ( handlep->poll_breakloop_fd == -1 ) {
+		numpollinfo = 1;
```
fenner, line 3438: A colleague noticed that `pollinfo[1].revents` is tested below, so should be zeroed here.

guyharris: ...with a comment noting that this is done so that both elements of `pollinfo[]` will be set after a successful call to `poll()` even when `numpollinfo` is 1.

(Setting them both requires extra stores but avoids a conditional branch, albeit one that, at least if the branch predictor can store enough predictions, will be predicted correctly most of the time; I'm not sure which is better from a performance point of view.

From a code-ugliness point of view, avoiding the `if` statement avoids another level of indirection.)

pcap-linux.c
fenner: I've added this check and a corresponding error, and test that the error value is returned in testprogs/nonblocktest.c.

pcap-linux.c
fenner: I've added the code and the comment; please see if you think it's clear enough.

guyharris: Look good.

PR #1129, opened 2022-10-19
fuzz: Generate unique temporary file names to enable parallel fuzzing

SweetVishnya:
This PR enables parallel fuzzing with libFuzzer. Previously the same pcap file name was used to write data buffer. So, parallel libFuzzer workers were overwriting the same file. Now all workers have their own unique temporary file.

### DIFF
--- a/testprogs/fuzz/fuzz_both.c
+++ b/testprogs/fuzz/fuzz_both.c
@@ -3,6 +3,7 @@
 #include <fcntl.h>
 #include <errno.h>
 #include <string.h>
+#include <unistd.h>
 
 #include <pcap/pcap.h>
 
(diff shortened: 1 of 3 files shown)


testprogs/fuzz/fuzz_both.c
```diff
@@ -3,6 +3,7 @@
 #include <fcntl.h>
 #include <errno.h>
 #include <string.h>
+#include <unistd.h>
```
guyharris, line 6: With this change, you will need to arrange that the fuzz-testing infrastructure not be compiled on Windows, as Windows doesn't have unistd.h, or come up with an alternative for Windows.  Currently the Windows build is failing with this pull request.

SweetVishnya: Yeah, I disabled fuzz targets build on Windows. Could you, please, restart netbsd-mips64 CI?

SweetVishnya:
The CI is completely green. I suppose this PR can be merged to upstream.

PR #1135, opened 2022-11-11
Fix external cmake

clementperon:
Hi,

I'm using FetchContent_Declare() and FetchContent_MakeAvailable() to build locally the libpcap.

But i got an issue when I link with pcap_static first the compiler can't find pcap.h.
This is because no include_directories are defined for the target.

Also when trying to install some path are not correct because we evaluate absolute path instead of keeping relative one


clementperon:
@infrastation @guyharris I have rebased on master, if you want to have a look.

guyharris:
> Hi, there is one more issue with FetchContent. Directive target_link_libraries is missing for static build so it does not work out-of-the-box.

Does that mean that this change is not sufficient to make *anything* work, or is that a separate problem that can be fixed with a separate change?

clementperon:
> > Hi, there is one more issue with FetchContent. Directive target_link_libraries is missing for static build so it does not work out-of-the-box.
> 
> Does that mean that this change is not sufficient to make _anything_ work, or is that a separate problem that can be fixed with a separate change?

Indeed, it's two seperate issues

clementperon:
> > Hi, there is one more issue with FetchContent. Directive target_link_libraries is missing for static build so it does not work out-of-the-box.
> 
> @Jihadist Does that mean that this change is not sufficient to make _anything_ work, or is that a separate problem that can be fixed with a separate change?

@guyharris I have created an example here with a CI to show that it works. 
https://github.com/clementperon/pcap-exam/blob/main/CMakeLists.txt
https://github.com/clementperon/pcap-exam/blob/main/CMakeLists.txt

@Jihadist Could you explain what is your issue ?

Jihadist:
@clementperon looks like it works. I have no details of old issue now. But you tested both scenarios and it ok.

Jihadist (approved):

fxlb:
> Also bump the minimum cmake version to 2.8.11 to allow generator expressions.

You don't do that. The currently minimum required versions are: 2.8.12 or 3.12 on Windows.

guyharris:
> Also bump the minimum cmake version to 2.8.11 to allow generator expressions.

This is why I love the `--amend` flag to `git commit`.  My pull/merge requests consist of one commit, and its commit message represents current reality, rather than changes that no longer apply.

fxlb:
> This is why I love the `--amend` flag to `git commit`. My pull/merge requests consist of one commit, and its commit message represents current reality, rather than changes that no longer apply.

+1

clementperon:
> > Also bump the minimum cmake version to 2.8.11 to allow generator expressions.
> 
> This is why I love the `--amend` flag to `git commit`. My pull/merge requests consist of one commit, and its commit message represents current reality, rather than changes that no longer apply.

Sorry I didn't understand it, But indeed I forget to edit the commit message when I rebase on top of master :(

guyharris:
> Sorry I didn't understand it

It means that all my pull/merge requests request that a *single* commit be merged.  If a change is need to that commit, I commit it to my local repository with `git commit --amend`, so that it modifies the previous commit rather than making a new commit, and lets the user change the commit message as necessary as well.

PR #1142, opened 2022-12-07
Add an option to force memory-mapped buffers to be mapped as accessible to 32-bit code

hansleidekker:
This is useful for Wine WoW64 support where 32-bit Windows applications run in a
64-bit Linux process. We can't pass 64-bit packet pointers back to the caller, obviously,
so we'd need to make a copy. The one shot buffer is a copy already so that would be
be copied twice, in libpcap and in Wine.

This option tells mmap() to allocate the buffers below 2GB so Wine can avoid copying them.

### DIFF
--- a/pcap-int.h
+++ b/pcap-int.h
@@ -108,6 +108,11 @@ extern int pcap_new_api;
 extern int pcap_utf_8_mode;
 
 /*
+ * Map packet buffers with 32-bit addresses.
+ */
+extern int pcap_mmap_32bit;
+
+/*
  * Swap byte ordering of unsigned long long timestamp on a big endian
  * machine.
  */
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -829,7 +829,7 @@ static void	pcap_cleanup_linux( pcap_t *handle )
 	}
 
 	if (handlep->oneshot_buffer != NULL) {
-		free(handlep->oneshot_buffer);
+		munmap(handlep->oneshot_buffer, handle->snapshot);
 		handlep->oneshot_buffer = NULL;
 	}
 
@@ -2661,14 +2661,14 @@ static int
 setup_mmapped(pcap_t *handle, int *status)
 {
 	struct pcap_linux *handlep = handle->priv;
-	int ret;
+	int ret, flags = MAP_ANONYMOUS | MAP_PRIVATE;
 
 	/*
 	 * Attempt to allocate a buffer to hold the contents of one
 	 * packet, for use by the oneshot callback.
 	 */
-	handlep->oneshot_buffer = malloc(handle->snapshot);
-	if (handlep->oneshot_buffer == NULL) {
+	handlep->oneshot_buffer = mmap(0, handle->snapshot, PROT_READ | PROT_WRITE, flags, -1, 0);
+	if (handlep->oneshot_buffer == MAP_FAILED) {
 		pcap_fmt_errmsg_for_errno(handle->errbuf, PCAP_ERRBUF_SIZE,
 		    errno, "can't allocate oneshot buffer");
 		*status = PCAP_ERROR;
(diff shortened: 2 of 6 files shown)


pcap-int.h
```diff
@@ -107,6 +107,11 @@ extern int pcap_new_api;
  */
 extern int pcap_utf_8_mode;
 
+/*
+ * Map packet buffers below 2GB on Linux.
```
guyharris, line 111: Wine isn't Linux-only, so this could conceivably be an issue on other platforms.

I might change the comment to "Map packet buffers with 32-bit addresses".

guyharris: Furthermore, if memory-mapped capture *isn't* being done, would it be necessary to allocate the buffer into which packets are read (e.g., with BPF devices) at a sufficiently-low address (as per [Microsoft's documentation of the address space layout in 32-bit Windows](https://learn.microsoft.com/en-us/windows/win32/memory/virtual-address-space))?

pcap-linux.c
```diff
@@ -2661,14 +2661,17 @@ static int
...
 	/*
 	 * Attempt to allocate a buffer to hold the contents of one
 	 * packet, for use by the oneshot callback.
 	 */
-	handlep->oneshot_buffer = malloc(handle->snapshot);
-	if (handlep->oneshot_buffer == NULL) {
+#ifdef MAP_32BIT
+	if (pcap_mmap_32bit) flags |= MAP_32BIT;
+#endif
+	handlep->oneshot_buffer = mmap(0, handle->snapshot, PROT_READ | PROT_WRITE, flags, -1, 0);
```
guyharris, line 2673: Presumably `mmap()`, at least on Linux, supports non-multiple-of-the-page-size ''length'' arguments, with the mapping size being rounded up to a multiple of the page size.

guyharris: POSIX notes that "Use of *mmap*() may reduce the amount of memory available to other memory allocation functions."  Should it use `malloc()` if `pcap_mmap_32bit` isn't set and `mmap()` if it is?

pcap-int.h
hansleidekker: > Wine isn't Linux-only, so this could conceivably be an issue on other platforms. 
> I might change the comment to "Map packet buffers with 32-bit addresses".

Thanks for the review. True, it would be nice to support this on macOS too, for example, but it's 
not straightforward there.

In the default 32-bit address layout on Windows the low 2GB is reserved for user space and the system uses the high 2GB. This maps nicely to the MAP_32BIT flag on Linux which enforces the same boundary. macOS does have a MAP_32BIT flag but it enforces a 4GB boundary, which may upset some Windows applications.

There's a non-default 3-1 split on Windows where apps can use 3GB, provided they set the Large Address Aware flag. They could still get confused by addresses reserved for the system but in practice this may work well enough, so it may still be worth supporting this scenario.

However, it's not as simple as passing the MAP_32BIT flag because 64-bit macOS maps all of the first 4GB to the zero page, making it inaccessible by default. We'd need to pass a
linker flag (-pagezero_size) to work around this limitation.

hansleidekker: > Furthermore, if memory-mapped capture _isn't_ being done, would it be necessary to allocate the buffer into which packets are read (e.g., with BPF devices) at a sufficiently-low address (as per [Microsoft's documentation of the address space layout in 32-bit Windows](https://learn.microsoft.com/en-us/windows/win32/memory/virtual-address-space))?

I'm not sure how you would allocate the buffer below e.g. the 2GB limit without using mmap(). It probably makes sense to return an error when PCAP_MMAP_32BIT can't be supported, so that the caller can fall back to a different strategy.

pcap-linux.c
hansleidekker: > Presumably `mmap()`, at least on Linux, supports non-multiple-of-the-page-size _length_ arguments, with the mapping size being rounded up to a multiple of the page size.

That's correct.

hansleidekker: > POSIX notes that "Use of _mmap_() may reduce the amount of memory available to other memory allocation functions." Should it use `malloc()` if `pcap_mmap_32bit` isn't set and `mmap()` if it is?

Perhaps. Large allocations with malloc() (i.e. those exceeding the page size) are backed by the same anonymous mappings, but because of the rounding we'd waste half a page on average. I'm not sure if saving that memory is worth complicating the code.

hansleidekker:
I have pushed a new version that extends support to BPF devices and updates the documentation as suggested.

hansleidekker:
Is there anything else I can do to make this pull request acceptable?

PR #1143, opened 2022-12-12
Added ALDL link-layer type 297

jakem72360:
Added link-layer type 297 for GM ALDL as described here: [Holden - ALDL (8192 Baud)](https://holden.asger.xyz/wiki/Holden:ALDL_(8192_Baud))

### DIFF
--- a/pcap-common.c
+++ b/pcap-common.c
@@ -1240,7 +1240,15 @@
  */
 #define LINKTYPE_AUERSWALD_LOG	296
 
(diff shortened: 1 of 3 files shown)


infrastation:
This DLT request at least refers to a specification. What is the DLT packet structure: is it raw ALDL frames or there is an additional per-packet header with meta-data?

infrastation:
Found the discussion in tcpdump-htdocs, please disregard the previous question.

Issue #1146, opened 2022-12-21
Two crashes caused by null function pointer invocation when fuzzing


hopper-vul:
Hi, libpcap team

   We have found two crashes by using our fuzzing tool. The crash details are described bellow.
 
   The crash 1 is caused by invoking `pcap_breakloop()` followed by the `pcap_open_dead()`.
   The crash 2 is caused by invoking `pcap_can_set_rfmon()` followed by the `pcap_fopen_offline()`.

The root causes of those two crashes are the inner implementation of `pcap_open_dead` has't initialized the `breakloop_op` callback and the inner implementation of `pcap_fopen_offline()` hasn't initialized the `can_set_rfmon_op` callback, and the document hasn't mentioned these.

By inspecting the related codes and understanding the functionality of the two APIs, i found that the initialization of those two callbacks may be missed, for example, the `pcap_open_dead()` nearly initializes all callbacks and return -1 as a result, as it produces a fake pcap_t which not support those operations, except the `breakloop_op` (which should also not support).

hopper-vul:
The two trigger cases mentioned above are listed bellow:

```
// file: test_1.c
// gcc -I/path/to/libpcap -L. -lpcap test_1.c -o test_1.out -g
#include "pcap.h"
#include <stdlib.h>
typedef int32_t i32;
int main() {
    i32 v0 = -1647366275; // arg1
    i32 v1 = 65535; // arg2
    struct pcap *v2 = pcap_open_dead(v0, v1); // arg1
    if (v2 == NULL) return 0;
    struct pcap *v4 = v2; // arg1
    pcap_breakloop(v4); // null function pointer crash!
}
```

```
// file: test_1.c
// gcc -I/path/to/libpcap -L. -lpcap test_2.c -o test_2.out -g
#include "pcap.h"
#include <stdlib.h>
#include <string.h>
typedef uint8_t   u8;   
typedef int8_t  i8;
typedef int32_t i32;
int main() {
    u8 v0_tmp[] = {212, 195, 178, 161, 2, 0, 0, 0, 26, 43, 60, 77, 251, 164, 21, 237, 140, 50, 93, 74, 154, 4, 31, 94, 215, 86, 241, 155, 161, 231, 0, }; // file_buf
    u8 *v0 = malloc(sizeof v0_tmp);
    memcpy(v0, v0_tmp, sizeof v0_tmp);
    char* v1 = "file___filename_0"; // __filename
[...]

PR #1147, opened 2022-12-21
Fix two null pointer crashes of breakloop and can_set_rfmon

hopper-vul:
`pcap_open_dead` and `pcap_fopen_offline` has not initialized the `breakloop_op` and `can_set_rfmon_op` callback respectively, if `pcap_breakloop()` is called followed by `pcap_open_dead()` and `pcap_can_set_rfmon()` is called followed by `pcap_fopen_offline()` then the null function pointer crashes will happen.

This commit adds two default implementations `pcap_breakloop_dead` and `sf_cant_set_rfmon` and uses them to initialize those two missed callbacks.

fixes: #1146

Signed-off-by: hopper-vul <hopper.vul@gmail.com>

### DIFF
--- a/pcap.c
+++ b/pcap.c
@@ -4212,6 +4212,14 @@ pcap_read_dead(pcap_t *p, int cnt _U_, pcap_handler callback _U_,
 }
 
 static int
+pcap_breakloop_dead(pcap_t *p)
+{
+	snprintf(p->errbuf, PCAP_ERRBUF_SIZE,
+	    "A breakloop cannot be set on a pcap_open_dead pcap_t");
+	return (-1);
+}
+
+static int
 pcap_inject_dead(pcap_t *p, const void *buf _U_, int size _U_)
 {
 	snprintf(p->errbuf, PCAP_ERRBUF_SIZE,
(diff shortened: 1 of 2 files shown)


pcap.c
```diff
@@ -4211,6 +4211,14 @@ pcap_read_dead(pcap_t *p, int cnt _U_, pcap_handler callback _U_,
 	return (-1);
 }
 
+static int
```
guyharris, line 4214: A `breakloop_op` has type `void (*)(pcap_t *)`, not `int (*)(pcap_t *)`, so this should have that type.

pcap.c
```diff
@@ -4211,6 +4211,14 @@ pcap_read_dead(pcap_t *p, int cnt _U_, pcap_handler callback _U_,
...
+pcap_breakloop_dead(pcap_t *p)
+{
+	snprintf(p->errbuf, PCAP_ERRBUF_SIZE,
```
guyharris, line 4217: `pcap_breakloop()` implies there's a loop inside pcap to be broken out of - i.e., that it's possible to capture packets from, or read packets from, the `pcap_t` in question.

It's not possible to capture packets from, or read packets from, a dead `pcap_t`, so `pcap_breakloop()` would be a no-op.  In addition, as `pcap_breakloop()` does not return a value, there's no way to indicate failure.  Therefore, this function should just return and do nothing.

pcap.c
hopper-vul: Oh my bad

pcap.c
hopper-vul: thanks. 

I have fixed that in the next commit.

fxlb:
Rebased.

guyharris:
I'll fix the comments in a subsequent commit.

guyharris:
> I'll fix the comments in a subsequent commit.

Done in 5bb93de447a57e70ade3169748117778f91f8d14.

PR #1148, opened 2023-01-01
Fix NULL pointer dereference at snf_read

egecetin:


infrastation:
(Fixed a typo and rebased on the current master branch.)

infrastation:
Thank you!

egecetin:
One thing I noticed, I added line `CHANGES` at 1.11.x since there was no 1.10.3 when i created PR. But after rebase I saw 1.10.3 section. Should I or you move description line of `CHANGES` to 1.10.3 or is 1.11 is valid?

infrastation:
This is correct as it is now, but when this change propagates into the 1.10 branch later, the entry will be moved accordingly.

guyharris:
What code path can produce this crash?

The only way `snf_read()` gets called is through the `read_op` function pointer in the `pcap_t`, so the only way it gets called is if you have a non-null `pcap_t *` pointing to a `pcap_t` with a `read_op` set to `snf_read`.

I.e., if you pass a null pointer to one of the routines in pcap.c that calls `p->read_op()`, you'll get a null pointer dereference in one of those routines, before it ever gets to `snf_read()`.

So the null-pointer check in `snf_read()` isn't necessary.

infrastation:
Does this mean that the NULL check in `snf_read()` is a no-op?

infrastation:
It does mean that, and my interpretation was based on the assumption that since the check is there, it is possible to have a `NULL` pointer, in which case making this change made sense.

Let's revert the commit then and drop the `if` instead, right?

guyharris:
> Let's revert the commit then and drop the if instead, right?

Sounds good.

I'll see whether, with the right attributes, we can coax GCC/Clang into doing some checks that could say "you could get here with a null pointer even though you don't want that".

egecetin:
So the right fix should be removing the null check since its unnecessary?

infrastation:
That's done now, thanks everyone for your input.

guyharris:
> I'll see whether, with the right attributes, we can coax GCC/Clang into doing some checks that could say "you could get here with a null pointer even though you don't want that".

It appears you can't do so in a fashion that would catch people doing, for example:

    p = pcap_open_live(device, ...);
    pcap_loop(p, 0, callback, data);

which would crash if there's an error opening the device.  All the "argument must not be null" function attribute stuff in GCC and Clang only warns if you explicitly pass a null pointer, not if you pass a pointer that the compiler can't prove is not null (or can prove might be null, based on the function not having the "result is never null" attribute), so it's more of a syntax check of calls rather than having the compiler do static analysis.

mcr:
> which would crash if there's an error opening the device. All the "argument must not be null" function attribute stuff in GCC and Clang only warns if you explicitly pass a null pointer, not if you pass a pointer that the compiler can't prove is not null (or can prove might be null, based on the function not having the "result is never null" attribute), so it's more of a syntax check of calls rather than having the compiler do static analysis.

C is not Rust :-(

Issue #1114, opened 2022-07-06
Haiku build is broken in various ways


infrastation:
Tested using `haiku-r1beta3-x86_64-anyboot.iso` (plus the `llvm9_clang` and `CMake` packages installed via HaikuDepot) and the current master branch at commit ff7024d (after my recent CMake change reverted and a couple cosmetic improvements).

### Autoconf 2.69 + GCC 8.3.0: (3 + 7) warnings and 2 errors
```text
$ MATRIX_CMAKE=no MATRIX_CC=gcc ./build_matrix.sh
Haiku shredder 1 hrev55181+66 Jul  6 2022 08:03: x86_64 x86_64 Haiku
OS identification: Haiku-hrev55181
Wed Jul  6 14:00:17 GMT 2022
PREFIX set to '/tmp/libpcap_build_matrix.dgVOcCTs'
===== SETUP 1: CC=gcc CMAKE=no REMOTE=no =====
$ ./build.sh
gcc (2019_05_24) 8.3.0
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Compiler identification: gcc-8.3.0
$ ./configure --prefix=/tmp/libpcap_build_matrix.dgVOcCTs --enable-remote=no

[...]

$ make -s
cc1plus: warning: command line option '-Wmissing-prototypes' is valid for C/ObjC but not for C++
cc1plus: warning: command line option '-Wpointer-sign' is valid for C/ObjC but not for C++
cc1plus: warning: command line option '-Wstrict-prototypes' is valid for C/ObjC but not for C++
pcap-haiku.cpp: In function 'int pcap_read_haiku(pcap_t*, int, pcap_handler, u_char*)':
pcap-haiku.cpp:55:21: warning: unused variable 'handlep' [-Wunused-variable]
  struct pcap_haiku* handlep = (struct pcap_haiku*)handle->priv;
                     ^~~~~~~
pcap-haiku.cpp:50:37: warning: unused parameter 'maxPackets' [-Wunused-parameter]
 pcap_read_haiku(pcap_t* handle, int maxPackets, pcap_handler callback,
                                 ~~~~^~~~~~~~~~
pcap-haiku.cpp: In function 'int pcap_inject_haiku(pcap_t*, const void*, int)':
[...]

PR #1153, opened 2023-01-16
Haiku: convert implementation to C

davidkaroly:
see #1114

@mmuman

### DIFF
--- a/pcap-haiku.c
+++ b/pcap-haiku.c
@@ -36,12 +36,12 @@ struct pcap_haiku {
 
 
 bool
-prepare_request(struct ifreq& request, const char* name)
+prepare_request(struct ifreq *request, const char* name)
 {
 	if (strlen(name) >= IF_NAMESIZE)
 		return false;
 
-	strcpy(request.ifr_name, name);
+	strcpy(request->ifr_name, name);
 	return true;
 }
 
@@ -93,7 +93,7 @@ pcap_read_haiku(pcap_t* handle, int maxPackets _U_, pcap_handler callback,
 	}
 
 	// fill in pcap_header
-	pcap_pkthdr header;
+	struct pcap_pkthdr header;
 	header.caplen = captureLength;
 	header.len = bytesReceived;
 	header.ts.tv_usec = system_time() % 1000000;
@@ -122,20 +122,20 @@ static int
 pcap_stats_haiku(pcap_t *handle, struct pcap_stat *stats)
 {
 	struct pcap_haiku* handlep = (struct pcap_haiku*)handle->priv;
-	ifreq request;
-	int socket = ::socket(AF_INET, SOCK_DGRAM, 0);
-	if (socket < 0) {
+	struct ifreq request;
+	int pcapSocket = socket(AF_INET, SOCK_DGRAM, 0);
+	if (pcapSocket < 0) {
 		return -1;
 	}
-	prepare_request(request, handlep->device);
-	if (ioctl(socket, SIOCGIFSTATS, &request, sizeof(struct ifreq)) < 0) {
+	prepare_request(&request, handlep->device);
+	if (ioctl(pcapSocket, SIOCGIFSTATS, &request, sizeof(struct ifreq)) < 0) {
 		snprintf(handle->errbuf, PCAP_ERRBUF_SIZE, "pcap_stats: %s",
 			strerror(errno));
-		close(socket);
+		close(pcapSocket);
 		return -1;
 	}
 
-	close(socket);
+	close(pcapSocket);
 	handlep->stat.ps_recv += request.ifr_stats.receive.packets;
 	handlep->stat.ps_drop += request.ifr_stats.receive.dropped;
 	*stats = handlep->stat;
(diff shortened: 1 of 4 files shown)


pcap-haiku.c
```diff
@@ -50,20 +50,23 @@ static int
 pcap_read_haiku(pcap_t* handle, int maxPackets _U_, pcap_handler callback,
 	u_char* userdata)
 {
+	int32_t captureLength;
```
guyharris, line 53: Do these need to be moved?  We require C99 support in the main and 1.10 branches.

pcap-haiku.c
```diff
@@ -50,20 +50,23 @@ static int
...
+	struct pcap_pkthdr header;
+
 	// Receive a single packet
 
 	u_char* buffer = (u_char*)handle->buffer + handle->offset;
 	struct sockaddr_dl from;
 	ssize_t bytesReceived;
 	do {
+		socklen_t fromLength = sizeof(from);
```
guyharris, line 62: Same here.

pcap-haiku.c
davidkaroly: You're right, these don't need to be moved. First I was testing with an older compiler and this was left in here like this.

pcap-haiku.c
davidkaroly: similarly

pcap-haiku.c
guyharris: > I was testing with an older compiler

Using the configure script or CMake?  Both of them should require C99 support, but perhaps they're not testing for declarations in the middle of code.  What compiler was this?

mmuman: If you want to target GCC 2.95 you'll have to do C89 sadly, as it doesn't know about C99. Thinking of it, it might actually be the reason for the file being C++ in the first place, it's a cheap way of allowing C99-type features like variable declarations anywhere.

guyharris: The latest configure script only warns about the lack of C99 features, because, when generated by the 2.71 version of autoconf, it rejects the Solaris Studio 12.3 compiler without bothering to try -xc99, but the compiler handles libpcap anyway.  Maybe that's what happened.

So your choices are 1) leave the file in C++ or 2) require a newer compiler.  (Does Haiku really provide the 2.95 C compiler by default?  If so, does it also provide the 2.95 C++ compiler?)

fxlb: > If you want to target GCC 2.95 you'll have to do C89 sadly, as it doesn't know about C99.

Release History: GCC 2.95.3 March 16, 2001 !

fxlb: Not sure we want to support this rather old version.

infrastation: "It is not old, it is a classic!".

On a more serious note, GCC 2.95 has been the compiler of the 32-bit version of Haiku for a while (Haiku developers probably have a reason for that).

To me it would make a lot of sense to remove C++ from the problem space of libpcap so workarounds like those made in commit 914e40f (thank you Guy) would not be required in the solution space, and C89 does not seem an inconvenience.

mmuman: Yes we do have a reason: we want to be binary compatible with BeOS, which was written in C++, using GCC2, but GCC3 changed their C++ ABI to the Itanium one, and so we are stuck to 2.95 for the system libs.
Now, I'm not sure we want to keep supporting the GCC2 build, unless there's a system lib that depends on it, but [the changes seem to be minimal](https://github.com/haikuports/haikuports/blob/master/net-libs/libpcap/patches/libpcap-1.10.3.gcc2.patchset).

davidkaroly: I think it's only the 32-bit gcc2 build of tcpdump and nmap that relies on the gcc2 build of libpcap. We could migrate them to the gcc 11.2 toolchain and then we can do away with libpcap legacy build. (both libpcap and tcpdump needs a gcc2 patch to get a successful build, mostly variable declarations in multiple files)

infrastation:
Whilst compiler versions and features as applied to the entire source tree are debatable and specific build method can evolve over time, the focus of this change is that it would be a simpler problem to solve without a C++ compiler in scope, or so it seems to me.

pcap-haiku.c
davidkaroly: Probably we don't need to maintain gcc2 legacy toolchain compatibility for libpcap on Haiku as it's possible to migrate dependencies to more up-to-date gcc11 toolchain. On the other hand it wouldbe good to move the Haiku backend to C instead of C++.
Are there any outstanding review comments on this PR that I should follow up on?

infrastation:
My test build confirms that on Haiku R1/beta4 these changes do not break the GCC half of the build matrix and do not fix the Clang half. I am going to merge this anytime soon to unblock subsequent removal of `CXX`isms from the build systems, which would unblock attempts to build using Clang on Haiku.

davidkaroly:
> My test build confirms that on Haiku R1/beta4 these changes do not break the GCC half of the build matrix and do not fix the Clang half. I am going to merge this anytime soon to unblock subsequent removal of `CXX`isms from the build systems, which would unblock attempts to build using Clang on Haiku.

Thanks for the update. 👍 Actually I already have a draft patch for removing C++ from `configure.ac` and `CMakeLists.txt`, I'd like to upload these for review once this PR is merged.

infrastation:
Thank you. When the dust settles, please also add a useful change log entry.

pcap-haiku.c
guyharris: The C language's ABI isn't as complicated as the C++ language's ABI, so doing the 32-bit build of libpcap with a non-antique C compiler might not break anything.

PR #1154, opened 2023-01-17
autoconf: fix searching for libbsd on Haiku

davidkaroly:
After commit 6493919a4c, the defines and library checks for Haiku (87bfc93a7a, 1d24bb8f71) don't have any effect.

Current behavior: Haiku-specific build options are not added to `CFLAGS` and `LIBS`.

Expected behavior - after adding this change:
* `CFLAGS` contains `-D_BSD_SOURCE`
* `LIBS` contains `-lbsd`

How to reproduce: run `./configure` on Haiku, then inspect the variables e.g. in `Makefile`

### DIFF
--- a/configure
+++ b/configure
@@ -6317,6 +6317,59 @@ fi
 	save_LIBS="$LIBS"
 	save_LDFLAGS="$LDFLAGS"
 
+if test "$GCC" = yes ; then
+	    { printf "%s\n" "$as_me:${as_lineno-$LINENO}: checking for ANSI ioctl definitions" >&5
+printf %s "checking for ANSI ioctl definitions... " >&6; }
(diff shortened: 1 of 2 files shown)


guyharris:
Why, in Haiku R1 beta 4, is it necessary to define `_BSD_SOURCE`?  I tried it with a configure script that didn't add `-D_BSD_SOURCE`, and configure and make seemed to work.  Is this a 32-bit-with-old-compiler-for-BeOS-binary-compatbility thing, so it only shows up with gcc 2.95?

davidkaroly:
I can see that a similar change was introduced on master recently so probably this PR can be cancelled.

davidkaroly:
Superseded by 3c7ab964ea and 892caaacca therefore closing this PR.

PR #1158, opened 2023-01-27
gencode: Unify NetBSD and non-NetBSD alignment, and make more general

jrtc27:
Currently the non-NetBSD code assumes that long is the maximum alignment needed for anything allocated from the chunk allocator. In practice, on typical 32-bit and 64-bit Unix platforms, this is true, since all the structures stored in it have only pointers and 32-bit integers. However, CHERI (on 64-bit architectures) uses 128-bit capabilities to implement C language pointers, which are a combination of a traditional 64-bit address, 64 bits of metadata and an additional 129th validity tag bit. Due to the use of tagged memory to support this tag bit they must be stored at naturally aligned addresses, i.e. 16 byte aligned, but long is still only 64-bit, since that is the natural machine word size for integer arithmetic, and thus if sdup tries to allocate a string of between 0 and 7 bytes mod 16 a subsequent allocation of any other type will not be sufficiently aligned for storing a pointer. Technically on 64-bit Windows long is 32-bit and thus insufficient alignment is used but in practice unaligned accesses work even if they are strictly UB.

Fix this by computing the actual alignment requirement for chunks based on the set of types that can be stored in a chunk, without making use of C11's _Alignof. Even if C11 were required, using an explicit list of types rather than max_align_t allows us to be slightly more efficient in memory usage, since there are types (like long double) that force it to be 16 byte aligned that we don't need.

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -232,6 +232,26 @@ struct chunk {
 	void *m;
 };
 
+/*
+ * A chunk can store any of:
+ *  - a string (guaranteed alignment 1 but present for completeness)
+ *  - a block
+ *  - an slist
+ *  - an arth
+ * For this simple allocator every allocated chunk gets rounded up to the
+ * alignment needed for any chunk.
+ */
+struct chunk_align {
+	char dummy;
+	union {
+		char c;
+		struct block b;
+		struct slist s;
+		struct arth a;
+	} u;
+};
+#define CHUNK_ALIGN (offsetof(struct chunk_align, u))
+
 /* Code generator state */
 
 struct _compiler_state {
@@ -600,13 +620,8 @@ newchunk_nolongjmp(compiler_state_t *cstate, size_t n)
 	int k;
 	size_t size;
 
-#ifndef __NetBSD__
-	/* XXX Round up to nearest long. */
-	n = (n + sizeof(long) - 1) & ~(sizeof(long) - 1);
-#else
-	/* XXX Round up to structure boundary. */
-	n = ALIGN(n);
-#endif
+	/* Round up to chunk alignment. */
+	n = (n + CHUNK_ALIGN - 1) & ~(CHUNK_ALIGN - 1);
 
 	cp = &cstate->chunks[cstate->cur_chunk];
 	if (n > cp->n_left) {


guyharris:
(Speaking of CHERI, the ["Beyond the PDP-11: Architectural support for a memory-safe C abstract machine"](https://www.cl.cam.ac.uk/~dc552/papers/asplos15-memory-safe-c.pdf) paper noted in Table 1 that tcpdump casted away `const`ness in 579 places.

This annoyed me, so I looked into it, and they were almost all, if not all, cases where the code was just being lazy and not bothering to declare pointers as `const`.  I spent some time fixing most of them, and later fixed at least some of the remaining ones (a code change made it easier to do so).  Afterwards, I sent an email to Brooks Davis, CCing David Chisnall, saying

> I seem to remember that, in your CHERI work, you had to change the capability model, or the compiler, to handle cases such as tcpdump casting away constness. (It was a while ago; I'm not sure which paper mentioned that.  I might be misremembering.)
> 
> The vast majority of those were cases where the cast could either be removed or changed not to cast away constness, as tcpdump was *not* attempting to modify the data; I was inspired by one of your CHERI papers to fix those.
> 
> There were a couple of cases where tcpdump really *was* overwriting a buffer handed to it with a const pointer; in the current master branch, those have been changed to, instead, allocate a new buffer, which is freed when the packet dissection is complete.
> 
> So, for what it's worth, any such changes should no longer be necessary for tcpdump.
> 
> As for the pointer-subtraction, yes, they exist as part of bounds checking done by tcpdump.  In the master branch, most if not all of the bounds checking is done as part of "fetch packet data" macros, so at least it's easier to write code that does the necessary bounds checking.
> 
> Wireshark handles this by encapsulating packet data into "tvbuffs" that contain a pointer and various lengths, and having the "fetch packet data" code take a pointer to a tvbuff and an offset as arguments; the bounds checking is done on offsets, with various exceptions being thrown (by a setjmp/longjmp-based exception library by Kaz Kyhelku) for different bounds being exceeded (to distinguish, for example, between "that field, which is supposed to be in a packet of this type, isn't present in the packet because the packet is malformed" and "that field isn't present in the packet because the packet was cut short by the capture process").  Perhaps that's the right way to handle that - it might get rid of both the "arbitrary pointer subtraction" and "computation of invalid intermediate results" idioms.

Tcpdump now has something similar to what Wireshark does, in that all fetches of packet data should go through macros or functions that do bounds checking and throw an exception with `longjmp()`, but it still does bounds checks with expressions that involve pointer subtraction.

There's the separate issue of tcpdump being run with root privileges, which should never have to be the case.  Ideally, there should be a way to allow libpcap, running in a completely non-privileged process, to be able to get capture handles, perhaps requiring that 1) the user running the capture program being granted packet-capture privileges, 2) the user running the capture program supplying a password for some administrative account, or 3) the capture program being granted packet-capture privileges, or some combination of those.)

PR #1163, opened 2023-02-22
CMake: install include files in CMAKE_INSTALL_INCLUDEDIR

davidkaroly:
On the current master CMake always installs include files in `include/` folder regardless of `CMAKE_INSTALL_INCLUDEDIR`

This patch changes this behavior to install include files in the folder specified by `CMAKE_INSTALL_INCLUDEDIR`
It does not change default behavior as the default value for `CMAKE_INSTALL_INCLUDEDIR` is `include`


infrastation:
Thank you for suggesting this change. If it fixes a practical use case, it would be useful to mention it in the change log. Also would it make sense to parametrise the following in the same go?
```
CMakeLists.txt:3347:    install(PROGRAMS ${CMAKE_CURRENT_BINARY_DIR}/pcap-config DESTINATION bin)
CMakeLists.txt:3348:    install(FILES ${CMAKE_CURRENT_BINARY_DIR}/libpcap.pc DESTINATION lib/pkgconfig)
```

infrastation:
Let's move this little bit out of the way and make any other changes later as and if required.

PR #1164, opened 2023-03-12
Clean up DECnet address handling.

infrastation:
Let's see if this passes the usual checks.


infrastation:
MSVC 2019 didn't like this:
```
C:\projects\libpcap\nametoaddr.c(757,3): warning C4061: enumerator 'INVALID' in switch of enum '<unnamed-enum-START>' is not explicitly handled by a case label
```
Apparently, the `default` is not good enough.

PR #1177, opened 2023-05-05
Do not have variables be unsigned if they are passed as signed variables

AtariDreams:
The results are implementation defined and confuses the compiler

### DIFF
--- a/etherent.c
+++ b/etherent.c
@@ -157,7 +157,7 @@ pcap_next_etherent(FILE *fp)
 		/* Use 'namesize' to prevent buffer overflow. */
 		namesize = sizeof(e.name) - 1;
 		do {
-			*bp++ = (u_char)c;
+			*bp++ = (char)c;
 			c = getc(fp);
 			if (c == EOF)
 				return (NULL);


mcr (approved):
Seems reasonable to me.

guyharris:
Please rebase this request, as the changes to nametoaddr.c no longer apply.

etherent.c
```diff
@@ -157,7 +157,7 @@ pcap_next_etherent(FILE *fp)
 		/* Use 'namesize' to prevent buffer overflow. */
 		namesize = sizeof(e.name) - 1;
 		do {
-			*bp++ = (u_char)c;
+			*bp++ = (char)c;
```
guyharris, line 160: `c` is an `int` that is the return value of `getc()`.  Section 7.19.7.1 "The `fgetc` function" of C99 indicates that the return value of `fgetc()` will either be a value in the value range for an `unsigned char` or will be `EOF`.

Subsection 7.19.1 "Introduction" of section 7.19 "Input/output `<stdio.h>`" says that `EOF` "expands to an integer constant expression, with type `int` and a negative value", so it's guaranteed *not* to be in the value range of an `unsigned char`.

Section 7.19.7.5 "The `getc` function" indicates that "The `getc` function is equivalent to `fgetc`, except that if it is implemented as a macro, it may evaluate `stream` more than once, so the argument should never be an expression with side effects.


On all the platforms that libpcap (and tcpdump and most if not all of the other code using libpcap) will ever support, that means it's either -1 (I think that's what `EOF` is on all UN\*Xes and on Windows) or it's a value in the range 0 to 255.

To quote section 6.3.1.3"Signed and unsigned integers" of C99:

> When a value with integer type is converted to another integer type other than `_Bool`, if the value can be represented by the new type, it is unchanged.
> Otherwise, if the new type is unsigned, the value is converted by repeatedly adding or subtracting one more than the maximum value that can be represented in the new type until the value is in the range of the new type.
> Otherwise, the new type is signed and the value cannot be represented in it; either the result is implementation-defined or an implementation-defined signal is raised.

Attempting to store an `int` value in the range 128 to 255 into `*bp` will involve converting a value that cannot be represented in the type of `*bp`, i.e. `char`, to that type, so either 1) the result will be implementation-defined or will result in an implementation-defined signal.

Section 6.5.4 "Cast operators" of C99 says, in the "Semantics" subsection:

> Preceding an expression by a parenthesized type name converts the value of the expression to the named type. This construction is called a *cast*. A cast that specifies no conversion has no effect on the type or value of an expression.

which does not appear to indicate that either of the "implementation-defined" parts of the description of the conversion in question are 1) specified as always happening and 2) given a non-implementation-defined specification, so, at most, the cast may end up squelching compiler warnings (which are outside the scope of the standard).

So, unless I've missed something, the only way to avoid implementation-defined behavior here appears to be "reject non-ASCII characters" - just tweaking the case won't do the trick, we'd have to discard values out of the range 0 to 127 or treat them as errors.  I don't know whether we'd want to require that for Ethernet host names, although if we don't know which flavor of "extended ASCII" is being used, that could be a problem.

(There's also the issue of what to do about non-printable characters.)

AtariDreams:
@guyharris Done!

PR #1178, opened 2023-05-07
compiler: parse port numbers in ranges using the standard number parser.

guyharris:
Run them through the same code that's used for numbers in the lexical analyzer, for consistency.


infrastation:
`freebsd-aarch64` has just switched to Clang 16, which required a trivial change to the matrix scripts. If you rebase this change on the latest master branch, the failed build should pass.

PR #1182, opened 2023-05-09
build.sh: remove the check for Clang 15 and 16.

guyharris:
We're working around the warning they produce for yynerrs in the parser being set but not used.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -54,11 +54,6 @@ clang-*/NetBSD-*)
     # 'long' to 'suseconds_t' (aka 'int') [-Wshorten-64-to-32]
     LIBPCAP_TAINTED=yes
     ;;
-clang-1[56].*/*)
-    # grammar.c:1369:14: warning: variable 'pcap_nerrs' set but not used
-    #   [-Wunused-but-set-variable]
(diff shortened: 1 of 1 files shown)


guyharris:
I can haz email?  I don't seem to be getting notifications mailed to me; either that, they're stuck somewhere, or they're trying not to bother me about things I did.

infrastation:
Curris CI and Appveyor don't not send any e-mails. Buildbot sends an e-mail when a particular build changes between passed and failed. As far as it seems to me, in this pull request all Buildbot builds were passing before the change and remained passing after, so there was nothing to notify about.

guyharris:
The notifications in question were from GitHub itself, not from builders, but the problem appears to be that either 1) they were queued up inside GitHub or 2) the process of generating them was stuck, as I eventually got notifications, long after the issues/pull requests were created or had comments made on them.

I just made the comment to see if some change had been made by GitHub where changes whether 1) made by me or 2) made on an issue or pull request that I filed were sent to others but not to me.  As noted, that wasn't the issue.

infrastation:
Judging by the `Received` headers, an e-mail with this comment reached my inbox just after you made it.

PR #1185, opened 2023-05-16
Remove unused variable retval in sock_present2network

AtariDreams:
This quiets the compiler since it is not even returned anyway, and is a misleading variable name.


infrastation:
Thank you. Next time please include details of the warning into the commit message (as was done in commit 31bb18a and elsewhere).

PR #1187, opened 2023-05-16
Remove redundant assigning of p1 to prev

AtariDreams:
This assignment is redundant, as the assignment to prev from p1 happens after the if statement anyway.

This was found by clang-tidy emitting the following warning: "Value stored to 'prev' is never read", as prev is reassigned to the same expression once outside the if-statement.

### DIFF
--- a/pcap-sita.c
+++ b/pcap-sita.c
@@ -533,7 +533,7 @@ static void sort_if_table(void) {
 
 	if (!acn_if_list) return;				/* nothing to do if the list is empty */
 
-	while (1) {
+	do {
 		p1 = acn_if_list;					/* start at the head of the list */
 		prev = 0;
 		has_swapped = 0;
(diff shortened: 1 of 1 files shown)


fxlb:
> This fixes compiler warnings [...]

A copy of the warning message should be added in the body of the commit message with the compiler name and version.

infrastation:
I wonder why `make -s CFLAGS=-Werror` didn't produce this warning using Clang 16.0.1. Did you use custom `CFLAGS` or anything else to get this warning?

fxlb:
A forced build? Because doc/README.sita states:
```
NOTE: this is not currently supported; the configure script doesn't
support --with-sita, and CMake doesn't support enabling SITA ACN
support.  The code currently does not compile; it should really be
implemented as an additional remote capture mechanism, using a URL,
rather than as a separate version of libpcap that supports only the ACN
product, but the infrastructure for that isn't yet available.
```
Not sure there is a need to take care of this source file.

infrastation:
Oh, the warning is in that file, not in the one I thought.

AtariDreams:
@fxlb Any updates on this? I get 'Value stored to 'p1' is never read' from clang-tidy.

infrastation:
Trying to compile `pcap-sita.c` yields 15 warnings and 2 errors. The redundant assignment is the least of its problems. Let's not waste time on this file until someone is ready to fix and test it or it is clear it has no purpose and can be retired.

PR #1188, opened 2023-05-16
Make i, n, index, and max_stmts unsigned integers

AtariDreams:
This fixes compiler warnings  regarding signedness and avoids casting.

maxval should be an unsigned integer because slength, the result of which is added to this number, returns an unsigned value.

In addition, opt_state->maxval being an unsigned int that is assigned via multiplying max_stmts is also a good case for max_stmts being unsigned. Finally, count_blocks should be unsigned since the minimum number is 0, and it assigns to n, which "i" is compared to, and is used for pointer arithmetic.

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -604,7 +604,7 @@ static struct block *gen_msg_abbrev(compiler_state_t *, int type);
 static void
 initchunks(compiler_state_t *cstate)
 {
-	int i;
+	u_int i;
 
 	for (i = 0; i < NCHUNKS; i++) {
 		cstate->chunks[i].n_left = 0;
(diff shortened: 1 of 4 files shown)


infrastation:
Please provide more details of the problem.

infrastation:
The main problem with the proposed change is that it does not quote any evidence of the compiler warnings it fixes. The only way to compile `pcap-dag.c`, is to have the DAG library, so it is a bit difficult to reason about the file. However, `gencode.c`, `optimize.c` and `rpcapd/rpcapd.c` are compiled in every CI run and do not currently generate any compiler warnings because libpcap does not currently generate any compiler warnings.

Therefore closing.

PR #1195, opened 2023-06-01
RFC: DSA/EDSA: Enable filter functionality

Hexxel:
This is a RFC, we want to use tcpdump filter functionality on a DSA/EDSA link layer device, but there doesn't seem to be supported right now. I'm not well versed in the libpcap impelmentation so I naively looked at the existing code and did the following changes in this PR. Please don't hit me over the head :/
Is this the right way of doing it or do you have any suggestions where I should look for a better solution/approach?

Best regards,

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -1717,6 +1717,20 @@ init_linktype(compiler_state_t *cstate, pcap_t *p)
 		cstate->off_nl_nosnap = 3;	/* 802.3+802.2 */
 		break;
 
+	case DLT_DSA_TAG_DSA:
+		cstate->off_linktype.constant_part = 16;
+		cstate->off_linkpl.constant_part = 18;	/* Ethernet header length */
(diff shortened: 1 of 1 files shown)


Hexxel:
No actions on this PR?

infrastation:
This will need a change log entry. A few filter tests would be useful too.

infrastation:
If these changes are applied and `off_nl_nosnap` is set to `OFFSET_NOT_SET` and the DLT contains an Ethernet packet (with a commensurate switch tag) that is a 802.2 LLC packet, will everything work as usual?

Hexxel:
I really don't know, sorry.

infrastation:
Alright, let's think about it.

infrastation:
As it turns out, LLC indeed may be present and there is more code paths to take into account. Commit 588655e implements `DLT_DSA_TAG_DSA` filtering and adds tests.

PR #1196, opened 2023-06-11
Prefer calloc over malloc + memset where possible

AtariDreams:
This offloads the memset work to the OS.

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -628,12 +628,11 @@ newchunk_nolongjmp(compiler_state_t *cstate, size_t n)
 			return (NULL);
 		}
 		size = CHUNK0SIZE << k;
-		cp->m = (void *)malloc(size);
+		cp->m = calloc(1, size);
 		if (cp->m == NULL) {
 			bpf_set_error(cstate, "out of memory");
 			return (NULL);
 		}
-		memset((char *)cp->m, 0, size);
 		cp->n_left = size;
 		if (n > size) {
 			bpf_set_error(cstate, "out of memory");
--- a/pcap-bpf.c
+++ b/pcap-bpf.c
@@ -2672,18 +2672,15 @@ pcap_activate_bpf(pcap_t *p)
 #ifdef HAVE_ZEROCOPY_BPF
 	if (!pb->zerocopy) {
 #endif
-	p->buffer = malloc(p->bufsize);
+	/* For some strange reason, calloc instead of malloc seems to prevent the EFAULT
+	 * problems we have experienced from AIX BPF. */
+	p->buffer = calloc(1, p->bufsize);
 	if (p->buffer == NULL) {
 		pcap_fmt_errmsg_for_errno(p->errbuf, PCAP_ERRBUF_SIZE,
 		    errno, "malloc");
 		status = PCAP_ERROR;
 		goto bad;
 	}
-#ifdef _AIX
-	/* For some strange reason this seems to prevent the EFAULT
-	 * problems we have experienced from AIX BPF. */
-	memset(p->buffer, 0x0, p->bufsize);
-#endif
 #ifdef HAVE_ZEROCOPY_BPF
 	}
 #endif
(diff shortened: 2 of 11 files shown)


pcap-bpf.c
```diff
@@ -2611,18 +2611,13 @@ pcap_activate_bpf(pcap_t *p)
...
 		pcap_fmt_errmsg_for_errno(p->errbuf, PCAP_ERRBUF_SIZE,
 		    errno, "malloc");
 		status = PCAP_ERROR;
 		goto bad;
 	}
-#ifdef _AIX
-	/* For some strange reason this seems to prevent the EFAULT
-	 * problems we have experienced from AIX BPF. */
-	memset(p->buffer, 0x0, p->bufsize);
-#endif
```
infrastation, line 2625: It would be better to save this comment even after the zeroing is no longer conditional.

pcap-dag.c
```diff
@@ -1274,9 +1274,9 @@ dag_get_datalink(pcap_t *p)
 	int index=0, dlt_index=0;
 	uint8_t types[255];
 
-	memset(types, 0, 255);
+	memset(types, 0, sizeof(types));
```
infrastation, line 1277: `uint8_t types[255] = { 0 };` would fit the problem space better.

pcap-netfilter-linux.c
```diff
@@ -612,7 +612,7 @@ netfilter_activate(pcap_t* handle)
 
 	if (type == NFLOG) {
 		handle->linktype = DLT_NFLOG;
-		handle->dlt_list = (u_int *) malloc(sizeof(u_int) * 2);
+		handle->dlt_list = (u_int *) malloc(2 * sizeof(u_int));
 		if (handle->dlt_list != NULL) {
```
infrastation, line 616: I wonder if this should fail with an error as is done below. In any case, this change is out of scope of this commit.

pcap.c
```diff
@@ -551,8 +551,8 @@ pcap_list_tstamp_types(pcap_t *p, int **tstamp_typesp)
 		**tstamp_typesp = PCAP_TSTAMP_HOST;
 		return (1);
 	} else {
-		*tstamp_typesp = (int*)calloc(sizeof(**tstamp_typesp),
-		    p->tstamp_type_count);
+		*tstamp_typesp = (int*)malloc(p->tstamp_type_count *
+			sizeof(**tstamp_typesp));
```
infrastation, line 555: This is the exact reverse of the point of this commit. Removing this initialization has potential to introduce a bug.

pcap.c
```diff
@@ -3016,14 +3016,14 @@ pcap_list_datalinks(pcap_t *p, int **dlt_buffer)
 		**dlt_buffer = p->linktype;
 		return (1);
 	} else {
-		*dlt_buffer = (int*)calloc(sizeof(**dlt_buffer), p->dlt_count);
+		*dlt_buffer = (int*)malloc(p->dlt_count * sizeof(**dlt_buffer));
```
infrastation, line 3019: Same as above, this change would introduce an unpleasant surprise.

gencode.c
```diff
@@ -632,12 +632,11 @@ newchunk_nolongjmp(compiler_state_t *cstate, size_t n)
 			return (NULL);
 		}
 		size = CHUNK0SIZE << k;
-		cp->m = (void *)malloc(size);
+		cp->m = (void *)calloc(1, size);
```
infrastation, line 635: Both `malloc()` and `calloc()` return a `void *`, so the type cast is noise.

pcap.c
AtariDreams: I have no idea how that got there.

pcap.c
AtariDreams: Not really, because memset

pcap.c
AtariDreams: memset takes care of that

AtariDreams:
@infrastation Done!

infrastation:
The most recent revision still replaces `calloc()` with `malloc()` without `memset()` twice: in `pcap_list_tstamp_types()` and in `pcap_list_datalinks()`. Excuse me, but we are talking in circles and these changes do not look safe, therefore closing.

PR #1197, opened 2023-06-13
pcap-filter(7): Add the filter "ifindex interface_index"

fxlb:
[skip ci]


infrastation:
It would be useful to say in "backward compatibility" that this was introduced in libpcap 1.10.0.

fxlb:
"backward compatibility"

Yes! Done.

PR #1199, opened 2023-06-21
bpf_bind() should parse link/interface specification to obtain zoneid

Sashan:
currently link/interface name is parsed in pcap_activate_bpf(), however zone-id is not passed down to bpf_bind() function. This prevents libpcap consumers on Solaris (for example tcpdump) to attach to link owned by non-global zone. For example:
	tcpdump -i test_zone/net0
in this case libpcap always opens link net0 which (if exists) is owned by global zone. However we want tcpdump which runs in global zone to start sniffing on net0 owned by test_zone

the proposed fix moves parsing down to bpf_bind(). the drawback here is we are losing some error messages.

### DIFF
--- a/pcap-bpf.c
+++ b/pcap-bpf.c
@@ -635,14 +635,56 @@ static int
 bpf_bind(int fd, const char *name, char *errbuf)
 {
 	int status;
-#ifdef LIFNAMSIZ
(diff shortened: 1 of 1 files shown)


fxlb:
Please cancel `Merge branch 'master' into solaris.ngz`.
Do a `rebase` on `the-tcpdump-group:master` for a better history.

PR #1200, opened 2023-06-24
Document why we do not do error handling if dlt_list is NULL

AtariDreams:
We can leave the dlt_list empty instead of populating it.


guyharris:
> We can leave the dlt_list empty instead of populating it.

[citation needed]

AtariDreams:
> > We can leave the dlt_list empty instead of populating it.
> 
> 
> 
> [citation needed]

Everywhere else in the code does this

guyharris:
> Everywhere else in the code does this

As of 8f1f6fcbe4b4c17d965b42374a2f170452a93289, nowhere in the code does this.  And even before that, there were places in the code that didn't.

But this did get me to review the code and test some things, and I found some unrelated bugs and fixed them, so thanks.

PR #1203, opened 2023-07-06
Allow build on OS X pre 10.5

sevan:
Availability.h was introduced in 10.5 (Leopard)
Guard off the inclusion so that it is possible to build libpcap on older releases such as 10.4 (Tiger)

### DIFF
--- a/pcap/funcattrs.h
+++ b/pcap/funcattrs.h
@@ -181,7 +181,11 @@
  * I've never seen earlier releases.
  */
 #ifdef __APPLE__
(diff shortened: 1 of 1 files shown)


guyharris:
I removed the include of <Availability.h> entirely in ff91e1b1270d99cceab96eb851ea98510e9981d4 (with a comment fixup in bdfd4f4da1a2b4a21ca8fab718831259edb82d75); it's no longer needed as a hint to Apple.

PR #1209, opened 2023-07-26
MSDOS packet driver interface is no longer testable

mcr:
Anyone who really needs it can use a previous version


fxlb:
FYI:
$ git grep 'msdos/'         
```
.gitignore:/msdos/.depend.dj
.gitignore:/msdos/bin2c.exe
.gitignore:/msdos/pkt_rx1.lst
.gitignore:/msdos/pkt_stub.inc
INSTALL.md: msdos/*         - drivers for MS-DOS capture support
Makefile.in:    msdos/bin2c.c \
Makefile.in:    msdos/makefile \
Makefile.in:    msdos/makefile.dj \
Makefile.in:    msdos/makefile.wc \
Makefile.in:    msdos/pkt_rx0.asm \
Makefile.in:    msdos/pkt_rx1.s \
Makefile.in:    msdos/pktdrvr.c \
Makefile.in:    msdos/pktdrvr.h \
Makefile.in:    msdos/readme.dos \
pcap-dos.c:  #include "msdos/pm_drvr/pmdrvr.h"
pcap-dos.c:  #include "msdos/pm_drvr/pci.h"
pcap-dos.c:  #include "msdos/pm_drvr/bios32.h"
pcap-dos.c:  #include "msdos/pm_drvr/module.h"
pcap-dos.c:  #include "msdos/pm_drvr/3c501.h"
pcap-dos.c:  #include "msdos/pm_drvr/3c503.h"
pcap-dos.c:  #include "msdos/pm_drvr/3c509.h"
pcap-dos.c:  #include "msdos/pm_drvr/3c59x.h"
pcap-dos.c:  #include "msdos/pm_drvr/3c515.h"
pcap-dos.c:  #include "msdos/pm_drvr/3c90x.h"
pcap-dos.c:  #include "msdos/pm_drvr/3c575_cb.h"
pcap-dos.c:  #include "msdos/pm_drvr/ne.h"
pcap-dos.c:  #include "msdos/pm_drvr/wd.h"
pcap-dos.c:  #include "msdos/pm_drvr/accton.h"
pcap-dos.c:  #include "msdos/pm_drvr/cs89x0.h"
pcap-dos.c:  #include "msdos/pm_drvr/rtl8139.h"
pcap-dos.c:  #include "msdos/pm_drvr/ne2k-pci.h"
pcap-dos.c:#include "msdos/pktdrvr.h"
pcap-dos.c:#include "msdos/ndis2.h"
pcap-dos.c:#include "msdos/pm_drvr/ioport.h"
pcap-dos.c:#include "msdos/pm_drvr/dma.h"
pcap-dos.h:  #include "msdos/pm_drvr/lock.h"
```

mcr:
I have removed pcap-dos.c and INSTALL.md, but I left a note about when pcap-dos went away.
I did not touch config.guess and config.sub, because I think that they ultimate come from autoconf sources.

guyharris:
> I did not touch config.guess and config.sub, because I think that they ultimate come from autoconf sources.

Correct.

fxlb:
`removed in 2023, after libpcap 1.9`
Should be
`removed in 2023, after libpcap 1.10.4`

fxlb:
Remain:
```
.gitignore:64:/msdos/.depend.dj
.gitignore:65:/msdos/bin2c.exe
.gitignore:66:/msdos/pkt_rx1.lst
.gitignore:67:/msdos/pkt_stub.inc
```

fxlb:
Also:
```
gencode.c:46:#ifdef MSDOS
gencode.c-47-#include "pcap-dos.h"
gencode.c-48-#endif
```
```
pcap.c:74:#ifdef MSDOS
pcap.c-75-#include "pcap-dos.h"
pcap.c-76-#endif
```
and other stuff with MSDOS.

infrastation:
It is fine just to delete `pcap-dos.*` from `INSTALL.md`. I was going to update `README.md` here and possibly in tcpdump to have a complete list of OSes that were supported in the past.

mcr:
> Remain:
> 
> ```
> .gitignore:64:/msdos/.depend.dj
> .gitignore:65:/msdos/bin2c.exe
> .gitignore:66:/msdos/pkt_rx1.lst
> .gitignore:67:/msdos/pkt_stub.inc
> ```

I left them there on purpose, in case someone wants to patch stuff back, and gets confused about the generated files.
We can remove them too if they bother you.

mcr:
> and other stuff with `MSDOS`.

A bunch of stuff is !defined(MSDOS).  I feel kinda like we should leave it as documentation, but probably it should go.

mcr:
What about HAVE_PACKET32?
Can those checks be removed from pcap.c?

mcr:
> > and other stuff with `MSDOS`.
> 
> A bunch of stuff is !defined(MSDOS). I feel kinda like we should leave it as documentation, but probably it should go.

okay, I've pushed a third commit removing all of it.  Made a mistake on pcap-int.h but fixed that.
I will squash it all after review.

mcr:
> It is fine just to delete `pcap-dos.*` from `INSTALL.md`. I was going to update `README.md` here and possibly in tcpdump to have a complete list of OSes that were supported in the past.

when you update README.md, then delete that line please.
I'd like to leave it for posterity.

guyharris:
> What about HAVE_PACKET32?

That's testing for the packet32 library from WinPcap/Npcap

The one packet it's used is in deciding whether to use pcap.c's `pcap_lookupdev()`; the comment is:

```
 * pcap-npf.c has its own pcap_lookupdev(), for compatibility reasons, as
 * it actually returns the names of all interfaces, with a NUL separator  
 * between them; some callers may depend on that.
 * 
 * MS-DOS has its own pcap_lookupdev(), but that might be useful only
 * as an optimization.
 * 
 * In all other cases, we just use pcap_findalldevs() to get a list of  
 * devices, and pick from that list.
```

So it should still test `HAVE_PACKET32`.  If you're building on Windows and *don't* have packet32 (which really means "don't have the SDK"), that means you can't build (packet32/NPF-based) packet capture support, and can't compile packet-npf.c, so you need pcap.c's `pcap_lookupdev()`.

INSTALL.md
```diff
@@ -281,8 +280,7 @@ in `/usr/include/sys/dlpi.h`, and find the corresponding value.
 	pcap-dbus.c	    - D-Bus capture support
 	pcap-dbus.h	    - D-Bus capture support
 	pcap-dlpi.c	    - Data Link Provider Interface support
-	pcap-dos.c	    - MS-DOS capture support
-	pcap-dos.h	    - headers for MS-DOS capture support
+	pcap-dos.c	    - removed in 2023, after libpcap 1.10
```
guyharris, line 283: Should we do that for the files removed in the past?

And should the list of "were here, but aren't here any more" files be a separate list?

pcap-int.h
```diff
@@ -293,7 +288,6 @@ struct pcap {
 	 */
 	int bpf_codegen_flags;
 
-#if !defined(_WIN32) && !defined(MSDOS)
```
guyharris, line 296: This should be either `#ifndef _WIN32` or `#if !defined(_WIN32)`, as Windows doesn't have UN\*X-style file descriptors on which you can do stuff such as `select()`, except for sockets, and libpcap doesn't offer a "get selectable descriptor" call on Windows for that reason.

pcap-int.h
```diff
@@ -308,7 +302,6 @@ struct pcap {
 	 * prepared not to see any packets from the attempt.
 	 */
 	const struct timeval *required_select_timeout;
-#endif
```
guyharris, line 311: Ditto.

INSTALL.md
infrastation: Only if there is a good reason, it seems to me. For example, files in `missing/` come and go, but the world does not need to know so long as the source compiles and works correctly.

pcap-int.h
mcr: like; e80574e2 ?

guyharris: Yes.

PR #1213, opened 2023-08-16
Add DLT and linktype for MDB

martin-kaiser:
Add a link-layer header type for the MDB (Multi-Drop Bus) protocol.

The MDB protocol defines messages between a vending machine controller and peripherals inside the vending machine, e.g. a card reader that processes payments.

Packets with DLT_MDB use a pseudo-header as defined in

https://www.kaiser.cx/pcap-mdb.html

### DIFF
--- a/pcap-common.c
+++ b/pcap-common.c
@@ -1269,7 +1269,19 @@
  */
 #define LINKTYPE_FIRA_UCI	299
 
-#define LINKTYPE_HIGH_MATCHING_MAX	299		/* highest value in the "matching" range */
+/*
+ * MDB (Multi-Drop Bus) protocol between a vending machine controller and
+ * peripherals inside the vending machine. See
+ *
+ *	https://www.kaiser.cx/pcap-mdb.html
+ *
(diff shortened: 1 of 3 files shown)


martin-kaiser:
Dear all,

I sent a mail to tcpdump-workers for requesting the new DLT about a week ago. It seems that the list is down.

----------------------

Dear all,

I would like to request a DLT value for the MDB protocol.

The MDB protocol defines messages between a vending machine controller
and peripherals inside the vending machine, e.g. a card reader that
processes payments.

As usual for such protocols, we need a pseudo-header before the actual
data that was captured. I defined the pseudo-header on

https://www.kaiser.cx/pcap-mdb.html

and added some background info and examples.

You can link to this page from https://www.tcpdump.org/linktypes.html.

Thanks,
Martin

mcr (approved):

guyharris (approved):

infrastation:
For posterity, `LINKTYPE_MDB` became available in libpcap 1.10.6.

PR #1228, opened 2023-10-24
Prefix functions declared in pcap-int.h with pcapint_

fxlb:
Keep pcap_ prefix for external functions.

It is a follow-up to commit d14fb11d0dfc06ee5dac58146497aa5b0aa7c6fa.


fxlb:
Any objections?

infrastation:
Looks straightforward to me, in that it makes the external and the internal surfaces easy to tell apart. Maybe a change log entry would be due.

fxlb:
> Maybe a change log entry would be due.

Yes, an update of the pevious one, before a backport.

PR #1229, opened 2023-11-08
More pcap_ to pcapint_ for non-API routines.

guyharris:


missing/strlcpy.c
```diff
@@ -1,4 +1,4 @@
-/*	$OpenBSD: pcap_strlcpy.c,v 1.12 2015/01/15 03:54:12 millert Exp $	*/
+/*	$OpenBSD: strlcpy.c,v 1.12 2015/01/15 03:54:12 millert Exp $	*/
```
infrastation, line 1: Search and replace misfired here (possibly somewhere else too).

guyharris: Nope, that's a fix to an earlier search-and-replace from a while ago that shouldn't have happened in the first place; we (somebody, probably me) grabbed the OpenBSD strlcpy.c, renamed the routine, and added it.  I fixed the /* OpenBSD */ comment to use the original source file's name.

infrastation: I stand corrected.

guyharris: It was probably my screwup in the first place, so don't worry about it.

scanner.l
```diff
@@ -658,6 +658,7 @@ stou(const char *yytext_arg, YYSTYPE *yylval_arg, compiler_state_t *yyextra_arg)
 {
 	stoulen_ret ret;
 
+fprintf(stderr, "%s\n", yytext_arg);
```
fxlb, line 661: Should be removed from this PR.

guyharris: Yeah, that was me trying to figure out what that memory sanitizer issue was about.  Removed.

fxlb: I had the same...

mcr (approved):

Issue #1233, opened 2023-11-24
Resource leak caused by the mismatch of pcap_create and pcap_close


PromptFuzz:
Hi,
 I have found that if call `pcap_close()` intermediatly after the `pcap_create()` could cause a eventfd leak.

The example code is:
```
    if (pcap_findalldevs(&devs, errbuf) == -1) {
     ...
    }
    // Open first device for capturing
    handle = pcap_create(devs->name, errbuf);
    //pcap_activate(handle);
    pcap_close(handle);   
```

My environment :
OS: CentOS 5.4
Run with non-privileged user account

Rerpoduce:
In 
[poc.zip](https://github.com/the-tcpdump-group/libpcap/files/13456803/poc.zip).

<img width="691" alt="fd_leak" src="https://github.com/the-tcpdump-group/libpcap/assets/142886726/6d548f37-b863-491a-9845-91203cd429db">

I have noticed that [pcap_create](https://www.tcpdump.org/manpages/pcap_create.3pcap.html) has announced:
>The returned handle must be activated with [pcap_activate](https://www.tcpdump.org/manpages/pcap_activate.3pcap.html)(3PCAP) before packets can be captured with it;

Actually, the leak was disappeared if i called `pcap_activate(handle);` after the `pcap_create()`.
But, for general use, the `pcap_create` and `pcap_close` should be match, regardless whether we use or not use the returned handler.

PromptFuzz:
I have provided a PR to fix this issue: #1235

guyharris:
> But, for general use, the pcap_create and pcap_close should be match, regardless whether we use or not use the returned handler.

Yes, the eventfd shouldn't be opened until we need it, so it should be opened in the activate routine, not the create routine.  See pull request #1236.

guyharris:
Fixed in c62d3d3ea880e0822e1973415556fe04f03041c2.

PR #1235, opened 2023-11-24
Fix resource leak caused by mismatch between pcap_create and pcap_close

PromptFuzz:
See issue: #1233
This commit initializes the cleanup_op callback of handler in pcap_create.
This avoid the resource leak caused by the mismatch of resource allocation and deallocation.


guyharris:
Another fix might be to move the allocation of the event FD into `pcap_activate_linux()` - it's not necessary unless the handle is activated and capturing is possible on it.

I'll see whether it was originally done in the activate routine and moved there for a reason, in which case the reason should be noted in a comment, or originally done in the create routine when the eventfd stuff was added, in which case I'll move it to the activate routine.

guyharris:
Fixed in #1236.

Issue #1239, opened 2023-11-28
pcap_create could crash by passing null errbuf


PromptFuzz:
Hi,
  when fuzzing the libpcap, I found pcap_create could be easily crashed when passing it with a NULL errbuf.
 
Your documentation in [pcap_create](https://www.tcpdump.org/manpages/pcap_create.3pcap.html) haven't said that the errbuf cannot be a NULL pointer.
So I just assume that if no custom error buf is provided, the error message will be delivered to stderr or other places.

However, after I fuzzed the `pcap_create("eth0", nullptr)` for some rounds until some operations could cause the `pcap_create` failed, the `pcap_create` will crash if it was passed with NULL errbuf.

My environment: 
CentOS 5.4

The PoC program is like that:
```
extern "C" int LLVMFuzzerTestOneInput_4(const uint8_t* data, size_t size) {
	
    pcap_t *handle = pcap_create("eth0", nullptr);
    ...//some operations to change the environment
    pcap_close(handle);
    return 0;
}
```

```
==707630==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7fc9bb8d0f81 bp 0x000000000000 sp 0x7ffeeb206180 T0)
==707630==The signal is caused by a WRITE memory access.
==707630==Hint: address points to the zero page.
    #0 0x7fc9bb8d0f81  (/lib/x86_64-linux-gnu/libc.so.6+0x8bf81) (BuildId: 1878e6b475720c7c51969e69ab2d276fae6d1dee)
    #1 0x55c4d98bcb1a in vsnprintf /work/llvm/llvm-project/compiler-rt/lib/asan/../sanitizer_common/sanitizer_common_interceptors.inc:1746:1
    #2 0x55c4d9a8037c in pcapint_vfmt_errmsg_for_errno /libpcap/src/libpcap/fmtutils.c:287:8
    #3 0x55c4d9a8020f in pcapint_fmt_errmsg_for_errno /libpcap/src/libpcap/fmtutils.c:275:2
    #4 0x55c4d9a23e6a in iface_get_ts_types /libpcap/src/libpcap/pcap-linux.c
    #5 0x55c4d9a23e6a in pcapint_create_interface /libpcap/src/libpcap/pcap-linux.c:354:6
    #6 0x55c4d99fcffc in pcap_create /libpcap/src/libpcap/pcap.c:2412:6
[...]

PR #1244, opened 2023-12-01
Add some more compile-time checks with __attribute__.

guyharris:
Mark all errbuf arguments as "must not be null".  Perhaps the warnings will prevent people from filing bugs such as #1239.

For most routines that return a success/error code value, warn if the result isn't used.  Perhaps this will reduce the number of "why isn't my code working?" questions, for the cases where the programmer doesn't, for example, bother to check whether pcap_activate() succeeds.

Fix up some test programs to do the checks that they weren't doing in the past.

### DIFF
--- a/pcap/funcattrs.h
+++ b/pcap/funcattrs.h
@@ -306,6 +306,41 @@
 #endif
 
 /*
+ * PCAP_NONNULL(...), after a function declaration, means "the arguments
+ * whose ordinal numbers are listed are pointer arguments that must be
+ * non-null".
+ */
+#if __has_attribute(nonnull) \
+    || PCAP_IS_AT_LEAST_GNUC_VERSION(3,3) \
+    || PCAP_IS_AT_LEAST_XL_C_VERSION(10,1)
+  /*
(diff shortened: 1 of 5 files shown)


fxlb:
With this change, we have now (GCC on Debian stable):
```
./tcpdump.c: In function 'main':
./tcpdump.c:2323:40: warning: ignoring return value of 'pcap_set_datalink' declared with attribute 'warn_unused_result' [-Wunused-result]
 2323 |                                 (void) pcap_set_datalink(pd, DLT_LINUX_SLL2);
      |                                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```

guyharris:
the-tcpdump-group/tcpdump#1104 should fix this; it appears that, while Clang recognizes the `(void)` as meaning "trust us, we're ignoring the return value for a perfectly good reason", GCC (or, at least, the one in Debian stable) doesn't and requires an explicit suppression of the warning.

infrastation:
Eventually this change should have a good change log entry because it is visible to the users.

PR #1251, opened 2023-12-23
Don't define SOCKET ourselves.

guyharris:
Doing so can cause namespace collisions, as per, for example, https://github.com/firnsy/barnyard2/issues/245.

Instead, define PCAP_SOCKET, and use that.

On Windows, this shouldn't break any APIs, as any code using the two active-mode APIs that return a socket handle would probably be assigning their return values to a SOCKET, and as, on Windows, we're defining PCAP_SOCKET as SOCKET, there would be no type clash.

On UN\*Xes, this might break code that uses one of the two active-mode APIs that return a socket handle if those programs were written to assign the return values of those APIs to a SOCKET, as we're no longer defining SOCKET.  However, as those APIs are only provided if libpcap is built with remote capture support - which is not the default - and as they're somewhat painful to use, there's probably little if any code that needs to compile for UN*X and that uses them.  If there *is* any such code, it could do

    #ifndef PCAP_SOCKET
        #ifdef _WIN32
            #define PCAP_SOCKET SOCKET
        #else
            #defube PCAP_SOCKET int
        #endif
    #endif

and use PCAP_SOCKET.

### DIFF
--- a/pcap-rpcap-int.h
+++ b/pcap-rpcap-int.h
@@ -35,7 +35,7 @@
 #define __PCAP_RPCAP_INT_H__
 
 #include "pcap.h"
-#include "sockutils.h"	/* Needed for some structures (like SOCKET, sockaddr_in) which are used here */
+#include "sockutils.h"	/* Needed for some data types (such as PCAP_SOCKET, sockaddr_in) that are used here */
 
 /*
  * \file pcap-rpcap-int.h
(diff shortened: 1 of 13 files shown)


guyharris:
@bonsaiviking as per the above, I don't think this will break any programs that build with Npcap on Windows; does this change look OK?

guyharris:
@bonsaiviking as per the above, I don't think this will break any programs that build with Npcap on Windows; does this change look OK?

PR #1262, opened 2024-01-26
Autoconf: Add autogen.sh, remove configure and config.h.in

fxlb:
Put autoconf-generated files in the release tarball.

Remove the corresponding task from the TODO file.

(And associated commits.)


fxlb:
Backported from master.

infrastation:
I thought the plan was not to backport this into the current stable branches.

fxlb:
In 1 year, there have been no problems in master with this change.
The fact that this change is only in master generates a systematic error when doing a checkout from 1.10 branch to master:
```
$ make
[...]
config.status: error: cannot find input file: `config.h.in'
make: *** [Makefile:921: Makefile] Error 1
```
This implies running `./autogen.sh`.
(Same with checkout from 4.99 branch to master for tcpdump.)
Doing frequent checkout 1.10 <-> master is needed for cherry-picks.
To allow smooth checkout 1.10 <-> master this change is required.

infrastation:
That's a fair point, it is indeed time to consider it again.

Another thing we discussed about this porting last time was completing the cleanups in the master branches before porting the entire thing to the stable branches. In this sense, as far as I gather, the most recent tcpslice is almost clean (there is still a workaround for SunOS 4 make in `Makefile.in`), tcpdump and libpcap in addition to that still have vestige of OSF/1, IRIX, Ultrix and SINIX in Autoconf, CMake and C code.

One particularly convoluted bit is `AC_LBL_LIBRARY_NET`, it can probably remain a while longer. Most of other occurrences look straightforward, I think it would be a good idea to remove from the problem space what can be safely and quickly removed. If you don't mind, I could make some of these changes before the end of this week.

fxlb:
> If you don't mind, I could make some of these changes before the end of this week.

No problem. This merge can wait one week...

infrastation:
I now remember it better: the plan for old OS support was to leave it in libpcap 1.10.x as it is, and to remove it from the master branch.  Which is still a work in progress, and which would not be blocked by introduction of `autogen.sh` in this change. With regard to the latter the only point I have is that it seems to duplicate the function of `AC_PREREQ([2.69])`, but I can live with that.

fxlb:
> duplicate the function of `AC_PREREQ([2.69])`

The two are about minimum version of Autoconf.
`AC_PREREQ` is the standard way to ensure that a recent enough version of Autoconf is being used:
```
configure.ac:16: error: Autoconf version 2.69 or higher is required
```

`autogen.sh` provides more information on non-standard configurations:
```Please install the 'autoconf' package version 2.69 or later.
If version 2.69 or later is already installed and there is no
autoconf default, it may be necessary to set the AUTORECONF
environment variable to enable the one to use, like:
AUTORECONF=autoreconf-2.69 ./autogen.sh
or
AUTORECONF=autoreconf-2.71 ./autogen.sh
```

PR #1265, opened 2024-01-30
Retire support for HP-UX before 10.20.

infrastation:
The changes look correct to me, but additional proof-reading would be appreciated. I don't have means to test the remaining HP-UX support.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3006,19 +3006,16 @@ if(CMAKE_SYSTEM_NAME STREQUAL "AIX")
         set(LIBS_PRIVATE "${LIBS_PRIVATE} -lodm -lcfg")
     endif()
 elseif(CMAKE_SYSTEM_NAME STREQUAL "HP-UX")
-    if(CMAKE_SYSTEM_VERSION MATCHES "[A-Z.]*9\.[0-9]*")
+    if(
+       CMAKE_SYSTEM_VERSION MATCHES "[A-Z.]*9\.[0-9]*"
+       OR CMAKE_SYSTEM_VERSION MATCHES "[A-Z.]*10\.[01]"
+    )
         #
         # HP-UX 9.x.
-        #
-        set(HAVE_HPUX9 TRUE)
-    elseif(CMAKE_SYSTEM_VERSION MATCHES "[A-Z.]*10\.0")
-        #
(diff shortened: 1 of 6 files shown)


guyharris:
Looks good to me.

> reject these versions in both Autoconf and CMake.

Given that, should we, when testing in the code whether we're building for HP-UX, just test whether any of `_hpux`, `hpux`, and `__hpux` are defined, as per [the operating systems page](https://sourceforge.net/p/predef/wiki/OperatingSystems/) in [the Pre-defined Compiler Macros Wiki](https://sourceforge.net/p/predef/wiki/Home/)?  (They don't show a way of testing the HP-UX version at compile time, but the configure script and CMakeLists.txt handle that.)

infrastation:
I considered switching to `__hpux`, but eventually left `HAVE_HPUX10_20_OR_LATER` in place to make it easy to add code specific only to a later HP-UX version if necessary. If you see a better way to arrange it, please do. The wiki link reminds me: `git grep MSDOS` still returns some conditional (i.e. dead) code, is there any reason to keep it?

guyharris:
> I considered switching to `__hpux`, but eventually left `HAVE_HPUX10_20_OR_LATER` in place to make it easy to add code specific only to a later HP-UX version if necessary. If you see a better way to arrange it, please do.

We can leave it in for now; I suspect there may not be too much in the way of later HP-UX versions, given that 1) Itanium is dead and 2) HPE appear to have no plan to port HP-UX to x86-64 - as far as I know, their plan is to do binary-to-binary translation of Itanium code to x86-64(!) inside an HP-UX emulator atop Linux, and I suspect that, even if they add some features to the emulated HP-UX, it probably won't be tweaks to DLPI that will affect us.

At that point, most packet capture on HPE servers might be done directly on Linux.

> The wiki link reminds me: `git grep MSDOS` still returns some conditional (i.e. dead) code, is there any reason to keep it?

Only if we plan to support a no-capture libpcap for MS-DOS, for use by programs that read capture files.  I don't think we do.

infrastation:
Done. Various documentation updates will be next, and that should be it.

infrastation:
For posterity, [here](https://github.com/cpredef/predef/blob/master/OperatingSystems.md) is a version of the above document with a few additional operating systems.

PR #1267, opened 2024-02-04
CMake: translate the changes for Hurd support from autotools.

guyharris:
Untested, but it should handle Hurd the same way it's handled by the configure script.


fxlb:
@sthibaul @LocutusOfBorg 
Please, could you test the Hurd support in libpcap with CMake build via:
`MATRIX_CMAKE=yes ./build_matrix.sh`

sthibaul:
There were a couple harmless warnings, fixed by https://github.com/the-tcpdump-group/libpcap/pull/1268

fxlb:
FYI, full test (Autoconf and CMake) can be done with:
`./build_matrix.sh`

sthibaul:
Yes, with #1268 applied, it all builds fine (`Tested setup count: 16`)

PR #1268, opened 2024-02-04
hurd: Avoid warnings

sthibaul:
The warning for line 37 was:
    
    ./pcap-hurd.c:196:35: warning: expression does not compute the number of elements in this array; element type is 'struct bpf_insn', not 'short' [-Wsizeof-array-div]
                              (filter_array_t_filter, FILTER_COUNT);
                                                      ^~~~~~~~~~~~
    
But device_set_filter really does want the size in units of shorts.

### DIFF
--- a/pcap-hurd.c
+++ b/pcap-hurd.c
@@ -34,10 +34,13 @@ static struct bpf_insn filter[] = {
 	{ BPF_RET | BPF_K, 0, 0, MAXIMUM_SNAPLEN },
 };
 
-#define FILTER_COUNT (sizeof(filter) / sizeof(short))
+/* device_set_filter calls net_set_filter which uses CSPF_BYTES which counts in
+ * shorts, not elements, so using extra parenthesis to silence compilers which
+ * believe we are computing wrong here. */
+#define FILTER_COUNT (sizeof(filter) / (sizeof(short)))
 
 static int
-pcap_read_hurd(pcap_t *p, int cnt, pcap_handler callback, u_char *user)
+pcap_read_hurd(pcap_t *p, int cnt _U_, pcap_handler callback, u_char *user)
 {
 	struct net_rcv_msg *msg;
 	struct pcap_hurd *ph;
@@ -258,7 +261,7 @@ pcapint_create_interface(const char *device _U_, char *ebuf)
 }
 
 int
-pcapint_platform_finddevs(pcap_if_list_t *alldevsp, char *errbuf)
+pcapint_platform_finddevs(pcap_if_list_t *alldevsp _U_, char *errbuf _U_)
 {
 	return 0;
 }


pcap-hurd.c
```diff
@@ -34,10 +34,10 @@ static struct bpf_insn filter[] = {
 	{ BPF_RET | BPF_K, 0, 0, MAXIMUM_SNAPLEN },
 };
 
-#define FILTER_COUNT (sizeof(filter) / sizeof(short))
+#define FILTER_COUNT (sizeof(filter) / (sizeof(short)))
```
fxlb, line 37: What was the warning with this line?

sthibaul: gcc complains that the division is possibly incorrect because it assumes that we wanted to count the number of elements, and we are not dividing the size of the array (in bytes) by the size of an element (in bytes) to get the number of elements. And indeed, bpf programs are counted in number of shorts.

infrastation: This one is from Clang 11.0 only:
```
./pcap-hurd.c:196:35: warning: expression does not compute the number of elements in this array; element type is 'struct bpf_insn', not 'short' [-Wsizeof-array-div]
                          (filter_array_t_filter, FILTER_COUNT);
                                                  ^~~~~~~~~~~~
./pcap-hurd.c:37:38: note: expanded from macro 'FILTER_COUNT'
#define FILTER_COUNT (sizeof(filter) / sizeof(short))
                             ~~~~~~  ^
./pcap-hurd.c:32:24: note: array 'filter' declared here
static struct bpf_insn filter[] = {
                       ^
./pcap-hurd.c:196:35: note: place parentheses around the 'sizeof(short)' expression to silence this warning
./pcap-hurd.c:37:38: note: expanded from macro 'FILTER_COUNT'
#define FILTER_COUNT (sizeof(filter) / sizeof(short))
                                     ^
```
This looks a genuine bug to me, should have been `sizeof(filter[0])`.

The other three are visible with both Clang 11.0 and GCC 10.2.

infrastation:
How to get a functional Hurd VM:
1. Install VirtualBox and qemu-utils.
2. Download `debian-hurd.img.tar.gz` as shown on [this page](https://www.debian.org/ports/hurd/hurd-install).
3. Extract the `.img` file from the tarball and convert it to a VirtualBox disk image:
   ```
   qemu-img convert -f raw -O vdi debian-hurd-20210812.img debian-hurd-20210812.vdi
   ```
4. Create a 32-bit VM and use the file as the disk.
5. Boot and log in as root with no password.
6. `apt-get update; apt-get install git cmake flex bison autoconf clang libpcap-dev`
7. Create and use an unprivileged user.
8. Clone the repositories and build as usual.

pcap-hurd.c
infrastation: If the size uses `sizeof(short)` as the unit, it would be useful to have a comment about it.

sthibaul: No, FILTER_COUNT really has to be the size in terms of shorts, that's how bpf is interfaced in device_set_filter, if you divide by `sizeof(filter[0])` things will get completely wrong.

It can look like an unfortunate choice, but it happens to be how it is, just like other set_status etc. calls of the Mach kernel.

sthibaul: The other size of device_set_filter is implemented here: https://git.savannah.gnu.org/cgit/hurd/hurd.git/tree/libbpf/bpf_impl.c#n589 with CSPF_BYTES defined here: https://git.savannah.gnu.org/cgit/hurd/hurd.git/tree/libbpf/bpf_impl.h#n93, filter_t being a short.

sthibaul: > If the size uses `sizeof(short)` as the unit, it would be useful to have a comment about it.

done so

infrastation:
I also have a few fixes for tcpdump, will commit shortly.

infrastation:
Also, if the system has `libdbus-1-dev` installed (as a dependency of `libpcap-dev`, or otherwise), libpcap will not pass `make check`:
```
$ testprogs/findalldevstest
dbus-system
        Description: D-Bus system bus
        Flags:

dbus-session
        Description: D-Bus session bus
        Flags:

Error in pcap_lookupnet: SIOCGIFADDR: dbus-system: No such device
```
But that's a different bug.

infrastation:
Any objections to merge these changes?

fxlb:
> Any objections to merge these changes?

It should be better to add the warning for L37 in the body on the commit message, like:
```
The warning for line 37 was:
<folded warning message>
```

pcap-hurd.c
guyharris: You might want to expand the comment to indicate that, while the parentheses around `sizeof(short)` are not required by the syntax of C, they're there to suppress a warning from some compilers that know about the "divide the size of an array by the size of an array element in order to find the number of elements of the array" idiom in C, and see that the divisor *isn't* the size of the array and warn that this *isn't* computing the number of elements in the array, and use the extra parentheses as an indication that the programmer is saying "trust me, this is correct even though the divisor isn't the size of an array element", so that we can forestall people submitting "helpful" patches/pull requests that remove the "unnecessary" parentheses.

sthibaul:
> > Any objections to merge these changes?
> 
> It should be better to add the warning for L37 in the body on the commit message, like:
> 
> ```
> The warning for line 37 was:
> <folded warning message>
> ```

done so

fxlb:
> Any objections to merge these changes?

I have no more objections.

infrastation:
Thank you.

infrastation:
#1269 it is.

PR #1273, opened 2024-02-12
Add support for filtering VXLAN packets

sysvinit:
This change extends libpcap with support for filtering VXLAN ("Virtual eXtensible LAN") packets. The implementation is based on the existing support for Geneve.

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -145,6 +145,7 @@ struct addrinfo {
 #endif
 
 #define GENEVE_PORT 6081
+#define VXLAN_PORT  4789
 
 #ifdef HAVE_OS_PROTO_H
 #include "os-proto.h"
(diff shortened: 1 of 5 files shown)


gmelikov:
Sorry if it's an off topic, but I wonder if it's possible to filter with another port (not only with 4789)? There may be cases when another port is used for vxlan. The question may be not about this exact patch, but about common method to override hardcoded ports.

guyharris:
> Sorry if it's an off topic, but I wonder if it's possible to filter with another port (not only with 4789)? There may be cases when another port is used for vxlan. The question may be not about this exact patch, but about common method to override hardcoded ports.

Yes, it's off-topic, and should be filed as a separate issue in t[he issues list](https://github.com/the-tcpdump-group/libpcap/issues).

sysvinit:
> This also needs a change log entry and a note in the "backward compatibility" section of the pcap-filter man page (and a bump of the last change date).

I've made the necessary changes in #1334.

satanaileo:
expr "vxlan and arp" not supported

infrastation:
`vxlan` is going to be available in libpcap release 1.11.0 later.

satanaileo:
> `vxlan` is going to be available in libpcap release 1.11.0 later.

I compiled libpcap of current master code, “vxlan and udp/tcp” works well and “vxlan and arp” not supported

infrastation:
If there is a reproducible problem, please make a new bug report with the exact steps to reproduce the problem.

Issue #1269, opened 2024-02-07
libpcap does not see any network capture devices on GNU/Hurd


infrastation:
These results are from the [20230608 snapshot of Debian 12 GNU/Hurd](https://cdimage.debian.org/cdimage/ports/12.0/hurd-i386/debian-hurd-20230608.img.tar.xz). The original problem manifested in libpcap build matrix, which runs `testprogs/findalldevstest`. Upon a closer look it became clear that other test programs also do not work as expected:
* `capturetest` and `nonblocktest` default to a non-network device and fail:
  ```
  capturetest: WARNING: SIOCGIFADDR: dbus-system: No such device
  capturetest: pcap_setnonblock failed: Non-blocking mode isn't supported for capturing on D-Bus
  ```
* `findalldevstest` fails with
  ```
  dbus-system
          Description: D-Bus system bus
          Flags: 
  
  dbus-session
          Description: D-Bus session bus
          Flags: 
  
  Error in pcap_lookupnet: SIOCGIFADDR: dbus-system: No such device
  ```
* `selpolltest` and `threadsignaltest` generate a warning:
  ```
  selpolltest: WARNING: SIOCGIFADDR: dbus-system: No such device
  Listening on dbus-system, using pcap_dispatch()
  1 packets seen, 1 packets counted after pcap_dispatch returns
  ^C
  ```
All of the above was attempted as a non-root user. The network interfaces are as follows:
```
$ ifconfig
/dev/eth0 (2):
  inet address  192.168.0.24
  netmask       255.255.255.0
  broadcast     192.168.0.255
  flags         UP BROADCAST RUNNING ALLMULTI MULTICAST
  mtu           1500
  link encap    Ethernet
  hardware addr 08:00:27:72:46:74

lo (1):
  inet address  127.0.0.1
  netmask       255.0.0.0
  flags         UP LOOPBACK RUNNING
  mtu           3924
  link encap    Local Loopback
```
However, tcpdump does not see any network interfaces (whether run as root or not):
```
$ ./tcpdump -D
1.dbus-system (D-Bus system bus) [none]
2.dbus-session (D-Bus session bus) [none]
```
To reproduce these results, libdbus-1-dev must be installed before compiling libpcap. If it is not installed, the list of capture devices is empty.

infrastation:
[...]

PR #1275, opened 2024-02-12
various improvements and bug fixes for Hurd

infrastation:
This among other things includes a workaround for #1269 and seems to work, but could use more eyeballs, so please proof-read if you can.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -45,6 +45,7 @@ DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
       Report non-existent devices correctly.
     Hurd:
       Support network capture devices too.
+      Fix a few device activation bugs.
 
 DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
   Summary for 1.10.5 libpcap release (so far!)
--- a/CHANGES
+++ b/CHANGES
@@ -43,6 +43,8 @@ DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
       Add a new test program for pcap_activate().
     Haiku:
       Report non-existent devices correctly.
+    Hurd:
+      Support network capture devices too.
 
 DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
   Summary for 1.10.5 libpcap release (so far!)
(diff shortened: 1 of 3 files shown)


infrastation:
One thing I notice now is that `pcap_setfilter_hurd()` does not need to test if the first argument is `NULL` because the function call happens from a function pointer in the same `pcap_t`, which implies the latter is not `NULL`.

Another thing is that the calls to `pcap_strerror()` should have been `pcapint_fmt_errmsg_for_errno()`.

infrastation:
This is going to be merged in one hour unless anyone sees anything else that requires improvement.

fxlb:
I look at the output of `findalldevstest` on a Hurd system.
It shows no loopback interface. Is it OK?

infrastation:
This is intended, until someone finds a way to add support for loopback packet capture. The existing code cannot do that.

infrastation:
That is, on GNU/Hurd the attempt to capture on loopback fails with an error. On Haiku it "succeeds", but all packet counters are zero, so perhaps meanwhile it would make sense to disable it there as well.

fxlb:
> so perhaps meanwhile it would make sense to disable it there as well.

Perhaps, indeed.

fxlb:
> The existing code cannot do that.

OK.

infrastation:
Loopback capture in Haiku turns out to be a bug, the bug fix is in the development branch, but is not yet in the beta branch. So that particular issue does not require a workaround.

guyharris:
> Loopback capture in Haiku turns out to be a bug, the bug fix is in the development branch, but is not yet in the beta branch. So that particular issue does not require a workaround.

Should it have a note in doc/README.haiku.md?

(Speaking of "you can open the device but you get no packets", I should add a note in doc/README.macos about monitor-mode captures on newer versions of macOS, at least with newer machines, not giving any packets.  There's apparently something that allows packet capturing, given that the "Sniffer" window in Apple's Wireless Diagnostics captures traffic by doing something and then running tcpdump on the Wi-Fi adapter with `-I`; it may be that it just disassociates the adapter from the network it's on.)

infrastation:
After the bug fix makes it into beta4 updates, it should not be difficult to tell the minimal "hrev" that has loopback capture working. I do not know when this will happen.

PR #1278, opened 2024-02-18
Haiku: Work around more portability issues. [skip ci]

infrastation:
I am going to merge these changes after re-testing one more time on Haiku hrev57585 from the next nightly build.

### DIFF
--- a/pcap-haiku.c
+++ b/pcap-haiku.c
@@ -15,6 +15,7 @@
 
 #include <sys/socket.h>
 #include <sys/sockio.h>
+#include <sys/utsname.h>
 
 #include <net/if.h>
 #include <net/if_dl.h>
@@ -26,6 +27,15 @@
 #include <string.h>
 #include <unistd.h>
 
+
+// IFT_TUN was renamed to IFT_TUNNEL in the master branch after R1/beta4 (the
+// integer value didn't change).  Even though IFT_TUN is a no-op in versions
+// that define it, for the time being it is desirable to support compiling
+// libpcap on versions with the old macro and using it on later versions that
+// support tunnel interfaces.
+#ifndef IFT_TUNNEL
+#define IFT_TUNNEL IFT_TUN
+#endif
 
 /*
  * Private data for capturing on Haiku sockets.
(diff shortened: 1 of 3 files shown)


pcap-haiku.c
```diff
@@ -288,6 +299,37 @@ pcap_activate_haiku(pcap_t *handle)
 //	#pragma mark - pcap API
 
 
+static int
+can_be_bound(const char *name)
+{
+	if (strcmp(name, "loop"))
```
guyharris, line 305: `strcmp(name, "loop") != 0` might make it a bit clearer that this is testing for names that aren't "loop".

pcap-haiku.c
```diff
@@ -296,6 +338,11 @@ pcapint_create_interface(const char *device, char *errorBuffer)
 		         "Interface name \"%s\" is too long.", device);
 		return NULL;
 	}
+	if (! can_be_bound(device)) {
+		snprintf(errorBuffer, PCAP_ERRBUF_SIZE,
+		         "Interface \"%s\" is not supported.", device);
```
guyharris, line 343: Perhaps "does not support capturing traffic" - it's "supported" in that you san presumably send packets to 127.0.0.1, but you can't see those packets in a capture in older versions of Haiku.

infrastation:
Updated as suggested.

guyharris:
> Still waiting on the development snapshot.

Presumably "the development snapshot" means Haiku hrev57585 from the next nightly build.

infrastation:
hrev57585 or later (the nightly builds stopped at hrev57581, see haiku/infrastructure#114). I am trying to build hrev57588 locally.

guyharris:
Should this be backported to 1.10?

guyharris:
...along with the previous improvements?

infrastation:
Yes, the recent change set makes tcpdump more useful. Let's give at least a few days for the dust to settle. Besides, I considered to validate packet AF for the DLT, and to size the buffer based on the MTU, but both directions were not fruitful for reasons I would like to document in the source code comments.

infrastation:
Indeed there is a minor bug in one of my recent commits, I am going to include the bug fix into these fixups.

infrastation:
Committed and now looks ready for back-porting to me. If anybody else can proof-read the recent Haiku commits, please do.

infrastation:
The change set has been backported to the 1.10 branch, libpcap compiles and works as expected on both R1/beta4 and the current master branch (hrev57592).

PR #1285, opened 2024-03-01
pcap: make the seconds and microseconds/nanoseconds fields unsigned.

guyharris:
That's what the latest version of the Editor's Copy of the pcap specification says they are.

Making the seconds field unsigned gives the pcap format a reprieve from the Y2.038K problem; it lasts only 68 years, though, so it has a Y2.106 problem that's trickier to fix.  It does mean that, if you have a capture file with network traffic prior to 1970, that won't work if it is, or is converted to, a pcap file, but RFC 791 was published in 1981, Xerox's experimental 3MB Ethernet came out in the mid-1970s, IBM's SNA came out in 1974, and even the ARPANET NCP was finalized in 1970, so ther's probably not a lot of network traffic that somebody would want to stuff into a pcap file.  (If you *must*, try a pcapng file, with an if_tsoffset option in the IDB - that option's value is signed.)

Making the microseconds/nanoseconds field unsigned means that you don't have to think about what the heck a negative value means.

Move the structures for pcap file formats into sf-pcap.c, to discourage people from relying on it (or, worse, changing it), and making it easier for libpcap code to avoid using it for anything other than pcap file reading and writing.


infrastation:
Should be either "Y2106" or "Y2.106K".

guyharris:
> Should be either "Y2106" or "Y2.106K".

Fixed.

PR #1289, opened 2024-03-05
Print MAC addresses in findalldevstest.

infrastation:
Any objections to make this change?

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -46,6 +46,7 @@ DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
       Make Sun C build warning-free.
       Make FreeBSD, Linux and macOS builds warning-free.
       Fix propagation of cc_werr_cflags() output.
+      Print MAC addresses in findalldevstest.
     Hurd:
       Support network capture devices too.
       Fix a few device activation bugs.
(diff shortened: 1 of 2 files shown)


fxlb:
It's better than printing `Address Family: Unknown (17)`. No objection.

guyharris:
On Windows, if `_WIN32_WINNT` is >= 0x0601:

- `AF_LINK` is defined;
- `struct sockaddr_dl` is defined, and has, as members:
   - `sdl_family` (an `ADDRESS_FAMILY`, which is a 16-bit unsigned integer);
   - `sdl_data` (an 8-byte array);
   - `sdl_zero` (a 4-byte array).

infrastation:
Thank you for the details, this convinced me not to try decoding the Windows version of the structure.

guyharris:
> this convinced me not to try decoding the Windows version of the structure.

Yeah, it has nothing obvious that indicates the type or the length of the link-layer address.

infrastation:
Ironically, `struct sockaddr_dl` is much more comprehensible, but I printed it in a way that works correctly on illumos only. I have a bug fix, going to double-check it and commit tomorrow.

PR #1290, opened 2024-03-05
Make pcap_strerror() more thread-safe.

infrastation:
I am reasonably sure it is right to require `strerror()`, but perhaps the pre-compile time checks could be omitted altogether and the compilation could fail the usual way in the unlikely case a system without `strerror()` manages to pass all the other checks that are based on C99. What do you think?

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -6,6 +6,7 @@ DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
       Remove an always-false pointer test from snf_read().
       Clean up DECnet address handling.
       struct pcap: Update buffer type from "void *" to "u_char *".
+      Make pcap_strerror() more thread-safe.
     Link-layer types:
       Add LINKTYPE_ETW/DLT_ETW.
       Add LINKTYPE_NETANALYZER_NG/DLT_NETANALYZER_NG (pull request
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -710,7 +710,6 @@ main(void)
 #
 # Now check for various system functions.
 #
-check_function_exists(strerror HAVE_STRERROR)
 check_function_exists(strerror_r HAVE_STRERROR_R)
 if(HAVE_STRERROR_R)
     #
--- a/cmakeconfig.h.in
+++ b/cmakeconfig.h.in
@@ -153,9 +153,6 @@
 /* Define to 1 if you have the <stdlib.h> header file. */
 #cmakedefine HAVE_STDLIB_H 1
 
-/* Define to 1 if you have the `strerror' function. */
-#cmakedefine HAVE_STRERROR 1
-
 /* Define to 1 if you have the <strings.h> header file. */
 #cmakedefine HAVE_STRINGS_H 1
 
--- a/configure.ac
+++ b/configure.ac
@@ -161,7 +161,6 @@ haiku*)
 	;;
 esac
 
-AC_CHECK_FUNCS(strerror)
 AC_CHECK_FUNC(strerror_r,
     [
 	#
--- a/fmtutils.c
+++ b/fmtutils.c
@@ -334,43 +334,10 @@ pcapint_vfmt_errmsg_for_errno(char *errbuf, size_t errbuflen, int errnum,
 	 */
 	if (!use_utf_8)
 		utf_8_to_acp_truncated(errbuf);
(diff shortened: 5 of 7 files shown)


guyharris:
> Always require strerror(), which is in C99.

It's in C90. [POSIX has had it since 2004](https://pubs.opengroup.org/onlinepubs/009695399/functions/strerror.html), and Visual Studio has supported it since [VS 2015](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/strerror-strerror-wcserror-wcserror?view=msvc-140).

It's probably in the configure script because libpcap and tcpdump predate 1989, and there were older UN\*Xes that *did* have `perror()` but not `strerror()`.  I don't think we need to check for it any more.

guyharris:
However, POSIX does *not* require `strerror()` to be reentrant, and thus doesn't require it to be thread-safe, so we should keep our thread-safe implementation on UN\*Xes.  Using `strerror_r()` if available is a potential enhancement.  We already do that for Windows, although Microsoft's documentation says it's thread-safe.

infrastation:
Thank you for clarifying. Some Linux man pages say it is C89, some say it is C99. Various BSD man pages say it is either 4.3BSD or 4.4BSD. Let me remove the unnecessary checks then.

infrastation:
Which one is the thread-safe implementation?

infrastation:
The next revision is slightly simpler. It does not change the effective behaviour, but makes it easier to see what it is. Making `pcap_strerror()` thread-safe remains to be done. If you are willing to do it anytime soon, I can move my changes out of the way.

guyharris:
> Some Linux man pages say it is C89, some say it is C99.

The man pages that say it's C89 might be correct; I no longer have my dead-tree C89 handy (I don't remember if I still have it at all), and don't have a PDF of it (even trying to find a PDF of C90 was a pain - Standards Australia had the only shop I could find that sold one), but I think the changes between C89 and C90 were minor; [according to the GCC documentation](https://gcc.gnu.org/onlinedocs/gcc/Standards.html), "There were no technical differences between these publications, although the sections of the ANSI standard were renumbered and became clauses in the ISO standard."

The man pages that say it's C99 were apparently written by somebody who was unfamiliar with the contents of C89 and C90.

guyharris:
> The next revision is slightly simpler. It does not change the effective behaviour, but makes it easier to see what it is. Making `pcap_strerror()` thread-safe remains to be done. If you are willing to do it anytime soon, I can move my changes out of the way.

Oh, hell, I missed that `pcap_strerror()` calls `strerror()` if it's available, so there's one more potential point of thread-unsafety in libpcap.  I think I *might* have checked the documentation or source code for the platforms on which we run and found that they all have thread-local buffers for the error message, so maybe in *practice* it's thread-safe, but we should probably use the explicitly thread-safe routines in the future.

infrastation:
Shall I try moving `strerror_r()` conditionals from `fmtutils.c` to `pcap_strerror()` to make `strerror()` the last-resort code path then?

guyharris:
> Shall I try moving `strerror_r()` conditionals from `fmtutils.c` to `pcap_strerror()` to make `strerror()` the last-resort code path then?

Given that it's an exported API, we can't eliminate it, even though it's only used in 2 places in pcap-ioplugin.c (which might be replaceable with one of the fmtutils.c routines) and one place in pcap-linux.c (where it's an fprintf to stderr, which should be eliminated as soon as we can).

So, yes, moving the conditionals there might be reasonable.

Note that, on Windows, `pcapint_vfmt_errmsg_for_errno()` provides a UTF-8 error string, *not* an error string in the local code page (unless your local code page is 65001, which is a code page using a multi-byte encoding called "UTF-8" :-)), so `pcap_strerror()` might have to turn the string into the local code page if it uses the `pcapint_vfmt_errmsg_for_errno()` code.

infrastation:
Alright, let me have a closer look...

infrastation:
I had a closer look at the code and produced some draft changes. Will post an updated version as soon as it is ready.

infrastation:
This is as much sense as I can currently make of this problem. With these changes it should be obvious that any remaining non-thread-safe problem space is negligible.

infrastation:
Now with slightly better source code comments. This is going to be merged as soon as CI checks have passed.

infrastation:
For posterity: Haiku, Hurd, OpenIndiana and Solaris with Sun C all passed.

PR #1291, opened 2024-03-06
Autoconf, CMake: Add a warning flag (-Wundef)

fxlb:


pcap-bpf.c
```diff
@@ -2341,7 +2341,7 @@ pcap_activate_bpf(pcap_t *p)
 		goto bad;
 	}
 #endif
-#if _BSDI_VERSION - 0 >= 199510
+#if defined(_BSDI_VERSION) && _BSDI_VERSION - 0 >= 199510
```
guyharris, line 2344: The C99 standard says, in the section about `#if`, that

> After all replacements due to macro expansion and the defined unary operator have been performed, all remaining identifiers are replaced with the pp-number 0, and then each preprocessing token is converted into a token.

which seems to indicate that the old version is legal, with the expression turning into `0 - 0 >= 199510`, but adding the additional test makes things a bit clearer.

fxlb: Does this mean that `_BSDI_VERSION - 0 >= 199510`could be replaced by `_BSDI_VERSION >= 199510`?

guyharris: Probably.  At least with the version of clang on my Mac, removing the `defined(_BSDI_VERSION ) &&` protection, and replacing what remains with `_BSDI_VERSION >= 199510`, doesn't cause an error.

Given that C90 also says

> After all replacements due to macro expansion and the defined unary operator have been performed, all remaining identifiers are replaced with the pp-number 0, and then each preprocessing token is converted into a token.

I suspect that the "{variable name} - 0" idiom *might* have been a way to handle those tests on pre-standard preprocessors, before "replace all undefined identifiers with 0" was required by the standard, e.g. if some preprocessor replaced them with empty space, so that `_BSDI_VERSION - 0 >= 199510` became `- 0 >= 199510`.

fxlb: > doesn't cause an error.

Even with `-Wundef`?

guyharris: I didn't try that, because `-Wundef` requests warnings for standard-conforming but perhaps unintended code, e.g. if you meant to say `_BSDI_VERSION` but said `_BSDO_VERSION` or something such as that, so I'd *expect* it to fail.  If `_BSDI_VERSION - 0 >= 199510` gets a warning with `-Wundef`, which it does, I'd expect `_BSDI_VERSION >= 199510` to get the same warning.

I was testing it because I didn't know about the "replace undefined identifiers with 0" rule and figured it'd get an error; when it didn't, I checked C99 and found that rule.

fxlb: Test on:
```
$ uname -a
Darwin cfarm104.cfarm.net 21.6.0 Darwin Kernel Version 21.6.0: Mon Aug 22 20:20:05 PDT 2022; root:xnu-8020.140.49~2/RELEASE_ARM64_T8101 arm64
```
```
./pcap-bpf.c:2344:5: warning: '_BSDI_VERSION' is not defined, evaluates to 0 [-Wundef]
#if _BSDI_VERSION >= 199510
    ^
```

fxlb: (With `-Wundef`)

fxlb:
Thus I will use: `#if defined(_BSDI_VERSION) && _BSDI_VERSION >= 199510`

pcap-bpf.c
fxlb: > Anyway, since BSD/OS is not supported anymore, the condition is most likely either always-true or always-false.

Does this mean we could just remove lines 2344-2364, saying that we no more support BSD/OS?

infrastation:
FWIW, this passes on Haiku and Hurd.

fxlb:
> FWIW, this passes on Haiku and Hurd.

Thus, for the error in `pcap-bpf` commit, there are two ways to fix:
1) Keep the PR as is.
2) Remove lines 2344-2364, saying that we no more support BSD/OS.

Do you have a preference?

fxlb:
> Please delete the dead code

Done.

PR #1293, opened 2024-03-09
more pcap_findalldevs() fixes

infrastation:
This improves SNR a bit and fixes a bug. It still looks a bit off that `get_sa_len()` defaults to `sizeof (struct sockaddr)` instead of `sizeof (struct sockaddr_storage)`, but I cannot quickly prove that change to work correctly in all cases, so I am not making it right now. Unless someone sees an issue with these changes, they will be merged soon.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -52,6 +52,8 @@ DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
       Fix a few device activation bugs.
       Count and timestamp packets better.
       Add kernel filtering, fix userland filtering.
+    Solaris:
+      Fix AF_LINK handling on illumos.
 
 DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
   Summary for 1.10.5 libpcap release (so far!)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1065,7 +1065,6 @@ if(WIN32)
 else(WIN32)
     set(CMAKE_EXTRA_INCLUDE_FILES unistd.h sys/socket.h)
 endif(WIN32)
-check_type_size("struct sockaddr_storage" STRUCT_SOCKADDR_STORAGE)
 check_type_size("socklen_t" SOCKLEN_T)
 cmake_pop_check_state()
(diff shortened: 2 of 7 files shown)


infrastation:
I managed to make some more sense of the problem, and it seems the AF_LINK bug fix must go into 1.10.5.

guyharris:
If you're curious about the history of `sa_len`, it was, as I remember, introduced to support the OSI stack, which had variable-length network and transport addresses.  It looks as if it first appeared in 4.3-Reno, so I'd expect it to be in all 4.4-Lite-based systems - so all of the \*BSDs, including CupertinoBSD :-) - but not necessarily be in any *other* systems.

The networking stack in SVR4, and thus, Solaris (SunOS 5.x) is based in the SunOS 4.x stack plus the STREAMS stuff from System V Release 3, although Sun/Oracle may have un-STREAMSified it somewhat.  The 4.x stack was based on 4.3BSD; they don't appear to have picked up the 4.3-Reno changes, so no `sa_len` in Solaris or Illumos.

Linux also doesn't have it, as the stack was developed independently; obviously, they made it BSD-compatible, but they may have decided to make it compatible with the 4.3 stack.

Haiku probably follows BeOS, which may have decided to go with 4.4-Lite.  Hurd may have done the same thing.

Not sure why AIX has it, given that it dates back before 4.3-Reno and, I think, has a BSD-derived stack.

guyharris:
I'm also a bit surprised that AIX 7 uses fad-gifc.c, but it apparently came from BSD/OS, and isn't in POSIX or any of the RFCs that define socket interfaces, so I guess other BSD-derived OSes, Linux, and Sun/Oracle decided to pick it up and neither IBM nor HP did.

infrastation:
Thank you for explaining. I tried to make sense of the long "This is fun." comment in `fad-getad.c`, but could not quickly tell which groups still exist as described and which OSes are in which group, so ended up finding the particulars and making a list. Could you update that comment where possible? There is another version of it in `fad-gifc.c`.

PR #1295, opened 2024-03-18
usb-linux: reorganize the isochronous length handling code.

guyharris:
Put useful inline functions, used both by the capture code and the savefile length fixup code, into pcap-usb-linux-common.h; some of those came from, or are based on routines in, pcap-usb-linux-common.c, which we remove.

When capturing, all we need to do is check whether a message with data is a "this is complete" incoming isochronous transfer event and, if so, use the routine to calculate its original length.

Pull the routine to fix up the original length into pcapint_post_process().

When reading savefiles, we need to do additional checks to guess whether the length needs fixing and to determine whether the number of descriptors is sane and whether we have all the descriptors; do that in pcapint_post_process().

Add some additional checks.

Call the "len" field of pcap's metadata header the "original length" rather than the "on-the-wire" length; that matches the terminology from the current versions of the pcap and pcapng specs, and doesn't imply that the length is the length of the packet on the transmission medium, as that's not always the case - for USB, it's computed based on information from the endpoint on which we're capturing, and may also include padding between isochronous data chunks.


infrastation:
(Cirrus CI does not have a feature to send an e-mail)
```
make whitespacecheck
pcap-usb-linux-common.h:56:	return (sizeof(pcap_usb_header_mmapped) + 
Error: Trailing space(s).
make: *** [Makefile:883: whitespacecheck] Error 1
```

pcap-usb-linux-common.h
```diff
@@ -19,8 +19,109 @@
...
+	}
+
+	/*
+	 * Return the sum of the total header length (memory-mapped
+	 * header and ISO descriptors) and the data length, clamped
+	 * to UINT_MAX.
+	 *
+	 * We've made sure that the number of descriptors is
+	 * <= USB_MAXDESC, so we know that the total size,
+	 * in bytes, of the desciptors fits in a 32-bit
```
fxlb, line 123: desciptors ==> descriptors

guyharris: Fixed in 98101580df09fe1d47690dec8b8950847a79aeab.

guyharris:
> ```
> make whitespacecheck
> pcap-usb-linux-common.h:56:	return (sizeof(pcap_usb_header_mmapped) + 
> Error: Trailing space(s).
> make: *** [Makefile:883: whitespacecheck] Error 1
> ```

Fixed in 98101580df09fe1d47690dec8b8950847a79aeab.

Issue #1052, opened 2021-09-23
Dont add DLT_LINUX_SLL headers for CAN dumps.


Elfoniok:
After replacing DLT_CAN_SOCKETCAN with DLT_LINUX_SLL, can dumps now records pseudo link layer header. This extend the size of the logs twofold.

I am wondering if it would be possible to add "save space" mode for can dumps? So it either strip SLL headers or use DLT_CAN_SOCKETCAN.

guyharris:
We switched to cooked mode for `ARPHRD_CAN` in commit 93ca5ff7030aaf1219e1de05ec89a68384bfc50b in order to allow readers to distinguish between CAN and CAN FD:

```
commit 93ca5ff7030aaf1219e1de05ec89a68384bfc50b
Author: Guy Harris <gharris@sonic.net>
Date:   Thu Aug 25 13:26:43 2016 -0700

    On Linux, handle all CAN captures with pcap-linux.c, in cooked mode.

    There's no need to support capturing on PF_CAN sockets; CAN interfaces
    show up as regular interfaces on which you can capture with PF_PACKET
    sockets, so just let pcap-linux.c handle them, and get rid of
    pcap-can-linux.c.

    Capture on them in cooked mode, so we get the protocol field and can
    distinguish between "classic" CAN and CAN FD.

    The hardware for which pcap-canusb-linux.c was intended never reached
    production:

        https://github.com/axos88/libpcap/commit/f3edbb599b8cbcc7e4560000dcba8e$

    so we don't need pcap-canusb-linux.c, either.

    This all removes the need for the "host-endian" link-layer header type;
    it never made it into a libpcap release, so we just remove it.  The
    "big-endian" link-layer header type is kept for the benefit of packets
    captured with the old pcap-can-linux.c code; we revert it to its old
    name.
```

[The Linux kernel added a bit in the flags in the SocketCAN header that's never set by the kernel, but that is reserved for flagging frames as CAN or CAN FD](https://git.kernel.org/pub/scm/linux/kernel/git/netdev/net-next.git/commit/?id=02546884221279da2725e87e35348290470363d7), so there's now a way to distinguish between them without the `DLT_LINUX_SLL` pseudo-header.

So we could switch back to `DLT_CAN_SOCKETCAN`, and, for each CAN packet received:

- put the CAN ID and flags field in host byte order;
[...]

PR #1296, opened 2024-03-18
socketcan: make CAN CC/FD/XL frame handling less invasive

hartkopp:
The different CAN frame types are defined by Linux SLL_P types in the sll_protocol field and the length of the frame.

LINUX_SLL_P_CANXL:
The frame length for CAN XL can be 12 + 1 to 12 + 2048 (13 .. 2060) byte. Additionally the CANXL_XLF flag has to be set.

LINUX_SLL_P_CAN and LINUX_SLL_P_CANFD:
The frame length for CAN CC is 16 byte and for CAN FD it is 72 byte. Additionally the CANXL_XLF flag is cleared in CAN CC/FD frames.

Reduce the modification and 'sanitizing' efforts only to the above requirements to be robust against incompliant content which might be provided via PACKET sockets but do not manipulate further content.

This helps to identify uninitialized data and to support new features in formerly reserved structure elements without changing the libpcap.


guyharris:
So which mechanisms can, in all versions of the Linux kernel starting with 2.6.27, reliably be counted on to provide a classic/FD/XL indication?  Is it sufficient just to check the protocol field - and then ensure that whatever flag bits are necessary to be set or clean to conform to [the LINKTYPE_CAN_SOCKETCAN specification](https://www.tcpdump.org/linktypes/LINKTYPE_CAN_SOCKETCAN.html) are set or cleared?

hartkopp:
**Inside** the kernel the different CAN frames are transported in skbuff structures with their own ETH_P type.

See in include/uapi/linux/if_ether.h
```
#define ETH_P_CAN       0x000C          /* CAN: Controller Area Network */
#define ETH_P_CANFD     0x000D          /* CANFD: CAN flexible data rate*/
#define ETH_P_CANXL     0x000E          /* CANXL: eXtended frame Length */
```


See in af_can.c (the forced setting of CANFD_FDF was added recently)
```
int can_send(struct sk_buff *skb, int loop)
{
        struct sk_buff *newskb = NULL;
        struct can_pkg_stats *pkg_stats = dev_net(skb->dev)->can.pkg_stats;
        int err = -EINVAL;

        if (can_is_canxl_skb(skb)) {
                skb->protocol = htons(ETH_P_CANXL);
        } else if (can_is_can_skb(skb)) {
                skb->protocol = htons(ETH_P_CAN);
        } else if (can_is_canfd_skb(skb)) {
                struct canfd_frame *cfd = (struct canfd_frame *)skb->data;

                skb->protocol = htons(ETH_P_CANFD);

                /* set CAN FD flag for CAN FD frames by default */
                cfd->flags |= CANFD_FDF;
        } else {
                goto inval_skb;
        }
```

These ETH_P types are also set correctly by the CAN hardware drivers - and they are visible via the PF_PACKET socket.

When it comes to the reading from a CAN_RAW socket, the user can enable the reception of CAN FD and CAN XL traffic via setsockopt() syscalls. When both are **enabled** a read() syscall can either deliver a CAN CC/FD/XL frame.

And then the check sequence in user space  is done like in this example:
https://github.com/linux-can/can-utils/blob/master/canlogserver.c#L404

1. Check for the minimum CAN XL size and the CANXL_XLF bit set -> CAN XL
2. Check for the CAN_MTU (== 16) -> CAN CC
3. Check for the CANFD_MTU (== 72) -> CAN FD
4. No match in the above checks -> failure

This sequence is stable and works on all Linux kernels since 2.6.25 (when SocketCAN became mainline).

And I thought this simple sequence could also be a candidate for libpcap.

hartkopp:
Additional comment:

The reason why CANXL_XLF is set resp. cleared in this libpcap patch is because of a bad player might create ETH_P_CANXL frames via PF_PACKET socket that has not set the CANXL_XLF bit.

But with the LINKTYPE_CAN_SOCKETCAN you face the same situation as when reading from a CAN_RAW socket (one buffer of some length you need to evaluate). And therefore the CANXL_XLF bit and the tvb_reported_length() are vital for a proper distinction of the CAN CC/FD/XL frame.

guyharris:
So I see several ways of determining the flavor of CAN:

1. Look at the protocol field - that works when capturing, but the `LINKTYPE_CAN_SOCKETCAN` header doesn't include the protocol, so it won't work when reading the capture.  (At one point we used a full `LINKTYPE_LINUX_SLL` header, but, in issue #1052, somebody complained, because it doubled the size of capture files, so we went back to `LINKTYPE_CAN_SOCKETCAN`.)
2. Look at the flag bits - as long as all versions of the kernel that support CAN FD set the FD flag for FD frames and all versions of the kernel that support CAN XL set the XL flag for XL frames, that should be sufficient, unless older versions of the kernel can't be relied on to make sure the FD and XL flags are *clear* for classic CAN frames and the XL flag is clear for CAN FD frames.
3. Look at the frame lengths - I infer from "Check for the minimum CAN XL size and the CANXL_XLF bit set -> CAN XL" that this doesn't suffice for CAN XL, but it can be used by itself to distinguish between classic and FD.

hartkopp:
> So I see several ways of determining the flavor of CAN:
> 
>     1. Look at the protocol field - that works when capturing, but the `LINKTYPE_CAN_SOCKETCAN` header doesn't include the protocol, so it won't work when reading the capture.  (At one point we used a full `LINKTYPE_LINUX_SLL` header, but, in issue [Dont add DLT_LINUX_SLL headers for CAN dumps. #1052](https://github.com/the-tcpdump-group/libpcap/issues/1052), somebody complained, because it doubled the size of capture files, so we went back to `LINKTYPE_CAN_SOCKETCAN`.)

Unfortunately I was not involved in this discussion. And today I'm not sure how not make a better "backwards compatible" improvement, than probably this patch here ...

>     2. Look at the flag bits - as long as all versions of the kernel that support CAN FD set the FD flag for FD frames and all versions of the kernel that support CAN XL set the XL flag for XL frames, that should be sufficient, unless older versions of the kernel can't be relied on to make sure the FD and XL flags are _clear_ for classic CAN frames and the XL flag is clear for CAN FD frames.

The CANXL_XLF bit is safe but the CANFD_FDF bit has been introduced in 2022 (10 years after CAN FD support). And we can not be sure about out-of-tree CAN hardware drivers:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/net/can/af_can.c?id=061834624c87282c6d9d8c5395aaff4380e5e1fc 

>     3. Look at the frame lengths - I infer from "Check for the minimum CAN XL size and the CANXL_XLF bit set -> CAN XL" that this doesn't suffice for CAN XL, but it can be used by itself to distinguish between classic and FD.

Yes. When we have the minimum CAN XL frame size, we are also safe to access the canxl_frame.flags element when checking for the CANXL_XLF flag.
This is really a different approach from the CAN CC/FD length-only check as the CAN XL frames are only of the size that the data content fits into the frame. Otherwise we always would need to copy 2060 bytes - even if we only had one byte data size.

hartkopp:
Summarizing the situation after getting some sleep:

When going back from DLT_LINUX_SLL to DLT_CAN_SOCKETCAN we went back from a three lane highway to a single lane road. Therefore the CAN frames going through this single lane road have to be self-contained in a way that we can evaluate the buffer and get the information about the CAN CC/FD/XL status.

Before CAN XL this was easy for Wireshark: 
```
            if (tvb_reported_length(tvb) == 72)
                can_packet_type = PACKET_TYPE_CAN_FD;
            else if  (tvb_reported_length(tvb) == 16)
                can_packet_type = PACKET_TYPE_CAN;
```
As the tvb_reported_length() is the right marker for CC/FD detection I wondered why there was a discussion about CANFD_FDF and all the sanitizing efforts?!? Libpcap and Wireshark should just display the content and not modifying it.

With CAN XL we now need to check for the CANXL_XLF bit first and then check for CAN FD/CC in Wireshark:

```
        canxl_flags = tvb_get_guint8(tvb, CANXL_FLAGS_OFFSET);
        if ((tvb_reported_length(tvb) > 12) && (canxl_flags & CANXL_XLF)) {
            can_packet_type = PACKET_TYPE_CAN_XL;
        } else {
            if (tvb_reported_length(tvb) == 72)
                can_packet_type = PACKET_TYPE_CAN_FD;
            else if  (tvb_reported_length(tvb) == 16)
                can_packet_type = PACKET_TYPE_CAN;
        }
```
When all these conditions are not met, the can_packet_type remains PACKET_TYPE_UNKNOWN.
And for Wireshark this is addressed with this patch:
https://gitlab.com/wireshark/wireshark/-/merge_requests/14886/diffs?commit_id=942125a423dbe05313518e0c745cd7cb986491c5

These two patches just implement the rules how to encode and decode the self-contained CAN frames on the single lane road analog to the reading of CAN CC/FD/XL frames from a CAN_RAW socket.

hartkopp:
Hello @guyharris ,

to answer your question above about the `CANFD_FDF` flag I updated the LINKTYPE_CAN_SOCKETCAN documentation and sent a pull request for the HTML code: https://github.com/the-tcpdump-group/tcpdump-htdocs/pull/36

The PR for Wireshark has already been merged: https://gitlab.com/wireshark/wireshark/-/merge_requests/14886

Do you still have concerns about merging this patch? It also enables the display of the Len 8 DLC value for CAN CC frames.
https://gitlab.com/wireshark/wireshark/-/merge_requests/14886/diffs?commit_id=a98b35d98884344400b20aad815473e39f28b20d

hartkopp:
Ping?

PR #1309, opened 2024-03-30
Require a proof of suitable snprintf(3) implementation.

infrastation:
libpcap was not working correctly on Solaris 9 already, the test makes it obvious.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -30,11 +30,6 @@ environment:
   #
   matrix:
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
-      GENERATOR: "MinGW Makefiles"
-      SDK: WpdPack
-      AIRPCAP: -DDISABLE_AIRPCAP=YES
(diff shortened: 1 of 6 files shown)


infrastation:
```
Environment: APPVEYOR_BUILD_WORKER_IMAGE=Visual Studio 2015, GENERATOR=MinGW Makefiles, SDK=WpdPack, AIRPCAP=-DDISABLE_AIRPCAP=YES, MINGW_ROOT=C:\mingw-w64\x86_64-8.1.0-posix-seh-rt_v6-rev0
[...]
-- The C compiler identification is GNU 8.1.0
[...]
-- Performing Test SUITABLE_SNPRINTF
-- Performing Test SUITABLE_SNPRINTF - Failed
```
I suspect the unsuitable `printf()` has more to do with MinGW rather than GCC 8.1, which dates back to 2018. The best I can do about it is to remove the only MinGW entry. tcpdump never had one. Let me know if that is not the correct course of action.

infrastation:
Now with some more clean-ups.

infrastation:
This is going to be merged as soon as the CI checks have settled.

PR #1312, opened 2024-04-03
Makefile.in: Update the .c.o build rule

fxlb:
Don't use the hacks for the rather old SunOS 4.

It's helps to do VPATH builds, e.g. 32-bit, 64-bit in two directories.
(https://www.gnu.org/software/automake/manual/html_node/VPATH-Builds.html)

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -79,12 +79,8 @@ RANLIB = @RANLIB@
 LEX = @LEX@
 BISON_BYACC = @BISON_BYACC@
 
-# Explicitly define compilation rule since SunOS 4's make doesn't like gcc.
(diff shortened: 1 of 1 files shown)


infrastation:
It would be a good idea to confirm this works on Solaris 10, 11 and AIX 7 (`make` is a bit different there).

infrastation:
Tested this change on Solaris 10, 11.4 CBE and AIX 7.1, did not observe any new breakage.

fxlb:
Thanks, Tested on some cfarm hosts?

infrastation:
Solaris 10: OpenCSW, Solaris 11.4 CBE: my local VM, AIX 7.1: cfarm.

PR #1325, opened 2024-05-28
rpcap: use struct sockaddr as a type in rpcap_deseraddr().

guyharris:
This eliminates one cast, and may also eliminate some compiler warnings that we're assigning to a struct sockaddr_storage * a pointer to a structure that's smaller than a struct sockaddr_storage.

### DIFF
--- a/pcap-rpcap.c
+++ b/pcap-rpcap.c
@@ -296,7 +296,7 @@ static int rpcap_read_packet_msg(struct pcap_rpcap const *, pcap_t *p, size_t si
 #define SOLARIS_AF_INET6		26
 
 static int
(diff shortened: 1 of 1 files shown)


fxlb:
> may also eliminate some compiler warnings

Yes, the two `-Walloc-size` warnings with gcc 14 are fixed.

guyharris:
> > may also eliminate some compiler warnings
> 
> Yes, the two `-Walloc-size` warnings with gcc 14 are fixed.

Yes, that was provoked me to look at that and make the change.

fxlb:
Thanks!

PR #1326, opened 2024-06-03
Add PCAP_ERROR_CAPTURE_NOTSUP error.

guyharris:
This means "the platform you're on lacks a capture mechanism that supports that device".

This was inspired by some Stack Overflow questions in which the user was probably trying to run a capture program on Windows Services for Linux 1, in which the NT kernel attempts to simulate Linux system calls but does not support packet sockets.

It can, and does, also handle a system in which the BPF capture mechanism is the native capture mechanism but that has no BPF devices.

The hope is that the error message they get will better indicate the underlying problem.


fxlb:
Should this change be backported in the 1.10 branch?

fxlb:
Done via 1025ec5a59e0ec6ef3930b618d9ba136840228b3 in the 1.10 branch and 9f1fa4b88606861130d9ec62bd88a1a858078ad5 in the main branch.

Issue #1328, opened 2024-06-17
pcap_stats woefully under-reports ifdropped packets on Linux


gtedesco-r7:
Many drivers do not report anything into `rx_missed_errors` or `rx_fifo_errors`, see [vmxnet](https://github.com/torvalds/linux/blob/6ba59ff4227927d3a8530fc2973b80e94b54d58f/drivers/net/vmxnet3/vmxnet3_ethtool.c#L179) for example.

At least one issue appears to be that [linux_if_drops()](https://github.com/the-tcpdump-group/libpcap/blob/master/pcap-linux.c#L771-L777) does not report the value of `rx_dropped`, which indicates the number of packets [not forwarded to the upper layers for processing](https://github.com/torvalds/linux/blob/6ba59ff4227927d3a8530fc2973b80e94b54d58f/Documentation/ABI/testing/sysfs-class-net-statistics#L44-L52).

The impact of this problem is that if I do a controlled test sending fixed size packets at line rate, I know a system can only handle 2-3gbps of traffic on a single core. But when I start capturing with libpcap and look at the drop rates, they stay at zero (or unbelievably low), even when the traffic is in excess of 15gbps...

 Although, when we added the `rx_dropped` stats in, we still saw the problem on `vmxnet3`, so maybe that's a red-herring, or maybe there are multiple issues here.

guyharris:
> At least one issue appears to be that [linux_if_drops()](https://github.com/the-tcpdump-group/libpcap/blob/master/pcap-linux.c#L771-L777) does not report the value of `rx_dropped`, which indicates the number of packets [not forwarded to the upper layers for processing](https://github.com/torvalds/linux/blob/6ba59ff4227927d3a8530fc2973b80e94b54d58f/Documentation/ABI/testing/sysfs-class-net-statistics#L44-L52).

That page lists *several different* receive errors.  `rx_dropped` is one of the more vaguely described ones:

> Indicates the number of packets received by the network device but dropped, that are not forwarded to the upper layers for packet processing.

Some others, even though they say "See the network driver for the exact meaning of this value.", are a bit more specific, such as:

- `rx_crc_errors`, which "Indicates the number of packets received with a CRC (FCS) error by this network device.", and which, though it states that "the specific meaning might depend on the MAC layer used by the interface.", a CRC error is a CRC error, which isn't a "lack of resources", it's a separate problem;
[...]

PR #1329, opened 2024-06-27
Account for rx_dropped packets in linux ifdropped

gtedesco-r7:
Many Linux drivers are not reporting anything into `rx_missed_errors` or `rx_fifo_errors`. They, instead, report dropped packets into `rx_dropped`.

Without including `rx_dropped` into `ps_ifdrop`, we can get a wildly optimistic view of performance.

This particularly affects vmxnet3, enic, ena, cxgb, and cxgb4 drivers.

Issue described in: #1328

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -771,9 +771,10 @@ linux_get_stat(const char * if_name, const char * stat) {
 static long long int
 linux_if_drops(const char * if_name)
 {
+	long long int dropped = linux_get_stat(if_name, "rx_dropped");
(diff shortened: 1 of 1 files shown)


infrastation:
Thank you for waiting. I have checked a few Linux hosts with a variety of 100Base-TX and 1000Base-T physical interfaces: in about half the cases `rx_dropped` was not zero, but `rx_missed_errors` was zero everywhere, so this seems to be a useful change for libpcap 1.11.1. It will need rebasing and a change log entry, but that would be trivial.

PR #1339, opened 2024-07-29
Fix out-of-bound pointer subtraction in BPF buffers

nathaniel-bennett:
The current implementation for BPF uses the following to determine if a buffer has been fully read:
```
    p->cc = (int)(ep - bp);
```
(where bp and ep are char* pointers.)

As stated in the source code, ep is set based on `read()`, which may not return a value that's a multiple of the alignment value for `BPF_WORDALIGN()`. This means that when bp is incremented by `caplen + hdrlen` padded to word alignment, it may point beyond the end of the buffer, such that bp > ep. Since subtraction is only well-defined for pointers that are both within a contiguous memory region (up to 1 past the end of a buffer), this results in undefined behavior.

This pull request resolves these issues by limiting how much bp can be incremented by so that it never exceeds ep. This ensures the pointer subtraction is well-defined and preserves the behavior of `p->cc` being assigned to 0 in such cases.

### DIFF
--- a/pcap-bpf.c
+++ b/pcap-bpf.c
@@ -1333,28 +1333,13 @@ pcap_read_bpf(pcap_t *p, int cnt, pcap_handler callback, u_char *user)
 		if (p->break_loop) {
 			p->bp = bp;
 			p->cc = (int)(ep - bp);
-			/*
-			 * ep is set based on the return value of read(),
-			 * but read() from a BPF device doesn't necessarily
-			 * return a value that's a multiple of the alignment
-			 * value for BPF_WORDALIGN().  However, whenever we
-			 * increment bp, we round up the increment value by
-			 * a value rounded up by BPF_WORDALIGN(), so we
-			 * could increment bp past ep after processing the
-			 * last packet in the buffer.
-			 *
-			 * We treat ep < bp as an indication that this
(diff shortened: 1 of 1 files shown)


guyharris:
> ep is set based on `read()`, which may not return a value that's a multiple of the alignment value for BPF_WORDALIGN()

If the return value of `read()` isn't a multiple of the alignment value, presumably that's because the last packet doesn't need to be padded to align the address of the next packet, because there *is* no next packet.  And, if we're processing the last packet, there's no need to increment bp to point to the non-existent next packet; we should just say "OK, that's it, no more packets to process in this buffer.

That might be a bit clearer.

guyharris:
The code has only one loop variable, `bp`; it doesn't use `cc` as a loop variable as well.  Perhaps the goal, at the time, was to micro-optimize the loop, but what might have been a useful optimization some time between 1988 and 1992 might not be one now.

Having `cc` be a loop variable might be clearer, and avoid this issue.

guyharris:
And pcap-bpf.c isn't the only module that has this issue.  See #1343 for a pull request that 1) uses `cc` as a loop variable and 2) fixes the other modules where it's an issue.

guyharris:
Included in now-merged #1343.

PR #1340, opened 2024-07-30
skip snprintf check for cross compile in cmake

junka:
just like what we did in configure.ac, skip check snprintf for cross compile.


junka:
> I understand on a Unix-like system with CMake 2.8.12 `CMAKE_CROSSCOMPILING` will not be defined and the workaround will not work, but will not affect other things.

Hi denis,
Any better solution or what should I do for this pr.

PR #1351, opened 2024-09-17
Call WSAStartup and WSACleanup for each use. Fixes #1333

### LINKED ISSUE #1333: Application Verifier stops in libpcap (wpcap.dll) included with Npcap 1.79
Hi Libpcap team!  @guyharris suggested we create this issue corresponding to our [npcap#742](https://github.com/nmap/npcap/issues/742).  We received a customer report that Application Verifier has two stops in wpcap.dll (which is all Libpcap 1.10.4 code) included in Npcap 1.79.  We're working to reproduce and debug and will update this with any further information we find.  And of course we appreciate anything you're able to determine.  Here's the customer report:

> Recently we tried to run our application (x64) with Application Verifier. With the "Networking" checks enabled, 'wpcap.dll' causes two verifier stops (https://learn.microsoft.com/en-us/windows-hardware/drivers/devtest/application-verifier-stop-codes-networking). We would like to enable these checks during our automated test runs. It's currently not possible due to these verifier stops.
> 
> The first one is calling 'pcap_findalldevs_ex' causing "A Winsock API was called before a successful WSAStartup() or after a balancing successful WSACleanup() call was made". Application Verifier seems to be tracking these calls on a per module basis. Even if 'main()' had called 'WSAStartup' before calling 'pcap_findalldevs_ex', the stop message is still issued.
>
> The second one happens after 'main()' had exited. When 'wpcap.dll' is unloading the Verifier issues "Illegal networking API called from DllMain".

dmiller-nmap:
`WSACleanup` cannot be called from `atexit()` because that's run in `DllMain` context. Instead, start and cleanup at each use, using existing reference counting functions from `sockutils.c`.

Performance impact ought to be small because these are not frequently-called functions in the lifetime of a capture handle, and because `WSAStartup`/`WSACleanup` are essentially just reference increment/decrement in the common case of an application that has already called `WSAStartup`.

I believe this commit covers all likely situations, with one unlikely exception: the functions in `nametoaddr.c` are exported by the library, though they are not documented in the manpages. They are only called internally from functions within `gencode.c` which should be handled by the startup/cleanup calls in `pcap_compile()`, and I have not found any applications on Github that call them directly. If any application had previously called them directly, relying on `pcap_init()` to call `WSAStartup()`, then that code would need to be changed to call `WSAStartup()` itself. I couldn't just add startup/cleanup within each function in `nametoaddr.c` because several of them return static buffers allocated by Winsock functions, which I assume could be invalidated by calling `WSACleanup()`.

### DIFF
--- a/rpcapd/log.c
+++ b/rpcapd/log.c
@@ -26,6 +26,9 @@
 #include <stdlib.h>
 
 #ifdef _WIN32
+/* Prevent inclusion of winsock.h, which causes redefinition errors when
+ * winsock2.h is included later */
+#define WIN32_LEAN_AND_MEAN
 #include <windows.h>
 #else
 #include <syslog.h>
--- a/gencode.c
+++ b/gencode.c
@@ -726,9 +726,6 @@ int
 pcap_compile(pcap_t *p, struct bpf_program *program,
 	     const char *buf, int optimize, bpf_u_int32 mask)
 {
-#ifdef _WIN32
-	static int done = 0;
-#endif
 	compiler_state_t cstate;
 	const char * volatile xbuf = buf;
 	yyscan_t scanner = NULL;
@@ -746,12 +743,9 @@ pcap_compile(pcap_t *p, struct bpf_program *program,
 		return (PCAP_ERROR);
 	}
 
-#ifdef _WIN32
-	if (!done) {
-		pcap_wsockinit();
-		done = 1;
+	if(0 != pcapint_sockinit()) {
+		return (PCAP_ERROR);
 	}
-#endif
 
 #ifdef ENABLE_REMOTE
 	/*
@@ -874,6 +868,8 @@ quit:
 	 * Clean up our own allocated memory.
 	 */
 	freechunks(&cstate);
+
+	pcapint_sockcleanup();
 
 	return (rc);
 }
--- a/pcap-npf.c
+++ b/pcap-npf.c
@@ -1080,9 +1080,6 @@ pcap_activate_npf(pcap_t *p)
 		}
 	}
 
-	/* Init Winsock if it hasn't already been initialized */
-	pcap_wsockinit();
-
 	pw->adapter = PacketOpenAdapter(p->opt.device);
 
 	if (pw->adapter == NULL)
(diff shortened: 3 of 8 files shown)


guyharris:
> using existing reference counting functions from sockutils.c

libpcap, even on Windows, can be compiled without remote capture support.  It *defaults* to building with that support on Windows, because WinPcap had it, but it should support building without it.

So perhaps we should have `static inline` `pcapint_sockinit()` and `pcapint_sockcleanup()` functions in portability.h that call `WSAStartup()` and `WSACleanup()`, respectively, on Windows, and do nothing elsewhere ("elsewhere" includes Haiku as well as UN\*Xes).  Those function would be the same as `sock_init()` and `sock_cleanup()`, except that they wouldn't maintain their own use count, but would unconditionally call `WSAStartup()` and `WSACleanup()`, and would replace `sock_init()` and `sock_cleanup()`.

As for the nametoaddr.c functions, as per your suggestion, let's let `pcap_compile()` handle the initialization for them for now.  I might be tempted to deprecate those functions, with a recommendation to "just do it yourself", to try to flush out any programs that use them.

dmiller-nmap:
I believe this is ready for review again. The linux-armv7l build failure does not appear to be related to my changes.

rpcapd/log.c
```diff
@@ -26,6 +26,7 @@
 #include <stdlib.h>
 
 #ifdef _WIN32
+#define WIN32_LEAN_AND_MEAN
```
fxlb, line 29: You should add a comment like `exclude some of the less common API declarations`.

dmiller-nmap: Will do. This was necessary because some of the appveyor builds were failing when Winsock2.h was included where windows.h had already included winsock.h.

fxlb: > This was necessary because some of the appveyor builds were failing when Winsock2.h was included where windows.h had already included winsock.h.

This should be explained in a comment.

dmiller-nmap:
Comment updated. The build errors look like network transient errors. I confirmed that the Npcap SDK is accessible.

guyharris:
> The build errors look like network transient errors. I confirmed that the Npcap SDK is accessible.

Appveyor is... not the most reliable build server.  It frequently has a problem either downloading stuff from Chocolatey or downloading the Npcap SDK; I've restarted the build for this patch request (as well as the build for another patch request or change to tcpdump)..

rpcapd/log.c
guyharris: The comment now includes @dmiller-nmap's explanation; @fxlb, does that suffice?

Issue #172, opened 2013-04-15
configure does not check for netlink/genl/genl.h


guyharris:
Converted from [SourceForge issue 3565777](http://sourceforge.net/support/tracker.php?aid=3565777), submitted by roflz

./pcap-linux.c:267:31: fatal error: netlink/genl/genl.h: No such file or directory

configure should check for its existence, and if it is needed, abort there

rofl0r:
nice. may i ask which release flavor you used for testing - img file for qemu or just the precompiled rootfs extracted and used via chroot, bwrap, or other sandboxing tool ?

PR #1361, opened 2024-09-29
bug fixes related to libnl-tiny (#172)

infrastation:
Looks good to me.


ftmacros.h
```diff
@@ -100,8 +100,12 @@
    * We don't limit this to, for example, Linux and Cygwin, because
    * this might, for example, be GNU/HURD or one of Debian's kFreeBSD
    * OSes ("GNU/FreeBSD").
+   *
+   * Avoid redefining, same as below.
```
rofl0r, line 104: a trick to make this one line shorter i came up with (also used elsewhere) is
```
#undef FOO 
#define FOO
```

PR #1371, opened 2024-10-14
Remove code related to obsolete HW AirPcap

AndersBroman:
The Riverbed® AirPcap Product has reached End of Availability (EOA) as of December 31, 2017 and it only supports older Windows versions.
https://support.riverbed.com/content/support/software/steelcentral-npm/airpcap.html


infrastation:
@guyharris, these changes among other things remove the Windows-specific `get_airpcap_handle_op_t` member from `struct pcap`, would this require any backward compatibility provisions regarding NPF?

infrastation:
These changes also remove a public Windows-specific function from the API:
```c
PCAP_API PAirpcapHandle pcap_get_airpcap_handle(pcap_t *p);
```
This needs to be considered before merging.

guyharris:
> These changes also remove a public Windows-specific function from the API:
> 
> ```
> PCAP_API PAirpcapHandle pcap_get_airpcap_handle(pcap_t *p);
> ```
> 
> This needs to be considered before merging.

That function should not be removed, so as not to break the ABI.

Instead, it should should probably be modified to fail with an "AirPcap is not supported" error and marked as deprecated.

guyharris:
> @guyharris, these changes among other things remove the Windows-specific `get_airpcap_handle_op_t` member from `struct pcap`, would this require any backward compatibility provisions regarding NPF?

No - neither Npcap nor WinPcap uses that field.

pcap.c
```diff
@@ -3967,19 +3952,6 @@ pcap_live_dump_ended(pcap_t *p, int sync)
 {
 	return (p->live_dump_ended_op(p, sync));
 }
-
```
guyharris, line 3970: `pcap_get_airpcap_handle()` is part of the ABI, so it shouldn't be removed.  Instead, it should always set `p->errbuf` to a message that indicates the AirPcap devices aren't supported and return NULL.

pcap/pcap.h
```diff
@@ -824,14 +824,6 @@ PCAP_API const char *pcap_lib_version(void);
 
   typedef struct pcap_send_queue pcap_send_queue;
 
-  /*!
```
guyharris, line 827: This should also not be removed, so that `pcap_get_airpcap_handle()` can be declared.

.appveyor.yml
```diff
@@ -32,62 +32,51 @@ environment:
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
       GENERATOR: "Visual Studio 14 2015 Win64"
       SDK: WpdPack
-      AIRPCAP: -DDISABLE_AIRPCAP=YES
```
guyharris, line 35: The number of test cases here should be reduced, with all tests not done with `-DDISABLE_AIRPCAP=YES` removed.

pcap/pcap.h
```diff
@@ -866,8 +858,6 @@ PCAP_API const char *pcap_lib_version(void);
 
   PCAP_API int pcap_start_oem(char* err_str, int flags);
 
-  PCAP_API PAirpcapHandle pcap_get_airpcap_handle(pcap_t *p);
```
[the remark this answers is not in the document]
AndersBroman: Could you please have a look at the latest push, I've been fighting with git :-)

pcap.c
```diff
@@ -3956,16 +3941,12 @@ pcap_live_dump_ended(pcap_t *p, int sync)
 }
 
 PAirpcapHandle
-pcap_get_airpcap_handle(pcap_t *p)
+pcap_get_airpcap_handle(pcap_t* p)
 {
-	PAirpcapHandle handle;
+	(void)snprintf(p->errbuf, sizeof(p->errbuf),
+		"AirPcap device no longer supported");
```
guyharris, line 3947: I'd say "AirPcap devices are no longer supported".

pcap.c
```diff
@@ -3956,16 +3941,12 @@ pcap_live_dump_ended(pcap_t *p, int sync)
...
 
-	handle = p->get_airpcap_handle_op(p);
-	if (handle == NULL) {
-		(void)snprintf(p->errbuf, sizeof(p->errbuf),
-		    "This isn't an AirPcap device");
-	}
-	return (handle);
+	return NULL;
```
guyharris, line 3949: Small stylistic tweak: `return (NULL)`.

pcap/pcap.h
```diff
@@ -825,12 +825,12 @@ PCAP_API const char *pcap_lib_version(void);
   typedef struct pcap_send_queue pcap_send_queue;
 
   /*!
-    \brief This typedef is a support for the pcap_get_airpcap_handle() function
-  */
-  #if !defined(AIRPCAP_HANDLE__EAE405F5_0171_9592_B3C2_C19EC426AD34__DEFINED_)
-    #define AIRPCAP_HANDLE__EAE405F5_0171_9592_B3C2_C19EC426AD34__DEFINED_
-    typedef struct _AirpcapHandle *PAirpcapHandle;
-  #endif
+  \brief This typedef is a support for the pcap_get_airpcap_handle() function
```
guyharris, line 828: I'd leave the indentation alone, to indicate that this is only done if `_WIN32` is defined, and keep the `*` styling the same.

pcap/pcap.h
```diff
@@ -866,7 +866,9 @@ PCAP_API const char *pcap_lib_version(void);
 
   PCAP_API int pcap_start_oem(char* err_str, int flags);
 
-  PCAP_API PAirpcapHandle pcap_get_airpcap_handle(pcap_t *p);
+  PCAP_AVAILABLE_1_10
```
guyharris, line 869: This API has always been available in the versions of tcpdump.org's libpcap for Windows; as it's now a deprecated API, I wouldn't bother decorating it with a `PCAP_AVAILABLE_` tag.

I'd also leave the `*` styling as is.

.appveyor.yml
```diff
@@ -99,6 +86,4 @@ build_script:
   - md build
   - cd build
   - cmake --version
-  - if NOT DEFINED PLATFORM cmake %AIRPCAP% %REMOTE% -DPacket_ROOT=c:\projects\libpcap\Win32\%SDK% -G"%GENERATOR%" -DPacket_ROOT=c:\projects\libpcap\Win32\%SDK% -DPacket_ROOT=c:\projects\libpcap\Win32\%SDK% -G"%GENERATOR%" -DAirPcap_ROOT=c:\projects\libpcap\Win32\Airpcap_Devpack -G"%GENERATOR%" ..
```
guyharris, line 102: These are the lines that run CMake to configure the build, and cannot and must not be removed; otherwise, the build fails, as it's doing on AppVeyor.

Just remove `%AIRPCAP%` from the arguments to CMake.

.appveyor.yml
```diff
@@ -99,6 +69,5 @@ build_script:
...
-  - if DEFINED PLATFORM cmake %AIRPCAP% %REMOTE% -DPacket_ROOT=c:\projects\libpcap\Win32\%SDK% -G"%GENERATOR%" -DPacket_ROOT=c:\projects\libpcap\Win32\%SDK% -DPacket_ROOT=c:\projects\libpcap\Win32\%SDK% -G"%GENERATOR%" -DAirPcap_ROOT=c:\projects\libpcap\Win32\Airpcap_Devpack -G"%GENERATOR%" -A %PLATFORM% ..
```
guyharris, line 103: This one shouldn't be removed, either.

This is really an if/else case, but the AppVeyor YAML doesn't support that; it does two different commands so that it can include `-A` for non-default platforms (i.e., other than IA-32) and not include it for the default platform (IA-32).

Defining `PLATFORM` as either an empty string or as "-A <whatever>" might allow that to be done in one command, but that's unrelated to dropping AirPcap support, and should probably be done in a separate change, so put back the `if DEFINED PLATFORM`... line, removing ` %AIRPCAP` from it as well (complete with removing the blank, as well as removing it from the `if NOT DEFINED PLATFORM`line, so there aren't two spaces in a row.

guyharris:
`buffer`, and the files in the `build` directory, should not be checked into the repository.

pcap.c
```diff
@@ -3958,14 +3943,10 @@ pcap_live_dump_ended(pcap_t *p, int sync)
 PAirpcapHandle
 pcap_get_airpcap_handle(pcap_t *p)
 {
-	PAirpcapHandle handle;
+	(void)snprintf(p->errbuf, sizeof(p->errbuf),
+		"AirPcap devices are no longer supported"");
```
guyharris, line 3947: Three's an extra `"` after the string.

.appveyor.yml
guyharris: The AppVeyor build is fixed.

.appveyor.yml
```diff
@@ -32,62 +30,34 @@ environment:
...
       GENERATOR: "Visual Studio 15 2017"
       SDK: WpdPack
-    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
-      GENERATOR: "Visual Studio 15 2017 Win64"
-      SDK: WpdPack
-      AIRPCAP: -DDISABLE_AIRPCAP=YES
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
       GENERATOR: "Visual Studio 15 2017"
       SDK: npcap-sdk
-      AIRPCAP: -DDISABLE_AIRPCAP=YES
```
guyharris, line 53: This one should not be removed; it does an x86-64 build.  `-DDISABLE_AIRPCAP=YES` should be removed, however.

.appveyor.yml
```diff
@@ -32,62 +30,34 @@ environment:
...
-    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
-      GENERATOR: "Visual Studio 15 2017 Win64"
-      SDK: npcap-sdk
-      AIRPCAP: -DDISABLE_AIRPCAP=YES
-    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
```
guyharris, line 58: This one should be removed, as it's doing an AirPcap build.

.appveyor.yml
```diff
@@ -32,62 +30,34 @@ environment:
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
       GENERATOR: "Visual Studio 14 2015 Win64"
       SDK: WpdPack
-      AIRPCAP: -DDISABLE_AIRPCAP=YES
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
       GENERATOR: "Visual Studio 14 2015"
       SDK: npcap-sdk
-    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
```
guyharris, line 39: This shouldn't be removed, as it's doing an x86-64 build.

.appveyor.yml
```diff
@@ -32,62 +30,34 @@ environment:
...
-      GENERATOR: "Visual Studio 14 2015 Win64"
-      SDK: npcap-sdk
-      AIRPCAP: -DDISABLE_AIRPCAP=YES
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
       GENERATOR: "Visual Studio 15 2017"
       SDK: WpdPack
-    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
```
guyharris, line 46: This shouldn't be removed, as it's doing an x86-64 build.

AndersBroman:
Thanks for all the help Guy!
Best regards
Anders

PR #1383, opened 2024-11-12
CMake: fix to support CMake 3.31.

guyharris:
1) remove SOURCE parameter from add_custom_command().

It was never a valid parameter, but, prior to CMake 3.31, it was silently ignored.  In 3.31 and later, it's treated as an error by default.

See
https://github.com/the-tcpdump-group/libpcap/pull/1380#issuecomment-2471731576
for a case where this breaks something.

2) expand the "still support old versions of CMake, for the benefit of people with older versions, but avoid deprecation warnings/errors" section to handle the deprecation of pre-3.10 versions by 3.31.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,14 +13,21 @@ else(WIN32)
     #  on a "long-term support" version # of some OS and that
     #  version supplies an older version of CMake;
     #
-    #  otherwise, require 3.5, so we don't get messages warning
-    #  that support for versions of CMake lower than 3.5 is
+    #  otherwise, if it's a version less than 3.10, require only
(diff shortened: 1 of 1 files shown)


infrastation:
Thank you. A change log entry would be useful.

guyharris:
> Thank you.

Backported to 1.10 in 134d6494f1f5e701bf69e23bc55bd15d9af01bfb.

> A change log entry would be useful.

Done in 8dbcaf8fbb563e1e377c3aea0446b8df99091e21 in the main branch and 002ea9b5b4956967573cc78d0b46f7044147ed88 in the 1.10 branch.

PR #1386, opened 2024-11-14
Bluetooth: Fix an error message and a notice message for Autoconf

fxlb:
The current Linux Bluetooth development library is: libbluetooth-dev.

[skip ci]


infrastation:
Also known as:
* `bluez-libs-devel` in AlmaLinux 9
* `bluez-dev` in Alpine Linux
* `libbluetooth-devel` in Void Linux

guyharris:
There are ways of determining the distribution on which you're running...

...but:

- they're complicated;
- you might be doing a cross compile for some *other* distribution, e.g. for some embedded distribution;

so I'm not sure it's worth the effort.

PR #1389, opened 2024-12-01
Linux: Refine connection status reporting.

infrastation:
This is a disputable change because the new behaviour improves reporting of some interface types, but has potential for making reporting of other types worse. The reason for that is that many logical interfaces pretend to be Ethernet. Perhaps it will be easier to understand the best course of action after studying at various examples.

Let's make this a draft for now.

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -103,6 +103,10 @@
   // Linux before 3.5
   #define ARPHRD_IEEE802154_MONITOR 805
 #endif
+#ifndef ARPHRD_IP6GRE
+  // Linux before 3.7
+  #define ARPHRD_IP6GRE 823
+#endif
 #ifndef ARPHRD_NETLINK
   // Linux before 3.11
   #define ARPHRD_NETLINK 824
@@ -1828,17 +1832,22 @@ get_if_flags(const char *name, bpf_u_int32 *flags, char *errbuf)
 				switch (arptype) {
 
 				case ARPHRD_LOOPBACK:
+				case ARPHRD_CAN:
+				case ARPHRD_TUNNEL:
+				case ARPHRD_TUNNEL6:
+				case ARPHRD_SIT:
+				case ARPHRD_IPGRE:
+				case ARPHRD_IP6GRE:
 					/*
 					 * These are types to which
 					 * "connected" and "disconnected"
 					 * don't apply, so don't bother
 					 * asking about it.
-					 *
-					 * XXX - add other types?
 					 */
 					close(sock);
 					fclose(fh);
 					free(pathstr);
+					*flags |= PCAP_IF_CONNECTION_STATUS_NOT_APPLICABLE;
 					return 0;
 
 				case ARPHRD_IRDA:
@@ -1886,13 +1895,8 @@ get_if_flags(const char *name, bpf_u_int32 *flags, char *errbuf)
 			/*
 			 * OK, this OS version or driver doesn't support
 			 * asking for this information.
-			 * XXX - distinguish between "this doesn't
-			 * support ethtool at all because it's not
-			 * that type of device" vs. "this doesn't
-			 * support ethtool even though it's that
-			 * type of device", and return "unknown".
 			 */
-			*flags |= PCAP_IF_CONNECTION_STATUS_NOT_APPLICABLE;
+			*flags |= PCAP_IF_CONNECTION_STATUS_UNKNOWN;
 			close(sock);
 			return 0;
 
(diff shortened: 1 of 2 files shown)


JonnyBnice84 (approved):

killerdevildog (commented):
What This Change Does
The PR makes two key improvements to how libpcap reports network interface connection status on Linux:

Expands "Not Applicable" Interface Types: Adds several tunnel and CAN interface types (ARPHRD_CAN, ARPHRD_TUNNEL, ARPHRD_TUNNEL6, ARPHRD_SIT, ARPHRD_IPGRE, ARPHRD_IP6GRE) to the list that should report PCAP_IF_CONNECTION_STATUS_NOT_APPLICABLE instead of trying to check their connection status.

Improves Error Handling: Changes the fallback behavior from PCAP_IF_CONNECTION_STATUS_NOT_APPLICABLE to PCAP_IF_CONNECTION_STATUS_UNKNOWN when ethtool fails with EOPNOTSUPP or EINVAL.

Fixes Missing Status Flag: Previously, interfaces marked as "not applicable" weren't actually setting the status flag - this was a bug that's now fixed.

Technical Soundness
✅ The changes are technically sound:

Correct Interface Classification: The added interface types (CAN, various tunnels) genuinely don't have meaningful "connected/disconnected" states
Better Error Semantics: UNKNOWN vs NOT_APPLICABLE provides more precise information
Bug Fix: Adding the missing *flags |= PCAP_IF_CONNECTION_STATUS_NOT_APPLICABLE; line fixes a real issue
Backward Compatibility: No breaking changes to the API
Why It's a Draft - The Author's Concerns
The author (@infrastation) marked this as draft because:

"This is a disputable change because the new behaviour improves reporting of some interface types, but has potential for making reporting of other types worse. The reason for that is that many logical interfaces pretend to be Ethernet."

Assessment of the Concerns
The author's concern is somewhat valid but overstated:

Limited Impact: The change only affects error conditions where ethtool fails
More Accurate: UNKNOWN is more honest than NOT_APPLICABLE when we can't determine status
Specific Interface Types: The newly added types are well-defined and genuinely don't have connection status

Recommendations
✅ This PR should move forward with these considerations:

Immediate Actions:
Remove Draft Status: The changes are safe and improve accuracy
Test Edge Cases: Verify behavior on systems with various tunnel interfaces
Documentation: Consider updating any documentation about connection status reporting

pcap-linux.c
```diff
@@ -1828,17 +1832,22 @@ get_if_flags(const char *name, bpf_u_int32 *flags, char *errbuf)
...
+				case ARPHRD_TUNNEL6:
+				case ARPHRD_SIT:
+				case ARPHRD_IPGRE:
+				case ARPHRD_IP6GRE:
 					/*
 					 * These are types to which
 					 * "connected" and "disconnected"
 					 * don't apply, so don't bother
 					 * asking about it.
-					 *
```
guyharris, line 1836: Are there any Linux `ARPHRD_` types that we're *not* explicitly handling, i.e. we're handling it by doing nothing in the (implicit) default case?

pcap-linux.c
```diff
@@ -1886,13 +1895,8 @@ get_if_flags(const char *name, bpf_u_int32 *flags, char *errbuf)
 			/*
 			 * OK, this OS version or driver doesn't support
 			 * asking for this information.
-			 * XXX - distinguish between "this doesn't
```
guyharris, line 1889: What do `EOPNOTSUPP` and `EINVAL` indicate here? Does that make the distinction in question?

infrastation:
@killerdevildog, third warning for posting software-generated comments and ban.

pcap-linux.c
infrastation: In this switch block? Yes, for example, `ARPHRD_ETHER` and `ARPHRD_NETLINK` (arguably, the latter may make sense in this case too, but I didn't test that). The default results in `PCAP_IF_CONNECTION_STATUS_UNKNOWN`.

pcap-linux.c
infrastation: As far as the comment says it, one (not sure which one) means "ethtool is not supposed to work with this interface" and the other means "ethtool is supposed to work with this interface, but it could not". Perhaps the final version should split the comments into two cases, or do something else that would be more sensible than removing the comment.

AFAIR, the intent of this chunk is to have a more sensible relation between interfaces that have a notion of the connection status and interfaces that are in ethtool domain. As the pull request comment says, it turned out to be a can of worms bigger than expected, so **this change is not safe for merging** and should be reworked to some extent. Also this would need some testing on 2.6 kernels to see how much of the semantics holds.

Also bitwise OR with `PCAP_IF_CONNECTION_STATUS_UNKNOWN` is a no-op, as I see now.

PR #1392, opened 2024-12-04
CI: add an ARM64 build for AppVeyor.

guyharris:
(Test only on AppVeyor, so:)

[skip ci]

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -99,6 +99,11 @@ environment:
       PLATFORM: x64
       SDK: npcap-sdk
       REMOTE: -DENABLE_REMOTE=YES
+    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2022
+      GENERATOR: "Visual Studio 17 2022"
+      PLATFORM: ARM64
+      SDK: npcap-sdk
+      REMOTE: -DENABLE_REMOTE=NO
 
 build_script:
   #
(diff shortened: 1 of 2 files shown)


guyharris:
> Why an update to gencode.c for an ARM64 build?

As per the updated commit message.  The volatiles are only there due to the setjmp/longjmp, and aren't being accessed from multiple threads, so the [memory model](https://en.wikipedia.org/wiki/Memory_model_(programming)) shouldn't matter here.

guyharris:
Most if not all of this is unintentional infinite recursion, fixed in 9f368821e10e32de9fbd7aaea3e01c25c0086d48.

Next step - have the libpcap buildbot builds compile the test programs and run tests, starting with the simple test `testprogs/filtertest EN10MB`, which would have caught this *before* it was committed and used in tcpdump tests.

guyharris:
> Next step - have the libpcap buildbot builds compile the test programs and run tests, starting with the simple test `testprogs/filtertest EN10MB`, which would have caught this _before_ it was committed and used in tcpdump tests.

That's pull request #1394.

infrastation:
I thought the tcpdump segfault on freebsd-aarch64 was a botched FreeBSD upgrade, and was going to reinstall it from scratch.

guyharris:
> I thought the tcpdump segfault on freebsd-aarch64 was a botched FreeBSD upgrade, and was going to reinstall it from scratch.

Sorry about that. That was my first guess when I saw the issue, but when I saw that it'd hit every buildbot, I ruled a system problem out.

PR #1394, opened 2024-12-05
CI: on UN*Xes, do a smoke test of the filter compiler.

guyharris:
Run "filtertest EN10MB", which makes sure that the filter compiler isn't completely broken, as it was after
df8296449a3ae6db4fd50cd0ecd32f595f14f74c; that wasn't detected by the buildbots until it was committed to the repository and tested in tcpdump builds triggered by the libpcap commit.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -96,9 +96,11 @@ run_after_echo "$PREFIX/bin/pcap-config" --additional-libs --static-pcap-only
 # VALGRIND_CMD is meant either to collapse or to expand.
 # shellcheck disable=SC2086
 if [ "$CMAKE" = no ]; then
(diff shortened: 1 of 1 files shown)


infrastation:
Within the last couple years there was a discussion somewhere, which considered having libpcap tests that would take a DLT and an input filter expression and compare the actual output of `filtertest` with the expected output. I remember offering someone to implement that, and my draft files towards this are dated 2023-03-18. Is it time to revive the draft?

guyharris:
> Is it time to revive the draft?

Probably. We have to remember to update it if the compiler changes to produce different code, but that's just like "make check" for tcpdump.

infrastation:
Alright, working on that now.

infrastation:
This is pull request #1399 now.

PR #1395, opened 2024-12-05
CMake: don't build with sslutils.c if you don't have a TLS library.

guyharris:
Some versions of Visual Studio's C compiler report a warning when compiling an empty translation unit, which is what sslutils.c is if we're building without a TLS library.


infrastation:
`make whitespacecheck` did not pass (Cirrus CI does not send notifications, unfortunately).

guyharris:
> `make whitespacecheck` did not pass (Cirrus CI does not send notifications, unfortunately).

Fixed.

PR #1397, opened 2024-12-08
autotools, CMake: fix issues with snprintf test and sanitizers.

### LINKED ISSUE #1396: configure problems with checks on snprintf() when using some sanitize options

On Debian 12:
```
CC="clang-16" CFLAGS='-fsanitize=undefined,integer,signed-integer-overflow -fno-sanitize-recover=undefined,integer -fno-omit-frame-pointer -fno-optimize-sibling-calls -fPIC -g3 -ggdb3 -O0' ./configure
[...]
checking whether snprintf is suitable... no
configure: error: The snprintf(3) implementation in this libc is not suitable,
libpcap would not work correctly even if it managed to compile.

in config.log:
conftest.c:43:37: runtime error: negation of 100 cannot be represented in type 'unsigned long'
    #0 0x55ea26c29867 in main /home/cpe/soft-origin/tcpdump/my-libpcap/conftest.c:43:37

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior conftest.c:43:37
```
Pb with:
`snprintf(buf, sizeof(buf), "%zd", -sizeof(buf));`

If lines 231-233 are commented in configure.ac, after ./autogen.sh, also:
```
[...]
checking whether snprintf is suitable... no
configure: error: The snprintf(3) implementation in this libc is not suitable,
libpcap would not work correctly even if it managed to compile.

in config.log:
conftest.c:47:42: runtime error: negation of 4294967296 cannot be represented in type 'uint64_t' (aka 'unsigned long')
    #0 0x55f0538e6873 in main /home/cpe/soft-origin/tcpdump/my-libpcap/conftest.c:47:42

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior conftest.c:47:42
```
Pb with:
`snprintf(buf, sizeof(buf), "%" PRId64, -t);`

guyharris:
Avoid trying to cast negative values to unsigned types, or doing shifts of signed types, in order not to have the test program fail if we're building with undefined-behavior sanitizers enabled.

Fix #1396.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -776,32 +776,50 @@ if (NOT CMAKE_CROSSCOMPILING)
     #include <inttypes.h>
     #include <sys/types.h>
 
+    #if defined(_WIN32) && !defined(_SSIZE_T_DEFINED)
+    /*
+     * On UN*Xes, this is a signed integer type of the same size as size_t.
+     *
+     * It's not defined by Visual Studio; we assume that ptrdiff_t will
+     * be a type that is a signed integer type of the same size as size_t.
+     */
+    typedef ptrdiff_t ssize_t;
+    #endif
+
+    /*
+     * Avoid trying to cast negative values to unsigned types, or doing
+     * shifts of signed types, in order not to have the test program fail
+     * if we're building with undefined-behavior sanitizers enabled.
+     */
     int main()
     {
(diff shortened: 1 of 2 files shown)


guyharris:
> Same for tcpdump?

In progress.

PR #1406, opened 2024-12-12
Fix "tcpdump -i <n>" for something-only libpcap builds.

guyharris:
pcap_create() returns NULL for all errors, so tcpdump tries to find evidence of ENODEV in the error buffer in order to retry using a different device name.  That does not work when the error string is "This version of libpcap only supports...", so use a format string that works and place it in pcap-int.h.

```
# tcpdump --version
tcpdump version 5.0.0-PRE-GIT
libpcap version 1.11.0-PRE-GIT (DAG-only)

# tcpdump -D
1.dag0 (alias for dag0:0) [none]
2.dag0:0 (Rx stream 0, 64 MiB, DAG 7.5G2 rev A at 0000:01:00.0) [none]
3.dag1 (alias for dag1:0) [none]
4.dag1:0 (Rx stream 0, 64 MiB, DAG 9.2X2 rev D at 0000:02:00.0) [none]
5.dag16 (alias for dag16:0) [none]
6.dag16:0 (Rx stream 0, 64 MiB, vDAG) [none]
```

Before:

```
# tcpdump -ni 1
tcpdump: This version of libpcap only supports DAG cards
```

After:

```
# tcpdump -ni abcde
tcpdump: No such device (this build of libpcap supports DAG devices only).

# tcpdump -ni 1
listening on dag0, link-type EN10MB (Ethernet), snapshot length 262144 bytes
```

(cherry picked from commit 8a5dcc8840bfbdc3256f4d50d9924b3234342fba)

Also, add some changes backported to the 1.10 branch.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,11 @@
 DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
   Summary for 1.10.6 libpcap release (so far!)
+    Source code:
+      Fix "tcpdump -i <n>" for something-only libpcap builds.
+      Remove some unneeded includes.
+    Packet filtering:
+      Make the chunk allocator's alignment more general and
+        platform-independent.
     Linux:
       Fix check for mac80211 phydev.
       Don't create monitor-mode interface if we're capturing on one.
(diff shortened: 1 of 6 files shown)


infrastation:
The commit message has lost the command because they begin with the comment character (`#`). To preserve the commands, you would need to configure:
```
git config --global commit.cleanup scissors
```

guyharris:
> The commit message has lost the command because they begin with the comment character (`#`). To preserve the commands, you would need to configure:
> 
> ```
> git config --global commit.cleanup scissors
> ```

Done; I'll update the commit message and the pull request message.

Presumably "scissors" means "scan until you find the line with the ASCII-art scissors '>8' and start cutting there".

guyharris:
(And I changed the pull request message to put the sample command and output in fixed-width form, so that the Markdown formatter doesn't think the root prompt means "this is a heading".)

PR #1408, opened 2024-12-13
CI: Remove VS 2015 build, add VS 2022 build.

guyharris:
VS 2015 is approximately 10 years old; get tid of the VS 2015 builds.

Add VS 2022 builds; try doing them with and without external capture support.

Remove include of <wspiapi.h>; it's only needed for IPv6 support on Windows 2000, which is almost 25 years old, and it has code that provokes warnings in VS2022 builds.

In filtertest, don't arrange to call WSACleanup() via atexit(), as that causes a pointer casting warning in VS 2022 builds.

(backported from commit 6ff82f40abdae2e9cfc457d55e0194fb7192838e)

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -29,31 +29,6 @@ environment:
   # For now, we disable the remote capture build there.
   #
   matrix:
-      # MinGW
-    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
-      GENERATOR: "MinGW Makefiles"
-      SDK: WpdPack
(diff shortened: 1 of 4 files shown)


guyharris:
(Do not merge yet; it's waiting for #1410, which is a backport of #1409, which I put into the main branch because, even though, with the lack of AirPcap support, `pcapint_find_function()` isn't used and thus doesn't provoke the warning, it's the Right Thing To Do. This has the changes in #1409/#1410, so that I could test whether it really *does* fix the problem, but I want those changes to be in a separate commit.)

PR #1418, opened 2024-12-15
tests: provide our own version of the timeout command.

guyharris:
That way, we don't have to worry about whether a platform has that command (which is not in the latest POSIX) or, if it does, whether it behaves in a particular fashion.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -46,12 +46,9 @@ macos_task:
     image: ghcr.io/cirruslabs/macos-runner:sonoma # last 3 versions of Xcode
   env:
     MAKEFLAGS: '-j 4' # macOS VMs run on 4 cores
-    # When GNU coreutils package is installed, Autoconf defaults to GNU
-    # install instead of the macOS install, so use the latter explicitly.
-    INSTALL: /usr/bin/install
   script:
     - brew update >/dev/null
-    - brew install openssl@3 coreutils
+    - brew install openssl@3
     - ./build_matrix.sh
 
 coverity_task:
--- a/Makefile.in
+++ b/Makefile.in
@@ -366,6 +366,7 @@ EXTRA_DIST = \
 	testprogs/fuzz/onefile.c \
 	testprogs/nonblocktest.c \
 	testprogs/opentest.c \
+	testprogs/our_timeout.c \
 	testprogs/reactivatetest.c \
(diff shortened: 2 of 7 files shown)


infrastation:
These changes do not undo the installation and use (`TIMEOUT_BIN`) of coreutils on macOS. Also the commit message has three typos.

guyharris:
> These changes do not undo the installation and use (TIMEOUT_BIN) of coreutils on macOS.

That installation didn't appear to prevent a tcpdump build from failing on macOS:

https://github.com/the-tcpdump-group/tcpdump/pull/1256/checks?check_run_id=34430017406

guyharris:
> Also the commit message has three typos.

Fixed.

guyharris:
> > These changes do not undo the installation and use (TIMEOUT_BIN) of coreutils on macOS.
> 
> That installation didn't appear to prevent a tcpdump build from failing on macOS:
> 
> https://github.com/the-tcpdump-group/tcpdump/pull/1256/checks?check_run_id=34430017406

That installation wasn't being done for tcpdump CI in macOS, which is why it was failing. Any macOS CI that builds and runs `make check` on libpcap needs to install Homebrew coreutils.  Fix incoming in the-tcpdump-group/tcpdump#1257.

That fix still means that any *other* place where "make check" is done on libpcap needs to somehow arrange to install the `timeout` command, unless we provide our own.

infrastation:
Arguably, `TESTRUN` could avoid using `timeout` if it is not available and `TIMEOUT_BIN` is not set (i.e. the user does not mind how long the test takes so long as it passes).

infrastation:
This would mean the test timeout would apply in "other places" by default on most Linuxes, all BSDs, illumos, Solaris CBE and Haiku, and in this development CI setup `timeout` on macOS would be either made available or not used. Would that be better?

guyharris:
> This would mean the test timeout would apply in "other places" by default on most Linuxes, all BSDs, illumos, Solaris CBE and Haiku, and in this development CI setup `timeout` on macOS would be either made available or not used. Would that be better?

That'd work.

guyharris:
Dropped in favor of the-tcpdump-group/tcpdump#1257, which installs GNU coreutils with Homebrew on macOS, so that `gtimeout` is available, and #1421, which avoids using `timeout` if it's not available.

PR #1419, opened 2024-12-15
CI: set environent variables for Homebrew on macOS. [skip appveyor]

guyharris:


guyharris:
The problem with macOS Cirrus CI builds of tcpdump is in tcpdump, which wasn't installing coreutils on macOS; the *libpcap* builds don't have a problem. Closing.

PR #1421, opened 2024-12-16
Make the timeout dependency conditional.

infrastation:
As discussed in #1418.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -46,12 +46,9 @@ macos_task:
     image: ghcr.io/cirruslabs/macos-runner:sonoma # last 3 versions of Xcode
   env:
     MAKEFLAGS: '-j 4' # macOS VMs run on 4 cores
-    # When GNU coreutils package is installed, Autoconf defaults to GNU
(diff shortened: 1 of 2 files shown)


infrastation:
@guyharris, I am going to leave libpcap on macOS with coreutils because it is in a better position to flex different build options that can potentially degrade performance. Then tcpdump build matrix can run without coreutils to prove that both libpcap and tcpdump are able to build without coreutils. Let me know if you see a better way to arrange this.

infrastation:
(Alternatively, both could do without coreutils if there is no factor that would make such regressions macOS-specific.)

infrastation:
Alright, removing coreutils from macOS.

infrastation:
For posterity, on macOS the updated script prints diagnostics as follows.

With coreutils:
```
$ testprogs/TESTrun
INFO: Using a test timeout of 1.
```

Without coreutils:
```
$ testprogs/TESTrun
Can't exec "timeout": No such file or directory at testprogs/TESTrun line 1318.
WARNING: Not using a test timeout (the default 'timeout' is not usable).
```

PR #1426, opened 2025-01-04
dlpi: fix open_dlpi_device() to handle Solaris/HP-UX logical interfaces.

guyharris:
If handed the name of a logical interface, strip off the logical interface number.  This means that in pcap_findalldevs() we won't skip the logical interfaces and will get the network addresses and netmasks they have.

(It also mean that if you try to capture on a logical interface, it'll capture on the physical interface, but that might be the right thing to do.)

We can't do it earlier in fad-gifc.c and fad-glifc.c, because we have to use the logical interface name when fetching the netmask, etc..

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -197,6 +197,8 @@ Friday, August 30, 2024 / The Tcpdump Group
       Allow attaching to links owned by a non-global zone.  (Based on
         pull request #1202.)
       Fix AF_LINK handling on illumos.
+      Fix not to ignore logical interfaces in fad-gifc.c and
+        fad-glifc.c.
     macOS:
       Redid the availability macros to be closer to what Apple's doing
         in recent SDKs, including tagging pcap-namedb.h routines.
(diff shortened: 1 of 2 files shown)


infrastation:
The change is in the 1.10.5 section.

guyharris:
> The change is in the 1.10.5 section.

The change - and the modification to CHANGES - has been backported to 1.10.5 in 618cc5d1bf59c17f021f523c5bd5442984c085ba.

infrastation:
Then it should be in the to-be-1.10.6 section, not 1.10.5.

guyharris:
> Then it should be in the to-be-1.10.6 section, not 1.10.5.

Ouch.  Sorry about that - I missed that it was 1.10.5.

Fixed in 4c69395478f5cef3ebec3e08d8b0c3926b853dc5 (main) and 86e01ccdebf0a136367124cf7461cbc3ce5debe7 (1.10).

infrastation:
Thank you.

PR #1428, opened 2025-01-07
Support both "igrp" and "eigrp" in filters.

guyharris:
Use our own header to define IPPROTO_ values, rather than leaving them up to the OS for which we're building.

Always use IP protocol 9 for "igrp", and use IP protocol 88 for "eigrp". 9 is assigned to "any private Internet gateway protocol", and is mainly used for Cisco's IGRP.  88 is for Cisco's EIGRP, which is not compatible on-the-wire with IGRP.

FreeBSD, DragonFly BSD and macOS define IPPROTO_IGRP as 88, rather than as 9, and their libpcaps use IP protocol 88 for "igrp".  Everybody else uses IP protocol 9 for "igrp".


guyharris:
This is one way to deal with the "igrp"/"eigrp" issues that b33f396845b3e37a8ffeda8956e090b656bb8433 works around.

It makes libpcap's filter compiler, when compiled in FreeBSD, DragonFly BSD, or macOS work differently for a filter `igrp` than do the FreeBSD, DragonFly BSD, and macOS libpcap filter compilers do - they translate `igrp` as a test for IP protocol 88, not IP protocol 9. All other systems will translate `igrp` as a test for IP protocol 9.

infrastation:
This is one of five (so far!) inconsistencies that would require changing the meaning of some keywords or removing/adding keywords to be solved properly. Before making any changes I had in mind to:
1. Post a detailed account of these issues on the list.
2. Document the imperfections thoroughly in libpcap 1.11.x.
3. Suggest addressing all issues at once in libpcap 2.0.0, so the users have to verify their filters only once and after an advance warning.

For this particular problem a potential solution could be replacing "eigrp" with the standard protocol names ("igp" for 9 and "eigrp" for 88).

guyharris:
> For this particular problem a potential solution could be replacing "eigrp" with the standard protocol names ("igp" for 9 and "eigrp" for 88).

Is there a standard protocol named "IGP" (Interior Gateway Protocol), or is that a name for a *type* of protocol, as per https://en.wikipedia.org/wiki/Interior_gateway_protocol?

guyharris:
> This is one of five (so far!) inconsistencies that would require changing the meaning of some keywords or removing/adding keywords to be solved properly.

What are the other four?

infrastation:
"igp" appears in Linux and FreeBSD `/etc/protocols`, which likely comes from [the registry](https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml). Let me post a write-up to the list.

guyharris:
> "igp" appears in Linux and FreeBSD `/etc/protocols`, which likely comes from [the registry](https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml).

So it's "pigp" without the "p".

infrastation:
Let me suggest documenting this better (the change log, the backward compatibility section of the man page) and waiting a bit until it is clear which other related changes go along.

infrastation:
Meanwhile commit a5874df documents the current non-portable behaviour of the `igrp` keyword.

infrastation:
Commit af0312e adds the FreeBSD et al. leg of filter tests for `igrp`.

PR #1430, opened 2025-01-07
Use correct size for VLAN tag load instruction

dmiller-nmap:
Not a true bug, since Linux kernel (net/core/filter.c) forces any load from SKF_AD_VLAN_TAG to a BPF_H width. However, this change makes libpcap code easier to understand.

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -9273,7 +9273,7 @@ gen_vlan_patch_vid_test(compiler_state_t *cstate, struct block *b_vid)
 	sjeq->s.jf = b_vid->stmts;
 	sappend(s, sjeq);
 
-	s2 = new_stmt(cstate, BPF_LD|BPF_B|BPF_ABS);
+	s2 = new_stmt(cstate, BPF_LD|BPF_H|BPF_ABS);
 	s2->s.k = SKF_AD_OFF + SKF_AD_VLAN_TAG;
 	sappend(s, s2);
 	sjeq->s.jt = s2;


infrastation:
One CI worker failed due to timeouts on a busy host, another failed due to a macOS issue that should disappear after rebasing on the current master branch.

dmiller-nmap:
The only failed check appears to be a timeout: `valid_wlan_subtype_qos_cf_poll_opt (alias 2): failed (filtertest timeout)`

Any comments on this change?

guyharris:
> Any comments on this change?

None on my part; as Denis noted, the timeout was just a timer that wasn't long enough for a busy (and, as I understand it, not all that fast) build host.

Denis, shall we merge this?

infrastation:
I had in mind to understand this a bit better because recently I came across two other cases of a load instruction size mismatching the packet field size (`pppoes` uses `ld` for a 16-bit session ID, `dpc` and `hdpc` use `ld` for a 14-bit point code). Let me try tomorrow.

Other than that, this will need a change log entry.

infrastation:
This won't reproduce in automatic build in future because linux-mips now does not use Valgrind anymore (it was too long). I tried reproducing the issue in a manual build, but it does not seem to reproduce anymore.

infrastation:
The code in `linux/net/core/filter.c:convert_skb_access()` in addition to making a `[vlan_tci]` load 16-bit (which the proposed change addresses) makes a `[vlanp]` load 32-bit, would it make sense to make that consistent as well?

infrastation:
A very similar inconsistency is `libpcap/gencode.c:gen_inbound()` using `BPF_H` for `[type]` for live captures on Linux, whilst `linux/net/core/filter.c:convert_skb_access()` enforces a `BPF_B` for `SKF_AD_PKTTYPE`. Although this does not affect the filtering (as originally discussed), it would be more useful to produce the same bytecode as what would actually filter packets in the kernel.

Also to preserve this coordination it would be useful to add a comment to every libpcap load instruction that is subject to the kernel-side size enforcement.

dmiller-nmap:
@infrastation I considered that a lesser problem, since the BPF code itself (and libpcap generally) does not treat the result as anything more than a 1-byte boolean value. The `vlan_tci` was a direct contradiction in that we "load" a byte and then access a halfword (16-bits).

infrastation:
One other thing that adds to this confusion is that some libpcap code patches the bytecode if it is going to be used on a live interface, and other libpcap code patches the bytecode if it is not going to be used on a savefile. These two conditions are not exactly the inverse of each other, so programs that use `pcap_open_dead()` see either the patched or the original bytecode:
```
$ ./testprogs/filtertest CAN_SOCKETCAN inbound
(000) ldh      [type]
(001) jeq      #0x4             jt 2	jf 3
(002) ret      #0
(003) ret      #262144

$ ./testprogs/filtertest EN10MB vlan 20
(000) ldh      [12]
(001) jeq      #0x8100          jt 4	jf 2
(002) jeq      #0x88a8          jt 4	jf 3
(003) jeq      #0x9100          jt 4	jf 8
(004) ldh      [14]
(005) and      #0xfff
(006) jeq      #0x14            jt 7	jf 8
(007) ret      #262144
(008) ret      #0
```

infrastation:
On a second thought, there is no need to mention this in the change log because the difference is mostly cosmetic. I am going to follow up on the other discussed imperfections separately later. Thank you for bringing this up in the first place.

PR #1443, opened 2025-01-17
TESTrun: Implement parallel testing.

infrastation:
Let's see how portable this updated Perl code is.


infrastation:
Apparently, Perl is non-threaded on OpenBSD. Let's see if there is a simple workaround...

infrastation:
Got working code, going to push a new revision soon.

infrastation:
This revision works on OpenBSD as well (without threads, as before).

infrastation:
Made a few more minor improvements and clean-ups, this is going to be merged right after the CI clears.

PR #1445, opened 2025-01-19
CMakeLists.txt: Add a "make check" target

fxlb:
Similar to tcpdump.

build.sh: Replace "testprogs/TESTrun" by "make -s check".

Makefile.in: Add "make -s check" commands in releasecheck target.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3585,3 +3585,19 @@ configure_file(
 
 add_custom_target(uninstall
     COMMAND ${CMAKE_COMMAND} -P ${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake)
+
+#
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -3585,3 +3585,19 @@ configure_file(
 
 add_custom_target(uninstall
     COMMAND ${CMAKE_COMMAND} -P ${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake)
+
+#
+# libpcap tests
+# We try to find the Perl interpreter and, if we do, we have the check
+# rule run tests/TESTrun with it, because just trying to run the TESTrun
```
infrastation, line 3592: Should be either "testprogs/TESTrun" or just "TESTrun".

PR #1448, opened 2025-01-23
Require a live capture for all Linux BPF extensions.

infrastation:
Please proof-read, this looks correct to me, but the changes concern all live captures on Linux, so it is desirable not to break things.

This is one of the related imperfections discussed in #1430.

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -8662,6 +8662,25 @@ gen_multicast(compiler_state_t *cstate, int proto)
 	/*NOTREACHED*/
 }
 
+#ifdef __linux__
+/*
+ * This is Linux; we require PF_PACKET support.  If this is a *live* capture,
+ * we can look at special meta-data in the filter expression; otherwise we
+ * can't because it is either a savefile (rfile != NULL) or a pcap_t created
+ * using pcap_open_dead() (rfile == NULL).  Thus check for a flag that
+ * pcap_activate() conditionally sets.
+ */
+static void
+require_basic_bpf_extensions(compiler_state_t *cstate, const char *keyword)
+{
+	if (cstate->bpf_pcap->bpf_codegen_flags & BPF_SPECIAL_BASIC_HANDLING)
+		return;
+	bpf_error(cstate, "%s not supported on %s (not a live capture)",
+	    keyword,
+	    pcap_datalink_val_to_description_or_dlt(cstate->linktype));
+}
+#endif // __linux__
+
 struct block *
 gen_ifindex(compiler_state_t *cstate, int ifindex)
 {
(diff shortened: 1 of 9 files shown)


infrastation:
I am going to update this commit with better tests later, but the essential change is as shown.

gencode.c
```diff
@@ -8662,6 +8662,25 @@ gen_multicast(compiler_state_t *cstate, int proto)
...
+ * can't because it is either a savefile (rfile != NULL) or a pcap_t created
+ * using pcap_open_dead() (rfile == NULL).  Thus check for a flag that
+ * pcap_activate() conditionally sets.
+ */
+static void
+require_basic_bpf_extensions(compiler_state_t *cstate, const char *keyword)
+{
+	if (cstate->bpf_pcap->bpf_codegen_flags & BPF_SPECIAL_BASIC_HANDLING)
+		return;
+	bpf_error(cstate, "%s not supported on %s (not a live capture)",
```
guyharris, line 8678: Yes, "not a live capture" is better than "when reading savefiles", as somebody could also be using `pcap_offline_filter()` on packets that have already passed the `pcap_setfilter()` filter, even if they're not reading a savefile. Thanks for catching and fixing that.

infrastation: That's a good point because I forgot about `pcap_offline_filter()`. If I understand it correctly, regardless of this change compiling bytecode using a `pcap_t` that has BPF extensions can produce bytecode with extensions, which will not be suitable for offline [userspace] filtering.

One potential way to address this problem automatically could be extending `struct bpf_program` to indicate whether the bytecode is suitable for kernel or userspace or both, and modifying functions that take compiled bytecode to reject it or to downgrade to userspace if necessary, but I am not ready to prepare such a change.

Meanwhile the most useful thing to do may be explaining this semantics better in pcap_offline_filter(3PCAP) and suggesting using a separate dead `pcap_t` to compile the bytecode that needs to be handled by `pcap_offline_filter()`.

Does it make sense?

guyharris: > One potential way to address this problem automatically could be extending `struct bpf_program`

That would probably introduce ABI problems.

There are some other API changes for filtering, e.g. with pcapng support, setting a filter on a handle from which you read a pcapng file means setting it to something in a form *independent* of the link-layer type, so that a pcapng file with multiple link-layer types can be handled. That might be a filter string to be compiled or some compiler intermediate representation.

That way, the filter would be compiled later if it's a filter string, or have code-generation done from the IR if it's an intermediate representation, and the calls that do that would supply information about the compiler target (that might also allow generating both cBPF and eBPF code with different backends).

> Meanwhile the most useful thing to do may be explaining this semantics better in pcap_offline_filter(3PCAP) and suggesting using a separate dead pcap_t to compile the bytecode that needs to be handled by `pcap_offline_filter()`.
> 
> Does it make sense?

Yes.

guyharris:
Looks good to me.

infrastation:
Thank you for the comments. This revision now includes a few minor cleanups and is going through a full CI run. Will be merged soon.

PR #1449, opened 2025-01-25
PPPoE, ARCnet, DECnet and MTP3 assorted improvements

infrastation:
This addresses some of the semi-related points raised in #1430.


infrastation:
A few similar fixups are in the works.

infrastation:
This will need an additional type cast to pass Appveyor. Also there is one other incomplete small change in my working copy, if it gets ready soon, it will go into the same pile.

@guyharris, the latest commit changes some of the memory management you implemented earlier, please confirm this is as safe as it looks to me. If it is indeed safe for ARCnet addresses, perhaps Ethernet addresses could be simplified the same way (giving a parsing function a pointer to a local variable instead of allocating the memory inside and managing the deallocation outside).

gencode.c
```diff
@@ -9071,13 +9070,16 @@ gen_acode(compiler_state_t *cstate, const char *s, struct qual q)
...
+			bpf_u_int32 addr;
+			/*
+			 * The lexer currently defines the address format in a
+			 * way that makes this error condition never true.
+			 * Let's check it anyway in case this part of the lexer
+			 * changes in future.
+			 */
+			if (! pcapint_atoan(s, &addr))
+			    bpf_error(cstate, "invalid ARCnet address '%s'", s);
+			return gen_ahostop(cstate, addr, (int)q.dir);
```
guyharris, line 9082: This gets

```
C:\projects\libpcap\gencode.c(9082): warning C4242: 'function': conversion from 'bpf_u_int32' to 'const uint8_t', possible loss of data [C:\projects\libpcap\build\pcap.vcxproj]
```

from Visual Studio, which is the *only* C compiler I've ever seen warn about converting a wider integer value to a narrower integer value (except for Apple's GCC/Clang, which do so only for 64-bit -> 32-bit, as part of their 64-bit-ification of macOS).

You probably need to do something such as

```
			if (! pcapint_atoan(s, &addr) || addr > UINT8_MAX)
			    bpf_error(cstate, "invalid ARCnet address '%s'", s);
```

to make sure that it's a value that fits in a 1-byte address field, and then cast `addr` to `uint8_t` to tell the compiler "yes, I've checked, this is OK".

infrastation: Or perhaps `pcapint_atoan()` could take a pointer to an `uint8_t` instead of trying to look consistent with the other functions...

infrastation:
Added a couple more clean-ups in my working copy, now trying to complete the other related improvements, will update when everything seems ready.

infrastation:
This now looks about the right number of inter-related commits and is going be merged soon. I understand the memory management change is safe in all contexts. Four commits make changes to the code generator. The filter tests make it easy to see (as far as the test coverage extends) that any changes to the filter programs went exactly as intended, everything else did not change. You are welcome to proof-read.

infrastation:
That's a good question. For the queued changes I have just run a few manual tests for ARCnet, DECnet and MTP3 filtering, everything looks as expected. DECnet has an imperfection that is not specific to this change: `./tcpdump -nve -r tests/DECnet_Phone.pcap 'decnet src not 1.1'` should reject all packets, but it fails to reject Ethernet Endnode Hello from node 1.1. This is exactly the same as in libpcap 1.10.3, for example, because `gen_dnhostop()` implements only two DECnet packet types out of several.

Generally speaking, a correctly looking filter program is necessary but not sufficient for correct filtering of packets, so such a class of tests ("this filter expression applied to this sequence of packets produces exactly this sequence of filtering results") is indeed meaningful and it would be useful to do more of them.  But how exactly?

On the one hand, tcpdump tests could use filter expressions more often, but on the other hand, tcpdump source tree is not the right place for tests that cover libpcap code. The `pppoes` and `geneve` filters got into tcpdump tests because there was nowhere in libpcap to put them at the time. But the only reliable way to do libpcap tests is keeping them in libpcap tree such that the tests correlate with the code, and it seems a good idea not to use tcpdump as a test tool for libpcap.

One potential solution could be using a new test program or extending an existing one, for example, `./testprogs/filtertest -r packets.pcap -F expression.txt` could print each packet's filter result (accepted/rejected, or even better the filter program return value), optionally with the packet number and the timestamp to assist debugging.  In this case instead of using `pcap_setfilter()` the extended `filtertest` would need to make a call to `pcap_offline_filter()` for every packet.

Does that make sense?

infrastation:
Let's see if this revision passes Appveyor. I do not see why the default case in a switch block would not be good enough, but the compiler demanded an explicit value:
```
warning C4061: enumerator 'HEX2' in switch of enum '<unnamed-enum-START>' is not explicitly handled by a case label
```

infrastation:
Let me see if the `filtertest` extension is as simple as it seems.

infrastation:
The C leg works:
```
./testprogs/filtertest -r ../tcpdump/tests/loopback.pcap 'greater 70 and ether src host not aa:00:04:00:69:04'
0
0
1500
0
1500
0
```
Prototyping the Perl leg...

infrastation:
The Perl leg works. Let me clean this up and add the new tests before these changes to demonstrate the actual filtering continues to work at least for the new test cases.

infrastation:
This revision automatically runs the type of tests I did manually earlier. Passes a complete build matrix with Valgrind on Linux.

infrastation:
This revision includes a workaround for the missing `<sysexits.h>` header on Windows.

infrastation:
This revision addresses a warning on Windows.

Issue #1367, opened 2024-10-10
unsupported DSA tag: qca


ErwanMAS:
Hi ,
What we need todo to solve this error ? 

```
root@OpenWrt:~# tcpdump -nei eth1
tcpdump: unsupported DSA tag: qca

```

```
root@OpenWrt:~# tcpdump --version
tcpdump version 4.99.5
libpcap version 1.10.5 (with TPACKET_V3)
32-bit build, 64-bit time_t
```

ErwanMAS:
> Which specific hardware is this, and which OpenWrt release?

so i have 2 dsa tag that are no supported qca / mtk 


Linksys EA7500 V1

```
cat /etc/openwrt_release ; cat /etc/board.json | jq .model ; tcpdump -nei eth0
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='24.10.0-rc2'
DISTRIB_REVISION='r28161-ea17e958b9'
DISTRIB_TARGET='ipq806x/generic'
DISTRIB_ARCH='arm_cortex-a15_neon-vfpv4'
DISTRIB_DESCRIPTION='OpenWrt 24.10.0-rc2 r28161-ea17e958b9'
DISTRIB_TAINTS=''
{
  "id": "linksys,ea7500-v1",
  "name": "Linksys EA7500 V1 WiFi Router"
}
tcpdump: unsupported DSA tag: qca

```

Linksys EA7500 V2

```

cat /etc/openwrt_release ; cat /etc/board.json | jq .model ; tcpdump -nei eth0
DISTRIB_ID='OpenWrt'
DISTRIB_RELEASE='24.10.0-rc2'
DISTRIB_REVISION='r28161-ea17e958b9'
DISTRIB_TARGET='ramips/mt7621'
DISTRIB_ARCH='mipsel_24kc'
DISTRIB_DESCRIPTION='OpenWrt 24.10.0-rc2 r28161-ea17e958b9'
DISTRIB_TAINTS=''
{
  "id": "linksys,ea7500-v2",
  "name": "Linksys EA7500 v2"
}
tcpdump: unsupported DSA tag: mtk
root@ramir:~#

```

0xKira:
Same problem here with OpenWrt 24.10.0-rc3
```
tcpdump: unsupported DSA tag: mtk
```

piyushgarg:
```
arp-scan --localnet
pcap_activate: unsupported DSA tag: mtk
```

[...]

PR #1451, opened 2025-01-27
Don't fail on unsupported DSA tag

vladimiroltean:
Up until commit 8f1f6fcbe4b4 ("If we can't allocate a DLT_ list, fail."), the iface_dsa_get_proto_info() return code was ignored, but now it isn't.

Many DSA tags are in use which libpcap has no idea about: https://github.com/the-tcpdump-group/libpcap/issues/1367

Let's keep the behavior as before, i.e. don't give up on these packets, even though we don't know what's inside.

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -5482,10 +5482,14 @@ iface_dsa_get_proto_info(const char *device, pcap_t *handle)
 		}
 	}
 
+	/*
+	 * Not an error. We don't know how to decode all DSA tag formats,
+	 * but we shouldn't give up on the packet.
+	 */
 	snprintf(handle->errbuf, PCAP_ERRBUF_SIZE,
-		      "unsupported DSA tag: %s", buf);
+		 "unsupported DSA tag: %s", buf);
 
-	return PCAP_ERROR;
+	return 0;
 }
 
 /*


minimaxwell:
Hello Vlad,

I'm all in for that one, I gave it a test and it works fine (debugging a ksz9477 setup whose tag format isn't supported as well). I strongly agree that this regression needs fixing.

Thanks,

Maxime

guyharris:
So what will `handle->linktype` be set to in this case? `DLT_EN10MB`?

vladimiroltean:
Yeah.

vladimiroltean:
Yeah. I mean, they look horrible in tcpdump, but it gets the job done.
```
root@debian:~# tcpdump -i end0 -e -n -XX
listening on end0, link-type EN10MB (Ethernet), snapshot length 262144 bytes
00:20:19.523547 d8:58:d7:00:ca:6d > 01:80:c2:02:00:00, ethertype Unknown (0xdadb), length 64: 
        0x0000:  0180 c202 0000 d858 d700 ca6d dadb 0c02  .......X...m....
        0x0010:  0027 4242 0300 0002 023e 7fff c87f 54df  .'BB.....>....T.
        0x0020:  a740 0000 4e20 8000 9632 d9a4 8e67 8002  .@..N....2...g..
        0x0030:  0100 0a00 0200 0800 0000 0000 0000 0000  ................
```

infrastation:
So these packets are not actually Ethernet, correct?

vladimiroltean:
They are absolutely Ethernet frames with an extra header (generically called "tag" because depending on switch vendor, it may not actually be "header" but also "trailer") that contains switch-specific information. For example, in the dump above, if you remove "dadb 0c02", what you get is a normal STP packet.
Actually DSA is the layer in the kernel which decodes this switch tag and creates virtual interfaces (stacked on top of the physical one, similar to VLAN) for each switch port (a field in this tag). If you run tcpdump on the virtual switch interfaces, it's normal Ethernet with all that entails - they're Ethernet switches. If you run tcpdump on the physical host interface you get this gibberish.
See page 5 of https://netdevconf.info/2.1/papers/distributed-switch-architecture.pdf.

infrastation:
Thank you for the comments. Do you mean in this specific case it is exactly one of those or something else?
* [LINKTYPE_DSA_TAG_EDSA](https://www.tcpdump.org/linktypes/LINKTYPE_DSA_TAG_EDSA.html)
* [LINKTYPE_DSA_TAG_DSA](https://www.tcpdump.org/linktypes/LINKTYPE_DSA_TAG_DSA.html)
* [LINKTYPE_DSA_TAG_BRCM](https://www.tcpdump.org/linktypes/LINKTYPE_DSA_TAG_BRCM.html)
* [LINKTYPE_DSA_TAG_BRCM_PREPEND](https://www.tcpdump.org/linktypes/LINKTYPE_DSA_TAG_BRCM_PREPEND.html)

vladimiroltean:
No, it's not one of the 4 protocols for which libpcap assigns a specific link type, but rather one of the remaining 25 for which it doesn't.
https://elixir.bootlin.com/linux/v6.13.2/source/include/net/dsa.h#L58
Mostly nobody bothers to write decoders for these, and would rather spend a few more minutes to look at a raw hex dump of the packet with no pretty printing, in the off occasion that debugging at this level is actually necessary at all.
The blamed patch actually made debugging impossible for the vast majority of DSA protocols.

guyharris:
> No, it's not one of the 4 protocols for which libpcap assigns a specific link type, but rather one of the remaining 25 for which it doesn't.

Then perhaps we should assign more of them, and do tcpdump and Wireshark dissectors for them.

> Mostly nobody bothers to write decoders for these, and would rather spend a few more minutes to look at a raw hex dump of the packet with no pretty printing, in the off occasion that debugging at this level is actually necessary at all.

Then it sounds as if there should be a `LINKTYPE_DSA_UNSUPPORTED` type, which is dissected as a raw hex dump by tcpdump/Wireshark/etc., rather than using `LINKTYPE_ETHERNET` and dissecting those as Ethernet.

 The blamed patch actually made debugging impossible for the vast majority of DSA protocols.

vladimiroltean:
I mean, they _are_ Ethernet packets, and comply to all Ethernet packet rules. The host port (which is most of the time a regular, DSA-unaware network card, and which any other day would actually use DLT_EN10M) parses as much of the packet as it can, and it treats it for what it is - an Ethernet frame with an unknown Ethertype. In the example above, "d8:58:d7:00:ca:6d > 01:80:c2:02:00:00, ethertype Unknown (0xdadb), length 64" is exactly as much as the protocol parser of the host port makes out of this packet. I don't know to what end we would lose even this small amount of information by dissecting as a raw hex dump in favor of just falling back to DLT_EN10M.

guyharris:
> I mean, they _are_ Ethernet packets, and comply to all Ethernet packet rules.

By which you presumably mean "they are blobs of bytes that, somewhere within them, contain an Ethernet packet, perhaps with additional stuff stuck in front of it, after it, or in the middle of it.

See, for example, [the `LINKTYPE_DSA_TAG_BRCM_PREPEND` type](https://www.tcpdump.org/linktypes/LINKTYPE_DSA_TAG_BRCM_PREPEND.html), in which the tag appears *before* the destination address; the DSA tag doesn't look as if it's inserted the way a VLAN tag is.

So, no, an Ethernet dissector will not necessarily...

> The host port (which is most of the time a regular, DSA-unaware network card, and which any other day would actually use DLT_EN10M) parses as much of the packet as it can, and it treats it for what it is - an Ethernet frame with an unknown Ethertype.

...parse the packet as being an Ethernet frame with an unknown Ethertype.  If `LINKTYPE_DSA_TAG_BRCM_PREPEND` is the ***ONLY*** tag type that's a prefix rather than either a trailer or something inserted in the fashion of a VLAN tag - i.e., its first two octets are an Ethertype that's been assigned so that it won't be used by anything *other* than the DSA tag in question - then, yes, it can be made to work with `LINKTYPE_ETHERNET`.

And if, in fact, that's the case, then we don't need `LINKTYPE_DSA_TAG_xxx` types for anything other than prepend or trailer types, as tcpdump, Wireshark, etc. can just use the Ethertype value to determine the DSA tag type.

However, if a DSA tag type is ever used for on-the-wire Ethernet traffic, a `LINKTYPE_` value is needed. I think there *might* be places where it is, but maybe I'm thinking of something else.



 In the example above, "d8:58:d7:00:ca:6d > 01:80:c2:02:00:00, ethertype Unknown (0xdadb), length 64" is exactly as much as the protocol parser of the host port makes out of this packet. I don't know to what end we would lose even this small amount of information by dissecting as a raw hex dump in favor of just falling back to DLT_EN10M.

guyharris:
> Arguably, if the frame was Ethernet, the hardware and the kernel would both handle it as Ethernet. If they both handle it as DSA, then most likely it is DSA. It would be most helpful to focus on the specification.

Which specification?  DSA doesn't specify what tags look like - that's up to the vendor, and many of them seem to do their own thing.

vladimiroltean:
So when I said that the host port is "most of the time a regular, DSA-unaware network card", I really selected my words carefully by saying "most of the time", and I wasn't trying to be dishonest about omitting the rest, it's just that I consider the exceptions to be irrelevant to this discussion. However you decided to pick out LINKTYPE_DSA_TAG_BRCM_PREPEND, which is one of the exceptions.

> By which you presumably mean "they are blobs of bytes that, somewhere within them, contain an Ethernet packet, perhaps with additional stuff stuck in front of it, after it, or in the middle of it.

Very broadly speaking, DSA switches encapsulate Ethernet frames when delivering them towards the host port, yes, but I don't want you to draw the wrong conclusion from that.
See, many DSA switches are discrete chips, and the switch vendor doesn't control which network card it will be integrated with, on the final PCB. Making sure that packets pass through the host interface's RX filters is a serious concern, so the encapsulation has to be backwards compatible with Ethernet to an acceptable degree. For these reasons, category 2 and 3 tagging protocols, [as described here](https://www.kernel.org/doc/html/latest/networking/dsa/dsa.html#switch-tagging-protocols), are popular with discrete chips - although they don't come without their own challenges. Just one example: since individual DSA switch port may have their own MAC address, potentially not equal to the host port's MAC address, the kernel framework has to explicitly manage the RX filtering lists of the host port to not drop these foreign MAC DA values.
Category 1 tagging protocols have a different philosophy. Some designs guarantee they'll bypass the host port's RX filter by using a long prefix as a substitute for the Ethernet header, so that the MAC DA appears as broadcast to the host port, [like Ocelot](https://elixir.bootlin.com/linux/v6.13.2/source/include/linux/dsa/ocelot.h#L45). Note that current the Ocelot tagging protocol in the kernel does not use the long prefix though, but the short one, [for reasons that were explained in this commit](https://github.com/torvalds/linux/commit/5124197ce58b5706bb60c2ecb3b79f4dfabab6e1). Anyway, compatibility with the Ethernet header is always a concern.

But you've jumped straight to LINKTYPE_DSA_TAG_BRCM_PREPEND, and I believe that only now do we have all the context on the table to explain it.

Sometimes, DSA switches are integrated on the same silicon die with an entire SoC, and there you still have an internal Ethernet host port for management traffic. Sometimes, the vendor of the host port IP is different than the vendor of the integrated switch IP, like NXP LS1028A, with an enetc host and an ocelot switch. So, that changes nothing in terms of having to be Ethernet-compatible, because that's still the common denominator even though they're both on the same SoC. But sometimes, an SoC vendor like Broadcom has the chance to optimize the integration, and break the compatibility with Ethernet for that internal MAC-to-MAC connection. Because all traffic that the host port is going to see is coming from the switch, these guys can modify the hardware protocol parser of the host port to essentially expect the DSA encapsulation. You can ask yourself why, if they bothered to optimize the design, they didn't go all the way and provide a more performant packet I/O mechanism, like have buffer descriptor rings and DMA for each switch port rather than the bottleneck of a single internal Ethernet link, and the answer has to do with the fact that it's just an evolution of the baseline design I already presented earlier. Anyway, you could make a valid point that DSA tagging protocols that aren't backwards compatible with Ethernet should have a dedicated link type, but that point is moot with LINKTYPE_DSA_TAG_BRCM_PREPEND which already does, but honestly, backwards compatiblity with Ethernet is not something that we as DSA maintainers monitor when we review new tagging protocol submissions, so I can't tell you for sure how many other such protocols are there in the wild. I'll try to think of some way to integrate in the medium term testing of new tagging protocols with the dsa_loop kernel driver (which is a way of making any Ethernet interface think it's attached to a DSA switch even if it physically isn't), and inform patch submitters that it's mandatory for them to submit a dedicated link type to libpcap if the protocol is not backwards-compatible with Ethernet. That's about the most I can promise, but I really want the assumed default to remain that DSA tagging protocols are compatible with Ethernet to some extent. Who wants to add a specific link type for a certain Ethernet-compatible protocol, perfectly fine, but don't force that.

vladimiroltean:
> And if, in fact, that's the case, then we don't need LINKTYPE_DSA_TAG_xxx types for anything other than prepend or trailer types, as tcpdump, Wireshark, etc. can just use the Ethertype value to determine the DSA tag type.

There are cases which are not as simple as that.
Case 1 - Original Marvell DSA (not EtherType DSA) which is a category 2 protocol but lacks an EtherType. You need to know you're looking at Marvell DSA before understanding what to make of it. Yes, original Marvell DSA is an imperfect design due to the fact that variable fields map over what the host port perceives as EtherType in ways that the host port doesn't like, [and there have been instances of that](https://lore.kernel.org/netdev/20210323102326.3677940-1-tobias@waldekranz.com/). Yet it exists, and there's nothing we can do about it.
Case 2 - NXP SJA1105 and SJA1110, whose documented tagging protocol is essentially a hybrid of a hardware tagging mechanism and a software-defined tagging mechanism. For link-local L2 multicast, the SJA1105 will patch up bytes of the original MAC DA with the source port and switch ID, in no way increasing the packet size, with the expectation that software recovers these values and patches up those MAC DA bytes back to zeroes. For any other traffic type, the hardware has no assist for source port identification, so the software configures it to send packets as VLAN-tagged towards the host port, with a custom VLAN TPID of 0xdadb, and with the VLAN ID identifying the source port of the packet. That method does increase the packet size. But when looking at the tagging protocol as a whole, not all traffic is identifiable by EtherType, and not even all tagged packets have the same tag length.

Look, I really appreciate that you two are trying to understand and leave no stone unturned, but you have to understand that this regression has a bit of an urgency to it. Even though the change was made a long time ago, it only made it to distributions recently, and it has started affecting people.

vladimiroltean:
> However, if a DSA tag type is ever used for on-the-wire Ethernet traffic, a LINKTYPE_ value is needed. I think there might be places where it is, but maybe I'm thinking of something else.

You need to define "wire" here. DSA-tagged traffic is passed between the switch and its host port. Since the host port is a dedicated Ethernet interface, that's pretty much "on the wire" as far as it's concerned, but outside of this system, DSA tags are not visible to the outside world. We would really hide the host port from being visible in ifconfig if we could, because as a user you're not supposed to interact with it, but directly with the switch user ports, but we can't, and it's occasionally useful to debug by running tcpdump on it. The host port is just a resource used by DSA, a dumb pipe, and the "wire" varies from Verilog wires (for integrated switches) to PCB traces to regular Ethernet cables (think Beaglebone Black connected to a switch evaluation board).

Though, mainly during developer testing, we can activate the dsa_loop kernel driver, and this lets any Ethernet interface speak a certain DSA tagging protocol with the outside world, with no switch to parse the tags. The public infrastructure for this isn't developed almost at all - anybody who does this has their own tooling for their own purpose. But we treat a DSA tagging protocol, as identified by its textual name, as stable ABI (at least after a certain point, there have been incompatible changes before, which we're trying not to repeat), and anybody who wants to transition a protocol from the Ethernet link type to a specific one can do so at any time in the future.

guyharris:
Yes, I'd already read the part of the DSA documentation about the three tag types.

For type 1 tags, such as `LINKTYPE_DSA_TAG_BRCM_PREPEND`, programs reading those packets will not dissect *anything* correctly unless it's assigned a `LINKTYPE_`/`DLT_` value separate from `LINKTYPE_ETHERNET`/`DLT_EN10MB`, as the frame doesn't begin with a 6-octet destination address, followed by a 6-octet source address, followed by a type/length field, it begins with a tag.

For type 2, *if* the tag begins with a two-octet big-endian field that contains an assigned Ethertype value that is used *only* for a particular tag type, *then* `LINKTYPE_ETHERNET`/`DLT_EN10MB` can handle that type - if tcpdump or Wireshark or... handle that Ethertype value, they can dissect the frame correctly, showing the MAC addresses, the tag, and the following Etherype/length and payload. Otherwise, it would need a separate `LINKTYPE_`/`DLT_` in order to dissect the tag and payload.

 For type 3, the frame can be dissected as `LINKTYPE_ETHERNET`/`DLT_EN10MB`, but will have extra stuff at the end.  Assigning a `LINKTYPE_`/`DLT_` value separate from `LINKTYPE_ETHERNET`/`DLT_EN10MB` would allow the tag to be dissected, rather than to be treated as part of the packet payload, which *might* cause it to be misdirected if the payload protocol doesn't have its own length field.

So if somebody can provide a list of the tag types that are:

- type 1;
- type 2 with a tag that does *not* begin with an assigned Ethertype value;
- type 2 with a tag that begins with an assigned Ethertype value, but where that Ethertype value is not sufficient to determine the packet format;
- type 3;

then we can assign `LINKTYPE_`/`DLT_` values to those, and use `LINKTYPE_ETHERNET`/`DLT_EN10MB` for *all* the other types. That would mean having the `dsa_protos[]` table contain only the types from the list, and the code would use`LINKTYPE_ETHERNET`/`DLT_EN10MB` for any tag type not fund in that table.

So:

- "brcm" is type 2 with a tag that does *not* begin with an assigned Ethertype value;
- "brcm-prepend" is type 1;
- "dsa" is type 2 with a tag that does *not* begin with an assigned Ethertype value;
- "edsa" is type 2 with a tag that begins with a 2-octet type, but I think that type is programmable, so there's no guarantee that it's not going to be set to a value that means something else for regular Ethernet traffic;
- "rtl4a" is type 2 with a tag that begins with an assigned Ethertype value, 0x8899, that is also used for Ethernet packets, but where the packet includes a packet type field that allows a dissector to distinguish between various packet formats with that type value;
- "rtl8_4" is type 2 with a tag that begins with the same Ethertype value, 0x8899;
- "rtl8_4t" is type 3.

Currently:

- we have `LINKTYPE_`/`DLT_` values  for "brcm", "brcm-prepend", "dsa", and "edsa", which is as it should be, as they're either type 1 or type 2 without a usable Ethertype value;
- we use `LINKTYPE_ETHERNET`/`DLT_EN10MB` for "rtl4a" and "rtl8_4", which is as it probably should be, as they have an assigned Ethertype and the data following the Ethertype has fields that allow distinguishing between different formats (tags and on-the-Ethernet Realtek protocols);
- we use `LINKTYPE_ETHERNET`/`DLT_EN10MB` for "rtl8_4t", which is not as it should be, as it's a trailer tag, but we'd need to write dissectors for them if we assigned it its own `LINKTYPE_`/`DLT_` value, and trailer-protocol dissectors have to be written carefully (so as to properly handle packets sliced with a snapshot length), so we'll leave that it is for now.

I'd be inclined, *for now*, to fall back on `LINKTYPE_ETHERNET`/`DLT_EN10MB` for unknown types, *but assign `LINKTYPE_`/`DLT_` values for type 1/type 2 without a usable Ethertype/type 3 tags in the future*. It would be nice if there were a way to somehow indicate to the program reading packets that the purported `LINKTYPE_ETHERNET`/`DLT_EN10MB` are assigned that `LINKTYPE_`/`DLT_` value as a fallback and that it might not be correct, but the libpcap API and the pcap file format has no way to do that.

It would be *very* helpful if somebody could provide a list of type 1/type 2 without a usable Ethertype/type 3 tags, so we could, if nothing else, put entries in the `dsa_protos[]` table that map them to `DLT_EN10MB`, with a comment noting that they really need their own link-layer type.

vladimiroltean:
> Currently:
>     we have LINKTYPE_/DLT_ values for "brcm", "brcm-prepend", "dsa", and "edsa", which is as it should be, as they're either type 1 or type 2 without a usable Ethertype value;
>     we use LINKTYPE_ETHERNET/DLT_EN10MB for "rtl4a" and "rtl8_4", which is as it probably should be, as they have an assigned Ethertype and the data following the Ethertype has fields that allow distinguishing between different formats (tags and on-the-Ethernet Realtek protocols);

As per your own criteria, you have edsa in the wrong category (it has its own link type but by your logic it shouldn't).
Which is something I don't fully understand. Why wouldn't you want to have unified handling for all kinds of DSA tags, but would rather single out the Ethertype ones?

guyharris:
> > Currently:
> > we have LINKTYPE_/DLT_ values for "brcm", "brcm-prepend", "dsa", and "edsa", which is as it should be, as they're either type 1 or type 2 without a usable Ethertype value;
> > we use LINKTYPE_ETHERNET/DLT_EN10MB for "rtl4a" and "rtl8_4", which is as it probably should be, as they have an assigned Ethertype and the data following the Ethertype has fields that allow distinguishing between different formats (tags and on-the-Ethernet Realtek protocols);
> 
> As per your own criteria, you have edsa in the wrong category (it has its own link type but by your logic it shouldn't).

Sorry, I must not have stated my logic clearly enough.

As I understand it, the tag used by "edsa" begins with a 2-octet value that comes from a driver-settable register in the hardware, so there's no guarantee that somebody won't set it to 0x0800 or whatever and, even if it's set to the one value that the IEEE site says is assigned to Marvell, 0x22E3, we don't know for certain whether that might also be used for on-the-Ethernet protocols that cannot be reliably distinguished from EDSA tags.

Therefore, it belongs in the "type 2 without a usable Ethertype value" - or, perhaps, "known-to-be-usable Ethertype value". 

> Why wouldn't you want to have unified handling for all kinds of DSA tags, but would rather single out the Ethertype ones?

Because:

- nothing in packets with type 1 tags can be properly dissected by an Ethernet dissector, so `LINKTYPE_ETHERNET`/`DLT_EN10MB` shouldn't be used for them;
- the only parts of packets with type 2 tags that do not begin with an assigned Ethertype value will be the destination and source MAC addresses;
- packets with type 2 tags that begin with an assigned Ethertype value, but where that value can also be used for other types of packets, would not be correctly dissectible without other out-of-band information, and a separate `LINKTYPE_`/`DLT_` value is a good type of out-of-band information, as it doesn't require the user to manually provide it;
- packets with type 3 tags can largely be dissected correctly, except perhaps if the tag is mistaken for the end of the Ethernet payload, and fixing that also requires out-of-band information.

vladimiroltean:
Ok, but confusingly you refer to type 2 tags with no usable EtherType as if there was any tag at all whose EtherType is in the IANA 802 numbers table. I don't think that's the case - none of them are.

guyharris:
> Ok, but confusingly you refer to type 2 tags with no usable EtherType as if there was any tag at all whose EtherType is in the IANA 802 numbers table. I don't think that's the case - none of them are.

I know it's the case, as the Realtek tags are.  They use Ethertype 0x8899, which is registered to "Realtek Semiconductor Corp.". Go to https://regauth.standards.ieee.org/standards-ra-web/pub/view.html#registries, select "EtherType(TM)" as the "Product", click the search button, enter 8899 in the filter box, and click "Filter".

vladimiroltean:
Here's what I've been able to extract from the kernel sources.

| proto         | type             | length (bytes)     | xmit format            | rcv format               |
| ------------  | ---------------- | ------------------ | ---------------------- | ------------------------ |
| ar9331        | 1                | 2                  | 15:14 - VERSION<br>13:12 - PRIORITY<br>10:8 - TYPE<br>7 - BROADCAST<br>6 - FROM_CPU<br>5:4 - RESERVED<br>3:0 - PORT_NUM        | like xmit |
| brcm          |  2 w/o EtherType |   4                | already documented |   already documented |
| brcm-legacy   |  2               |   6                | EtherType 0x8874<br>30:29 - TYPE<br>3:0 - PORT_ID |    like rcv |
| brcm-prepend  |  1               |   4                | like brcm            | like brcm |
| dsa           |  2 w/o EtherType |   4                | already documented   | already documented |
| edsa          |  2               |   8                | already documented   | already documented |
| gswip         |  1               |   4 TX, 8 RX       | 26:24 - SLPID<br>18:16 - DPID<br>15 - PORT_MAP_EN<br>14 - PORT_MAP_SEL<br>13 - LRN_DIS<br>12 - CLASS_EN<br>11:8 - CLASS<br>6:1 - PORT_MAP<br>0 - DPID_EN |        6:4 - SPPID |
| hellcreek     |  3               |   1                | 1:0 - PORT |           like xmit |
| ksz8795       |  3               |   1                | 7 - LOOKUP<br>6 - OVERRIDE<br>1:0 - PORT |           1:0 - PORT |
| ksz9477       |  3               |   2 RX<br>6 RX with tstamp<br>1 TX<br>5 TX with tstamp   |  |  |
| ksz9893       |  3               |   2 RX<br>6 RX with tstamp<br>1 TX<br>5 TX with tstamp             |  |  |
| lan937x       |  3               |   2 TX<br>6 TX with tstamp<br>2 RX<br>6 RX with tstamp             |  |  |
| lan9303       |  2               |   4                | EtherType 0x8100 |  |
| mtk           |  2 w/o EtherType |   4                |  |  |
| none          |  no tag          |   0                |  |  |
| ocelot        |  1               |   20               |  |  |
| seville       |  1               |   20               |  |  |
| ocelot-8021q  |  2               |   4                | EtherType 0x8100  |  |
| qca           |  2 w/o EtherType |   2                |  |  |
| rtl4a         |  2               |   4                | EtherType 0x8899  |  |
| rtl8_4        |  2               |   8                | EtherType 0x8899  |  |
| rtl8_4t       |  3               |   8                | tail tag variant of rtl8_4  |  |
| a5psw         |  2               |   8                | EtherType 0xe001  |  |
| sja1105       |  2               |   4                | EtherType 0xdadb or 0x8100  |  |
| sja1110       |  hybrid          |   4 TX<br>12 TX link local<br>(8 header, 4 trailer)<br>4 RX<br>21-36 RX link local<br>(8 header, 13 trailer, 0-15 padding)           | EtherType 0xdadc  |  |
| trailer       |  3               |   4                |  |  |
| vsc73xx-8021q |  2               |   4                | EtherType 0x8100  |  |
| xrs700x       |  3               |   1                |  |  |

guyharris:
Is "sja1105" just a standard VLAN tag if it has an Ethertype of 0x8100? 0xdadb doesn't appear to be in the IEEE database.

Are "lan9303", "ocelot-8021q", and "vsc73xx-8021q" all standard VLAN tags? (The "8021q" in the names of the latter two suggest that those two are.)

vladimiroltean:
They are all VLAN tags with specific meanings superimposed on top of the VID field. In the case of "sja1105", the TPID can either be 0x8100 or 0xdadb, depending on whether the switch is VLAN-aware or not - the TPID doesn't affect the interpretation of the other fields in the tag. The value of 0xdadb is self-assigned, no guarantee there won't appear a protocol using this EtherType in the future.
The field meanings are the same for "sja1105", "ocelot-8021q" and "vsc73xx-8021q" and [documented here](https://elixir.bootlin.com/linux/v6.13.2/source/net/dsa/tag_8021q.c#L15). For "lan9303", the meanings are different and [documented here](https://elixir.bootlin.com/linux/v6.13.2/source/net/dsa/tag_lan9303.c#L24).

guyharris:
Of the non-8100 Ethertypes:

- 8874 is assigned to Broadcom, so the only question is whether it's also used for any on-the-Ethernet protocols - Wireshark already dissects it as a "Broadcom MAC Management" tag;
- 8899 is assigned to Realtek, and it's used for both tags and on-the-Ethernet protocols, but there's a subprotocol field that's used to distinguish between the various protocols for which it's used and the various tag types for which it's used (as dissected by Wireshark);
- e001, dadb, and dadc are not in the IEEE database.
-

guyharris:
> The value of 0xdadb is self-assigned, no guarantee there won't appear a protocol using this EtherType in the future.

So it's probably best to assign a `LINKTYPE_`/`DLT_` for "sja1105" (that would also allow the VID field to be interpreted specially).

vladimiroltean:
What does this all mean for this patch?

guyharris:
> What does this all mean for this patch?

My inclination right now is to expand the table of known tag types to include all he ones currently used in the Linux kernel, *temporarily* mapping all those that don't have `LINKTYPE_`/`DLT_` values assigned to `LINKTYPE_ETHERNET`/`DLT_EN10MB`, add comments to *all* of them to indicate whether they're type 1, type 3, type 2 with a usable Ethertype, type 2 with at least one Ethertype being unreliable (self-assigned rather than IEEE-assigned), or type 2 without an Ethertype, and continue to return an error for unknown tag types.

If somebody adds a tag type, they should tell us about it, so we can add it to the list.

After that, we assign `LINKTYPE_`/`DLT_` values to those that need them, update the table, and at least make tcpdump dissect them (and maybe add Wireshark after that).

infrastation:
Using `DLT_EN10MB` for unsupported DSA encodings introduces easy ways for things to go wrong, e.g. allows to run higher-level filters on wire encodings that do not match Ethernet in the first place. Guy earlier suggested using `LINKTYPE_DSA_UNSUPPORTED`, this seems a step in the correct direction to me.

If the DSA protocol/type is known at the capture time and there is no specific DLT for it, it would seem reasonable to state: "this is a DSA frame, it was captured on Linux and Linux indicated the variety of DSA was X". In other words, it could be something such as `DLT_LINUX_DSA` with at least one field that specifies the DSA type (possibly with a frame direction indicator). For those who cannot wait to see the data this DLT would be trivially printed as a hex dump. For those who want to develop code for filtering and dissecting it properly this DLT would provide the required information to do it (`DLT_EN10MB` does not even imply that the frame is DSA, let alone specific DSA type).

Does it make sense?

infrastation:
Alternatively, there could be something like `DLT_RAW_BYTES` with no claims about the origin or the structure of the packets, and naturally fit for hex dumps only.

infrastation:
@guyharris, do you mind if I prototype the changes to avoid mixing DSA into `DLT_EN10MB`?

guyharris:
> @guyharris, do you mind if I prototype the changes to avoid mixing DSA into `DLT_EN10MB`?

Nope, go ahead.

thomas725:
I have 3 routers which I upgraded to OpenWrt 24.10.0 (r28427-6df0e3d02a) today.
On two of them tcpdump now fails to start with this error:
> tcpdump: unsupported DSA tag: mtk
* Xiaomi Mi Router AX3000T
* Xiaomi Mi Router 4A Gigabit Edition

Is there a reasonable simple way to install your fix on my devices without being either an OpenWRT nor libpcap developer, or do I need to downgrade OpenWRT to a previous version (how old does it need to be?) that doesn't have this bug?

infrastation:
According to OpenWrt device pages ([AX3000T](https://openwrt.org/inbox/toh/xiaomi/ax3000t), [4A Gb](https://openwrt.org/inbox/toh/xiaomi/xiaomi_mi_router_4a_gigabit_edition)), these devices are supported. Perhaps the documentation is not 100% accurate in every regard, but this is not uncommon. A paying customer of the vendor that makes these Xiaomi devices could make a point that OpenWrt developers didn't do a thorough QA for this combination of hardware and firmware, and the counter-points could be that OpenWrt is free software, the work is done by volunteers, no warranty implied etc. But this would not improve libpcap support for DSA on Linux, would it?

The workaround for these newer DSA types in libpcap is not yet complete. It is important to make sure it does not break other things for other users on Linux and other operating systems. It will take some time for the solution to reach the end users. Your best course of action would be rolling back to the previous working firmware and waiting until support for this hardware improves. Alternatively, you could try a different hardware that works out of the box. This has been my own experience with OpenWrt in the last ~13 years, and with open source software in general in the last ~25 years (that said, quite often users get the same experience after buying proprietary software). Sometimes such incomplete or lacking hardware support even motivates the users to become contributors or developers themselves.

In future, if anyone expects support for a particular _new_ software or hardware in libpcap, please do your research/testing/QA as soon as you can and report any problems as soon as you can because the correct solution may be more complicated than it seems.

mcr:
I think that you'll only see the DSA tags when dumping on the physical interface that connects to the switch.
So, on my ancient 3800, that means eth0. [eth1 is WAN link].  If you dump on the vlan or bridge interfaces, then you should not any DSA tags.

olerem:
Hi all. Are there any progress on this issue?

infrastation:
Thank you for waiting. As discussed, I have prototyped some code to introduce a new "opaque unstructured data" DLT and to use it in `map_arphrd_to_dlt()` and `iface_dsa_get_proto_info()`, then it became obvious the new DLT would mostly duplicate `DLT_DOCSIS`, which was introduced to solve a similar problem (which arguably no longer exists because the hardware has been EOL for quite a while). However, one meaningful difference is that in the Linux DSA case some implementations use a different encoding for different frame directions, so the new DLT would need to have a direction field. Then I paid attention to the comments in `dsa_protos[]` that suggested using `DLT_LINUX_SLL` for certain DSA types and noted that:
* `DLT_LINUX_SLL2` would be better because it would also save the interface index in a file.
* `DLT_LINUX_SLL2` could work better as the default DLT for all unsupported Linux DSA types.
* In case it does not work exactly as expected, it would be nice to be able to test the changes before making a release.

Then I got distracted by other work. Would you be able to compile the proposed better solution when it becomes ready and to try it on the hardware with the unsupported DSA type?

olerem:
Sure, just ping me.

guyharris:
> Then I paid attention to the comments in dsa_protos[] that suggested using DLT_LINUX_SLL for certain DSA types

I'm not sure whether that was my idea, or somebody else's suggestion that I added in comments, but I'm not sure it's the bet idea; `DLT_LINUX_SLL` and `DLT_LINUX_SLL2` involve capturing from a `PF_PACKET`/`SOCK_DGRAM` socket, and, for those sockets, the Linux kernel strips off link-layer headers, *presumably including the DSA header*, so it wouldn't provide the DSA header even if it were only dissected as a raw binary blob.

What perhaps should be done is:

1. first, have a `LINKTYPE_LINUX_DSA_UNKNOWN`/`DLT_LINUX_DSA_UNKNOWN` to which DSA types that don't yet have their own `LINKTYPE_`/`DLT_` value are assigned - it would prepend a direction flag to the raw packet data (which I think can be done in the reserved space that memory-mapped capture provides, just as we prepend the SLL/SLL2 header for cooked captures), and have tcpdump just "dissect" that as raw hex data;
2. then, over time, assign link-layer types to the DSA types that need them but don't have them and, if they have different transmit and receive headers, prepend a direction flag, and have tcpdump dissect them.

(And update Wireshark to support the new link-layer types as well.)

infrastation:
Apparently, one of my old network devices can run a recent release of OpenWrt, which exposes two DSA network interfaces with live traffic: one uses `brcm-legacy` and the other — `brcm`.  I managed to set cross-compiling up and studied how the current code handles these two types, which have been around for a while and which supposedly could be used to model the support for newer types.

A packet capture on the `brcm-legacy` DSA interface produces `DLT_EN10MB` with EtherType 0x8874. tcpdump does not recognize the EtherType, but can print a complete hex dump of the packets.  Wirehark 4.2.6 recognizes the EtherType, displays the first 5 bytes as "Broadcom tag protocol", but displays the rest of the frame in the hex dump view only, no matter what the packet size.  I did not improve anything in this regard because I do not know what the specification is.

A packet capture on the `brcm` DSA interface produces `DLT_DSA_TAG_BRCM` with the infix tag. Wireshark 4.2.6 does not support the DLT.  tcpdump supports the DLT, but the decoder output did not quite make sense, so I paid closer attention and fixed the decoder (the-tcpdump-group/tcpdump@c207823) and the documentation (the-tcpdump-group/tcpdump-htdocs@4595dc8).

Filtering of packets on `DLT_DSA_TAG_BRCM` did not work.  With a few simple changes my working copy of libpcap seems to support that as expected both for a live capture on the device and for a savefile on my PC. If this still looks good after I add tests for the new code paths, it will be committed next.

After that it should be easier to prototype unsupported DSA handling, commenting the `brcm` element in `dsa_protos[]` seems a good first step. I do not understand yet if the packet direction would be available in all cases.

infrastation:
I had a look at the EtherType'd `brcm-legacy`, which is supposed to deliver `DLT_EN10MB` frames of the following format:
```
original destination MAC-48 address (6B)
original source MAC-48 address (6B)
DSA EtherType 0x8874 (2B)
DSA frame type, flags and length high (1B)
DSA length low (1B)
DSA 0x00 (1B)
DSA port ID (1B)
original EtherType (2B)
```
However, in a file I captured earlier (and in a live capture on the device at the time of this writing) incoming (to CPU) frames are as follows:
```
original destination MAC-48 address (6B)
original source MAC-48 address (6B)
DSA EtherType 0x8874 (2B)
DSA? 0x003b0005 (4B)
802.1Q? 0x81000000 (4B)
DSA? 0x00000402 (4B)
802.1Q? 0x81000001 (4B)
EtherType 0x0800 (2B)
IPv4 header
```
The inner 802.1Q header (?) correlates with 802.1Q VLAN ID 1 prepended to incoming (to CPU) frames on the `brcm` DSA interface. The first 802.1Q header correlates, to an extent, with a comment in `linux/net/dsa/tag_brcm.c:brcm_leg_tag_rcv()`:
```c
        /* The internal switch in BCM63XX SoCs always tags on egress on the CPU
         * port. We use VID 0 internally for untagged traffic, so strip the tag
         * if the TCI field is all 0, and keep it otherwise to also retain
         * e.g. 802.1p tagged packets.
         */
```
However, this does not explain why there are two 802.1Q headers with what looks not entirely dissimilar from another 4 bytes of a DSA header in between. In the same file the additional 4 bytes are present in outgoing (from CPU) frames as well, but this time there are no 802.1Q headers, which is consistent with the `brcm` DSA interface:
```
original destination MAC-48 address (6B)
original source MAC-48 address (6B)
DSA EtherType 0x8874 (2B)
DSA? 0x40000005 (4B)
DSA? 0x20000004 (4B)
EtherType 0x0800 (2B)
IPv4 header
```
I thought the additional 802.1Q header is the result of libpcap reconstructing it from a cooked capture, but further debugging shows the capture is not cooked, and both the additional 802.1Q header and the additional DSA(?) data are present in the data at `frame + tp_mac` in `pcap_handle_packet_mmap()` before the latter changes any data.

This data layout does not seem consistent with the comments in `linux/net/dsa/tag_brcm.c`, but it is far from obvious whether the code and the comments there are exactly the same. **For practical purposes, it would be pointless to try fixing Wireshark and tcpdump for EtherType 0x8874 before it is clear which encoding `brcm-legacy` is supposed to use exactly and whether the frames arrive correctly encoded from Linux kernel.**

infrastation:
Also, in case it is relevant, `vlan_offset` is -1 when capturing from `brcm` and 12 when capturing from `brcm-legacy`.

mcr:
Denis Ovsienko ***@***.***> wrote:
> However, this does not explain why there are two 802.1Q headers with what looks not entirely dissimilar from another 4 bytes of a DSA header in between. In the same file the additional 4 bytes are present in outgoing (from CPU) frames as well, but this time there are no 802.1Q headers, which is consistent with the `brcm` DSA interface:

could it be kernel rebuilding 1q from skbuff tag, but actually it was not
correctly removed?

guyharris:
> could it be kernel rebuilding 1q from skbuff tag,

Kernel or libpcap. What we get from the kernel on Linux has a tag in the packet metadata; we put it back in the pack data. See the code in `pcap_handle_packet_mmap()`:

```
	if (tp_vlan_tci_valid &&
		handlep->vlan_offset != -1 &&
		tp_snaplen >= (unsigned int) handlep->vlan_offset)
	{
		struct vlan_tag *tag;

		/*
		 * Move everything in the header, except the type field,
		 * down VLAN_TAG_LEN bytes, to allow us to insert the
		 * VLAN tag between that stuff and the type field.
		 */
		bp -= VLAN_TAG_LEN;
		memmove(bp, bp + VLAN_TAG_LEN, handlep->vlan_offset);

		/*
		 * Now insert the tag.
		 */
		tag = (struct vlan_tag *)(bp + handlep->vlan_offset);
		tag->vlan_tpid = htons(tp_vlan_tpid);
		tag->vlan_tci = htons(tp_vlan_tci);

		/*
		 * Add the tag to the packet lengths.
		 */
		pcaphdr.caplen += VLAN_TAG_LEN;
		pcaphdr.len += VLAN_TAG_LEN;
	}
```

infrastation:
That's what I suspected, but the unexpected headers are present even before that (or there is something I do not understand).

infrastation:
Meanwhile the next improvements in the `brcm` DSA department are the-tcpdump-group/tcpdump@0b2041f and pull request #1586. Also the complicated handling of `brcm-legacy` DSA and the relatively straightforward handling of `brcm` DSA indicate that it would be a good approach to allocate a DLT to every DSA type by default. The difference is not obvious in tcpdump space, but in libpcap space mixing non-standard Ethernet and standard Ethernet together quickly makes the problem space unmanageable: one time at capture time, and another time at filtering time. 

In the latter sense, I imagine users would often like to filter packets that come from a DSA interface ("how many DHCP packets have appeared on this 10Gb/s DSA interface in the last 24 hours?"), but it would not be an option to generate EtherType-based alternative branches for every variety of pseudo-Ethernet DSA for every use case of `DLT_EN10MB`. Also it does not seem a good solution to introduce additional syntax such as `dsa <sometype> and udp port 67 or 68` — first, the user would have to know which DSA type the interface uses, second, this type would be the same for every packet on the interface anyway. In this sense `DLT_NETANALYZER`, and more recently `DLT_DSA_TAG_EDSA` are examples of a much better solution to this problem.

So, as far as it seems to me, if `bcrm-legacy` gets to work in libpcap as documented, the sooner it becomes a DLT of its own would be the better.

infrastation:
Pull request #1587 implements `DLT_LINUX_DSA_UNKNOWN`,  please review and test if you can. This includes only the frames, not the direction or any other metadata. The frame direction, as far as I understand, is not available. I considered including the ifindex and the tag name (as a string), but then decided to keep it as simple as possible.

With these changes `brcm-legacy` still arrives with more headers than discussed in the spec, so it could be a kernel bug, or an incomplete spec. I do not know if anything depends on that specific interface.

infrastation:
libpcap now supports filtering of `DLT_DSA_TAG_DSA`. It seems a good idea to have at least some support for DLTs that have already been allocated and use a simple tag structure.

infrastation:
I have just tested my working copy cross-compiled for a Linksys EA7500v2 with the same exact OpenWrt as the Netgear 3700v1 above, but different DSA tag:
```
# cat /sys/class/net/eth0/dsa/tagging
mtk

# ./tcpdump -i eth0 -L
Data link types for eth0 (use option -y to set):
  LINUX_DSA_UNKNOWN (Linux DSA unknown tag type, for manual debugging only) (printing not supported)
```
Capturing packets on this interface expectedly produces hex dumps only. In the hex dumps I see what is most likely my SSH session into the device (the MAC addresses and EtherType 0x0800 are visible), but the matter is, only the frames from my PC to the device (network -> CPU) have the additional 4 bytes (0x00100000) between the source MAC address and the EtherType. Frames going from the device to my PC (CPU -> network) have the EtherType right after the source MAC address, in other words, are not DSA-tagged, not using the `mtk` convention, anyway.

From the `net/dsa/tag_mtk.c` file in Linux kernel source it is not obvious whether this is the intended behaviour, also a description of Mediatek DSA tag either does not exist or is not trivial to find on the public Internet. It may be a bug of libpcap, or an undocumented intended behaviour of the Linux driver, or a bug in the Linux driver. **It seems best to leave `mtk` a hexdump-only DSA tag for now.**

vladimiroltean:
> only the frames from my PC to the device (network -> CPU) have the additional 4 bytes (0x00100000) between the source MAC address and the EtherType. Frames going from the device to my PC (CPU -> network) have the EtherType right after the source MAC address, in other words, are not DSA-tagged, not using the mtk convention, anyway.

How are you sending these packets in the first place?
As documented in https://docs.kernel.org/networking/dsa/configuration.html#configuration-with-tagging-support, if you put the IP address on one of the DSA network interfaces (or on a stacked virtual device on top of those, like a bridge, VLAN, macvlan etc), then the packet will pass through the kernel from the user port to the conduit interface and the mtk tagging protocol driver will insert the tag.

If you put your IP address directly on the conduit interface, the DSA tagging protocol's xmit function is bypassed, and what you get is DSA-untagged packets. These should be dropped on ingress by the CPU port of the switch, and will typically not make it to the wire. It sounds like that is what you're doing.

Don't get confused by the "Configuration without tagging support" chapter, which shows you can put the IP address on the conduit interface or one of its stacked upper devices (in that example, eth0.1, eth0.2, eth0.3).

For any piece of hardware, there is no user choice to be made between following one set of configuration steps or the other. If `/sys/class/net/eth0/dsa/tagging` shows `none`, you **have** to use the configuration without tagging, otherwise the one with tagging.

infrastation:
The network interface configuration is how OpenWrt arranges it by default for this model (do not mind the nflog/nfqueue interfaces, that's a side effect of the cross-compile build):
```
# ./tcpdump -D
1.eth0 [Up, Running, Connected]
2.lan4 [Up, Running, Connected]
3.br-lan [Up, Running, Connected]
4.any (Pseudo-device that captures on all interfaces) [Up, Running]
5.lo [Up, Running, Loopback]
6.lan1 [Up, Disconnected]
7.lan2 [Up, Disconnected]
8.lan3 [Up, Disconnected]
9.nflog (Linux netfilter log (NFLOG) interface) [none]
10.nfqueue (Linux netfilter queue (NFQUEUE) interface) [none]
11.wan [none, Disconnected]

2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1504 qdisc mq state UP qlen 1000
    link/ether AA:AA:AA:AA:AA:AA brd ff:ff:ff:ff:ff:ff
    inet6 fe80::XXXXXX/64 scope link 
       valid_lft forever preferred_lft forever
3: wan: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN qlen 1000
    link/ether BB:BB:BB:BB:BB:BB brd ff:ff:ff:ff:ff:ff
4: lan1@eth0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue master br-lan state LOWERLAYERDOWN qlen 1000
    link/ether CC:CC:CC:CC:CC:CC brd ff:ff:ff:ff:ff:ff
5: lan2@eth0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue master br-lan state LOWERLAYERDOWN qlen 1000
    link/ether CC:CC:CC:CC:CC:CC brd ff:ff:ff:ff:ff:ff
6: lan3@eth0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue master br-lan state LOWERLAYERDOWN qlen 1000
    link/ether CC:CC:CC:CC:CC:CC brd ff:ff:ff:ff:ff:ff
7: lan4@eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master br-lan state UP qlen 1000
    link/ether CC:CC:CC:CC:CC:CC brd ff:ff:ff:ff:ff:ff
8: br-lan: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP qlen 1000
    link/ether CC:CC:CC:CC:CC:CC brd ff:ff:ff:ff:ff:ff
    inet XXXX/24 brd XXXX scope global br-lan
       valid_lft forever preferred_lft forever
    inet6 fe80::XXXXXX/64 scope link 
       valid_lft forever preferred_lft forever
```
In this case my PC connects to the IPv4 address XXXX and the captured packets use MAC address CC:CC:CC:CC:CC:CC (that is, the SSH client seems to communicate with `br-lan` as expected). So based on your comment above, the configuration is supposed to work (and it does in fact work because I log into the device using SSH), but the DSA tag is not there.

Having written that, I checked again and realised that my previous comment was wrong: the DSA tag is present in _outgoing_ packets, but is absent in _incoming_ packets. Could you confirm if this is what is supposed to come from the driver? I could then debug libpcap end of the capture if necessary.

vladimiroltean:
> Having written that, I checked again and realised that my previous comment was wrong: the DSA tag is present in outgoing packets, but is absent in incoming packets. Could you confirm if this is what is supposed to come from the driver? I could then debug libpcap end of the capture if necessary.

Ah, that's entirely different.

Some DSA switches are tightly integrated with their host port (typically if the switch is from the same vendor and is integrated into the SoC) and implement hardware offloading of the DSA tag parsing in the conduit Ethernet driver. This is currently implemented only in the ingress direction.

If you search for https://elixir.bootlin.com/linux/v6.17.8/A/ident/METADATA_HW_PORT_MUX in the kernel source code, you'll see that a small number of device drivers (mtk_eth_soc.c, airoha_eth.c) use this mechanism to avoid "cooking" a DSA tag which is pushed inside the packet, since they will already have DSA information in their receive descriptors. That's how these systems work, and if DSA finds an skb with skb_metadata_dst of the METADATA_HW_PORT_MUX type at its ingress, it uses the source port from that metadata structure directly, having understood that there is no DSA tag to be found in the packet: https://elixir.bootlin.com/linux/v6.17.8/source/net/dsa/tag.c#L71

I'm unsure if libpcap can see that metadata structure though.

infrastation:
Thank you for explaining. One problem here is that the `mtk` DSA tag is not documented at the time of this writing, correct? If hypothetically it was documented and the specification said the 4-byte infix header applies to both Tx and Rx frames, as far as the wire encoding goes, then a driver that applies the tag to the outbound frames only would not match the specification and ought not to declare `mtk` as the tag name, correct?

vladimiroltean:
The only documentation is the source code or vendor datasheets (if you have those) for most switch tagging protocols.

The specification is sufficiently lax that this metadata structure can be taken by the kernel for the tag itself, without the need of faking it into the packet (which defeats the point of an offload). It doesn't really change the fact that the tagging protocol is "mtk", save for the fact that the RX tag was consumed prior to entering the network stack. Taken another way: connected to a different Ethernet controller which was unaware of MTK DSA tags, you would have seen the tag in the packet if it was the same switch IP. If you look at net/dsa/tag_mtk.c, you'll find an actual receive procedure for this protocol, which is invoked for such cases, but bypassed for offloads.

This decision to keep naming it "mtk" rather than "mtk_rx_offloaded" was taken without thinking too much about what pcap wants, mainly due to not understanding what pcap wants. 

For example, AF_PACKET has TP_STATUS_VLAN_VALID based on which pcap reconstructs offloaded VLAN headers. Would similar bits for the port mux metadata help here?

infrastation:
One of the problems libpcap tries to solve is delivering the captured packets in a format that encodes the right amount of detail, is sufficiently well documented, stable, and easy to process — the latter includes live kernel filtering (on and off the `any` pseudo-interface), live userland filtering (on and off the `any` pseudo-interface), offline userland filtering and subsequent decoding by a protocol analyser.

One traditional way to avoid unnecessary complications is to avoid mixing conflicting protocols/formats in one DLT, which is the main pressing reason to resolve this matter in a way that does not add a new problem space.

Another traditional way is to use the same packet encoding at as many stages as possible, because it is rather cumbersome (and tends to spawn bugs) to generate (and to debug!) a different filter program in different contexts, as mentioned above.

In the sense of encoding libpcap often aims to deliver packets that are as close to what goes onto the wire and comes from the wire, as can be seen by another host monitoring the same wire externally. As you explain, for DSA-tagged frames the "wire" is an internal link between different components of the same host and can be driven in one or another driver-specific way so long as packets make it through fine. It seemingly does not matter until there is a problem and it becomes necessary to see what is being transmitted and received.

If the user is interested in the Ethernet frame only, it would be more appropriate to capture on the pure Ethernet presentation of the data, such as `lan1@eth0` above, instead of taking the inconvenience of capturing on `eth0` and stripping each DSA tag individually to produce proper `DLT_EN10MB`, and working around cases that have the tag pre-stripped on some packets. This effectively would be a more difficult userland reimplementation of what the kernel does already.

If the user is interested in the DSA tag as well (which physical port did this frame come from?), then the correct solution would be to have a tag in every frame and to deliver `DLT_DSA_xxxxxxx`. To have a tag in every frame, it would not be unfeasible to reconstruct it from the metadata in libpcap as is done for 802.1Q — if the user needs to know the Rx port, this metadata needs to be somewhere, whether the driver encoded it in the frame or not. However, considering how much difficulty 802.1Q support in libpcap has experienced on Linux and how many DSA support is experiencing now, let's think twice before attempting that. Also, arguably, if some driver switches to descriptors for both Tx and Rx, that may justify a separate DLT, and perhaps such a fluid problem space isn't a good fit for libpcap, it requires specialized kernel debugging tools.

Much of this discussion boils to the relation between a network stack (this is how to calculate a checksum in this piece of hardware, which is not the CPU) and a network packet (for this packet in this header this checksum is exactly this value). Notwithstanding the benefits of parallelisation, hardware offloading, scatter/gather, GSO etc. and internal flexibility at run time in the OS kernel(s), in network protocols data is supposed to be eventually/virtually serialised into a packet. That's where libpcap perspective usually is, which can be different from Linux kernel perspective.

That said, if you capture transmitted IPv4 packets on a host that lets the NIC do checksum offloading, you will see that the values do not match the packets, but the placeholder checksum fields are there with zeroes or random data, so that the other header fields can be found exactly where they are supposed to be. The received packets will have their checksum(s) coming from the network, but despite the internal differences a network analyser can parse the resulting protocol header as usual. That's not to say every design should be like this, but it is a worthwhile approach to consider.

In practical terms, the first priority is to stop mixing Linux DSA into `DLT_EN10MB`, the second priority is to enable minimal debugging of Linux DSA, the third priority is to add support for DSA tags that are documented and trivial to implement/maintain (contributions would be welcome). After these changes get sufficiently well documented, deployed and prove to work well, it would be a good time to consider support for more DSA tags, in terms of available documentation, implementation difficulty, user demand and available resources. If anybody sees a better plan, please make your point before long.

vladimiroltean:
> I had a look at the EtherType'd `brcm-legacy`, which is supposed to deliver `DLT_EN10MB` frames of the following format:
> 
> However, this does not explain why there are two 802.1Q headers with what looks not entirely dissimilar from another 4 bytes of a DSA header in between. In the same file the additional 4 bytes are present in outgoing (from CPU) frames as well, but this time there are no 802.1Q headers, which is consistent with the `brcm` DSA interface:

Can you list the interfaces with "ip a" so I can see their iflinks and understand the topology?

I think I forgot to tell you about tag stacking, where DSA switches which don't natively understand each others' tags can still be connected to each other. In that case, `/sys/class/net/eth0/dsa/tagging` holds the tagging protocol of the switch closest to the host port, and the user ports of that DSA switch can be DSA conduits themselves, so `/sys/class/net/swp0/dsa/tagging` will also exist and hold a different value. You shouldn't expect that if a DSA tag exists, it's the only tag.

vladimiroltean:
Worse, the order in which the tags need to be processed is purely given by the hardware topology, not by the packet layout. Consider the following (perhaps unrealistic, but theoretically possible) hardware design:

```
   +-----------------------------------+
   |+------+ +------+ +------+ +------+|
   || sw2p0| |sw2p1 | |sw2p2 | |sw2p3 ||
   |+------+-+------+-+------+-+------+|
   |          DSA switch driver        |
   +-----------------------------------+
                                  |  ^
                                  |  |
                                  |  |
                                  v  v
   +-----------------------------------+
   |+------+ +------+ +------+ +------+|
   ||sw1p0 | | sw1p1| | sw1p2| |sw1p3 ||
   |+------+-+------+-+------+-+------+|
   |          DSA switch driver        |
   +-----------------------------------+
                                  |  |
                                  |  |
                               +--v--|-----------------------------+
                               |+------+ +------+ +------+ +------+|
                               ||sw0p0 | |sw0p1 | |sw0p2 | |sw0p3 ||
                               |+------+-+------+-+------+-+------+|
                               |          DSA switch driver        |
                               +-----------------------------------+
                                             |        ^
                                Tag added by |        | Tag consumed by
                               switch driver |        | switch driver
                                             v        |
                               +-----------------------------------+
                               | Unmodified host interface driver  | Software
-------------------------------+-----------------------------------+------------
                               |       Host interface (eth0)       | Hardware
                               +-----------------------------------+
                                             |        ^
                             Tag consumed by |        | Tag added by
                             switch hardware |        | switch hardware
                                             v        |
                               +-----------------------------------+
                               |               Switch              |
                               |+------+ +------+ +------+ +------+|
                               ||sw0p0 | |sw0p1 | |sw0p2 | |sw0p3 ||
                               ++------+-+------+-+------+-+------++
                                  |  ^
                                  |  |
                                  |  |
                                  v  v
   +-----------------------------------+
   |               Switch              |
   |+------+ +------+ +------+ +------+|
   || sw1p0| |sw1p1 | | sw1p2| | sw1p3 ||
   ++------+-+------+-+------+-+------++
                                  |  ^
                                  |  |
                                  |  |
                                  v  v
   +-----------------------------------+
   |               Switch              |
   |+------+ +------+ +------+ +------+|
   || sw2p0| |sw2p1 | | sw2p2| | sw2p3||
   ++------+-+------+-+------+-+------++
```

When you direct tcpdump to dissect ingress traffic on eth0, that traffic can have 1 tag (if it came from the sw0p1, sw0p2 or sw0p3 hardware ports, and will go towards the network interfaces of the same name).
Or it can have 2 tags, if it came from the sw1p0, sw1p1 or sw1p2 ports.
Or it can have 3 tags, if it came from the sw2p0, sw2p1, sw2p2 or sw2p3 ports.
You don't know what to expect, because you'd need to reconstruct the packet's path through the kernel stack, to see which user port processed which tag and in which order.
You can't figure out the order looking at just the packet, because not all DSA tags are headers, or trailers, they can be a combination. So the packet:

```
+---------+---------+--------------+---------------+-----------+----------+--------------+
| MAC DA  | MAC SA  | DSA header 1 |  DSA header 2 | EtherType | .......  | DSA trailer  |
+---------+---------+--------------+---------------+-----------+----------+--------------+
```

could have equally come from:
- Switch 0 inserted DSA header 1, Switch 1 inserted DSA header 2, Switch 2 inserted DSA trailer
- Switch 0 inserted DSA header 1, Switch 1 inserted DSA trailer, Switch 2 inserted DSA header 2

infrastation:
The diagnostics is below. That said, if you have time to spend on this matter, please consider spending it on identifying DSA tags with the least surprising implementations.
```
# grep . /sys/class/net/*/dsa/tagging
/sys/class/net/eth0/dsa/tagging:brcm-legacy
/sys/class/net/extsw/dsa/tagging:brcm

# ip ad
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host proto kernel_lo 
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1510 qdisc fq_codel state UP group default qlen 1000
    link/ether DD:DD:DD:DD:DD:DD brd ff:ff:ff:ff:ff:ff
    inet6 fe80::XXXXXX/64 scope link proto kernel_ll 
       valid_lft forever preferred_lft forever
3: extsw@eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1504 qdisc noqueue state UP group default qlen 1000
    link/ether DD:DD:DD:DD:DD:DD brd ff:ff:ff:ff:ff:ff
    inet6 fe80::861b:5eff:fe48:5bcf/64 scope link proto kernel_ll 
       valid_lft forever preferred_lft forever
4: wan@extsw: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master switch state UP group default qlen 1000
    link/ether DD:DD:DD:DD:DD:DD brd ff:ff:ff:ff:ff:ff
5: lan4@extsw: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master switch state UP group default qlen 1000
    link/ether DD:DD:DD:DD:DD:DD brd ff:ff:ff:ff:ff:ff
6: lan3@extsw: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master switch state UP group default qlen 1000
    link/ether DD:DD:DD:DD:DD:DD brd ff:ff:ff:ff:ff:ff
7: lan2@extsw: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master switch state UP group default qlen 1000
    link/ether DD:DD:DD:DD:DD:DD brd ff:ff:ff:ff:ff:ff
8: lan1@extsw: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master switch state UP group default qlen 1000
    link/ether DD:DD:DD:DD:DD:DD brd ff:ff:ff:ff:ff:ff
9: switch: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether DD:DD:DD:DD:DD:DD brd ff:ff:ff:ff:ff:ff
    inet6 fe80::XXXXXX/64 scope link proto kernel_ll 
       valid_lft forever preferred_lft forever
10: switch.1@switch: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether DD:DD:DD:DD:DD:DD brd ff:ff:ff:ff:ff:ff
    inet XXXX/24 brd XXXX scope global switch.1
       valid_lft forever preferred_lft forever
    inet6 fe80::XXXXXX/64 scope link proto kernel_ll 
       valid_lft forever preferred_lft forever

[    0.949413] bcm63xx-spi 10000800.spi: at [mem 0x10000800-0x10000f0b flags 0x200] (irq 9, FIFOs size 542)
[    1.036180] bcm6368-mdio-mux 10f000b0.mdio: Broadcom BCM6368 MDIO mux bus
[    1.048069] b53-switch spi0.1: found switch: BCM53115, rev 8
[    1.054880] b53-switch 10f00000.switch: found switch: BCM63xx, rev 0
[    1.062634] bcm6368-enetsw 10006800.ethernet: mtd mac DD:DD:DD:DD:DD:DD
[    1.148271] bcm6368-enetsw 10006800.ethernet: eth0 at 0xb0006800, IRQ 0
[    1.156868] bcm7038-wdt 1000005c.watchdog: Registered BCM7038 Watchdog
[...]
[    1.468746] b53-switch spi0.1: found switch: BCM53115, rev 8
[    1.475177] b53-switch 10f00000.switch: found switch: BCM63xx, rev 0
[    1.598249] b53-switch 10f00000.switch: Using legacy PHYLIB callbacks. Please migrate to PHYLINK!
[    1.612100] bcm6368-enetsw 10006800.ethernet eth0: entered promiscuous mode
[    1.619832] DSA: tree 0 setup
[    1.625556] b53-switch spi0.1: found switch: BCM53115, rev 8
[    1.730583] b53-switch spi0.1: Using legacy PHYLIB callbacks. Please migrate to PHYLINK!
[    1.743412] b53-switch spi0.1: Configured port 8 for rgmii
[    1.753712] b53-switch spi0.1 wan (uninitialized): PHY [dsa-1.0:00] driver [Generic PHY] (irq=POLL)
[    1.771583] b53-switch spi0.1 lan4 (uninitialized): PHY [dsa-1.0:01] driver [Generic PHY] (irq=POLL)
[    1.788619] b53-switch spi0.1 lan3 (uninitialized): PHY [dsa-1.0:02] driver [Generic PHY] (irq=POLL)
[    1.805371] b53-switch spi0.1 lan2 (uninitialized): PHY [dsa-1.0:03] driver [Generic PHY] (irq=POLL)
[    1.823293] b53-switch spi0.1 lan1 (uninitialized): PHY [dsa-1.0:04] driver [Generic PHY] (irq=POLL)
[    1.842240] b53-switch 10f00000.switch extsw: entered promiscuous mode
[    1.849407] DSA: tree 1 setup
```

vladimiroltean:
Like I said.

eth0 is the physical Ethernet controller doing the DMA.
extsw@eth0 is the user port of the first DSA switch connected to eth0.
wan@extsw, lan1@extsw, lan2@extsw, lan3@extsw, lan4@extsw are user ports of the second DSA switch connected to extsw.
"switch" is a bridge interface (yay for naming!) and switch.1 is a VLAN upper of the bridge.

vladimiroltean:
> if you have time to spend on this matter, please consider spending it on identifying DSA tags with the least surprising implementations

What are you trying to achieve?

Looking at hex dumps with DLT_EN10MB was perfectly fine. Essentially the only 2 cases where you'd need that were:
- You have a connectivity problem and you want to see the tag to understand why the Linux drivers don't redirect the packet towards the correct port
- You're Alice lost in wonderland

You'll see the same packet data presented again to you if you open tcpdump on the user ports instead of the conduit interface, so you can put filters there all you want, and the kernel has stripped the tags for you in the correct order.

infrastation:
As discussed before, the hex dumps are going to be available for Linux DSA, but in a way that does not cause problems to other users of libpcap and associated file formats. Let's be wise about which problems to solve and which problems to avoid creating.

vladimiroltean:
If you want to have fun, you can compile the Linux kernel for a QEMU virtual machine or some sort of throwaway board with `CONFIG_NET_DSA_LOOP=y`. This is a dummy DSA driver that attaches to whatever interface in your system is named "eth0" (the name is hardcoded in the source code) and creates 4 interfaces, lan1 to lan4. If you modify `dsa_loop_get_protocol()` in the kernel, you can set this driver to use any of the available DSA tagging protocols.

From there on, you can at least send some traffic through lan1 and look at tcpdump on eth0, and you'll see the DSA driver add tags automatically.

Reception is more difficult, you can craft packets with a DSA tag put into them and inject them into eth0's RX, then dsa_loop should decode them correctly and redirect towards the proper user port if they're OK, and reject them otherwise.
We don't have any "reverse header injector" software as far as I'm aware, all traffic testing of this sort that I've done was with manual packet header crafting. Perhaps it could be created as a user space application driving a tun/tap interface, but I haven't had this amount of time to investigate the possibilities.

infrastation:
That could be useful if it allowed to connect two VMs back to back DSA-tagged, for example, and to test them passing traffic forth and back. Given how many DSA tags there is, it would not be realistic to maintain a complete collection of routers just for the hardware-based tests, and to run the tests on a regular basis.

I can get an occasional piece of hardware for debugging if nothing else works, but often a virtual test lab would do, as eventually turned out with SocketCAN, after I got the adapters. For ARCnet tests the only practicable solution was to use actual hardware because nobody bothered to implement virtual ARCnet-over-non-ARCnet adapters.

vladimiroltean:
> That could be useful if it allowed to connect two VMs back to back DSA-tagged, for example, and to test them passing traffic forth and back. Given how many DSA tags there is, it would not be realistic to maintain a complete collection of routers just for the hardware-based tests, and to run the tests on a regular basis.

No, two VMs back to back isn't how this works, given that DSA tags are asymmetric. You need a software entity that represents a switch. It consumes tags added by the xmit procedure of the tagger on eth0 egress, and inserts (different) tags when sending packets to eth0 ingress, so that the dsa_loop can decode them. This is the piece of software that I said doesn't exist, but maybe it could be written in user space with relative ease with tun/tap.

infrastation:
I considered the DSA diagram above for a while, and one thing that looks off there is that some of the peripheral arrows are pointed one way, some are pointed both ways, and some are not pointed at all. Also, as far as I understand, double- and triple-tagging would work as described only for packets that go between a user port and the CPU; in other cases, for example, trying to switch directly between the user ports sw2p2 and sw0p2 would require chip sw1 to drive chip sw2 directly and would require chip sw0 to drive chip sw1 directly. In which case, even if you disregard the differences between different DSA tags and respective hardware vendors, perhaps it would simplify the matter a lot if the root switch chip (sw0) absorbed all downstream network topology, reported 9 user ports and always presented exactly one DSA tag to the host.

On a more general note about protocol encoding, stacking of tags/labels tends to scale reasonably well only when the type of protocol is the same for the entire stack (e.g. MPLS). Multiple-level 802.1Q is a specific case of one common network design that identifies every nested protocol explicitly (e.g. EtherType, IP protocol number). Another common design is to specify exactly one nested protocol (e.g. IP-within-IP). Of course, a host's internal bus is not the Internet, and it is fine to cut some corners in DSA, until the bus needs to be more like a network, then the mismatch between the problem space and the solution space becomes larger.

As discussed earlier, libpcap's view of a bus/network is a packet that comes out of it, and any topology exists only insofar as the packet encoding and any available metadata convey it. In this sense multiple-level DSA tagging is a relatively difficult problem space, so please do not expect too much in the solution space too soon. It has to be done one safe step at a time.

vladimiroltean:
> one thing that looks off there is that some of the peripheral arrows are pointed one way, some are pointed both ways, and some are not pointed at all

Yeah, sorry, I started from this picture and insufficiently adapted it: https://docs.kernel.org/networking/dsa/dsa.html#graphical-representation

> double- and triple-tagging would work as described only for packets that go between a user port and the CPU; in other cases, for example, trying to switch directly between the user ports sw2p2 and sw0p2 would require chip sw1 to drive chip sw2 directly and would require chip sw0 to drive chip sw1 directly.

User ports don't switch packets between each other unless they're a part of the same bridge. And even if they are, they can switch directly (if they part of the same so-called hwdom - hardware forwarding domain) or indirectly (if they are part of different hwdoms). The hwdoms are [bridge concepts](https://elixir.bootlin.com/linux/v6.17.8/source/net/bridge/br_switchdev.c#L212) based on netif_get_port_parent_id() that [DSA associates with the tree index](https://elixir.bootlin.com/linux/v6.17.8/source/net/dsa/devlink.c#L321) (i.o.w. switch ports from the same tree can forward autonomously from one to another; switches from different trees can't).

When a switch doesn't know what to do with a packet it floods it, and one of the flooding destinations is the CPU. There, DSA decapsulates one by one all receive headers, sees it came from sw2p2, and gives the packet to the bridge. The bridge sees that sw0p2 is its other port, checks its hwdom, sees it's different than sw2p2, so it figures out it couldn't have reached that port in hardware, so it sends it in software. The packet is reencapsulated with the DSA tags required to reach sw0p2.

> perhaps it would simplify the matter a lot if the root switch chip (sw0) absorbed all downstream network topology, reported 9 user ports and always presented exactly one DSA tag to the host

The root switch can't present more ports than it physically has, DSA switches are physical and that's it. The switch tree abstraction does however permit exactly what you say - circulation among 9 user ports with a single tag. In a sense, a switch tree is the hardware domain in which that tag acts as valid 'currency' and the other switches understand it. Normally, the goal is to have a single switch tree for directly connected switches, if at all possible. But you're free to connect switches which don't understand each others' tags, and your own board shows vendors do that. In that case, each switch would form its own single-element tree, and would encapsulate any other DSA tag in its own tag.

Fact of the matter is that both models exist, but you can't simplify the "disjoint trees" case to the "single tree" one.

> In this sense multiple-level DSA tagging is a relatively difficult problem space, so please do not expect too much in the solution space too soon. It has to be done one safe step at a time.

I'm not expecting anything, I'm just explaining (perhaps a bit too late, perhaps a bit ahead of time) that expecting filters to properly run on DSA-tagged packets is only ever going to properly work if there's a single DSA tag in them. In the other cases, you don't know what format to adjust for.

infrastation:
The current master branch implements varying degrees of support for a few DSA tags ("brcm", "brcm-prepend", "dsa" and "edsa") and captures DSA-tagged frames as raw debugging material for unsupported DSA tags. Please test it and confirm your results, this solution is going to be available in libpcap-1.10.6.

infrastation:
If anybody in future is going to add support for the `mtk` DSA tag (the nominal variety, which tags frames in both directions) using a dedicated DLT, it would seem most reasonable not to use the DLT if the interface claims the tag, but uses a driver that multiplexes the switch→CPU DSA tag away, such as is the case with my test device:
```
# cat /sys/class/net/eth0/dsa/tagging 
mtk

# ethtool -i eth0
driver: mtk_soc_eth
version: 6.6.110
firmware-version: 
expansion-rom-version: 
bus-info: 1e100000.ethernet
supports-statistics: yes
supports-test: no
supports-eeprom-access: no
supports-register-dump: yes
supports-priv-flags: no
```

In practical terms this would mean using `ETHTOOL_GDRVINFO` and comparing the driver with `mtk_soc_eth`.

infrastation:
The better solution (not mixing DSA tag and Ethernet) has been backported to the libpcap-1.10 branch and works as expected in my tests using `brcm`, `brcm-legacy` and `mtk`. If anybody is interested to test this code before it becomes available in libpcap 1.10.6, now would be the right time.

infrastation:
libpcap 1.10.6 is supposed to get this right, please test and post your results.

infrastation:
To the best of my understanding this problem has been solved and the solution has not created new problems. Closing as resolved.

PR #1455, opened 2025-02-02
support old Perl and old Linux better

infrastation:
Solaris 10 support for libpcap testing is the main change. I tested it on an old Linux and discovered more breakage in the main code. One bug fix is included and the other (Linux BPF extensions do not detect properly after my recent change) is in my working copy, it requires a bit more finish before committing. Meanwhile let's see if the Perl part passes CI on all recent systems.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2139,6 +2139,11 @@ else(WIN32)
         endif()
 
         check_struct_has_member("struct tpacket_auxdata" tp_vlan_tci linux/if_packet.h HAVE_STRUCT_TPACKET_AUXDATA_TP_VLAN_TCI)
+
+        # This check is for TESTrun purposes, not for the C code.
+        # This and the #cmakedefine01 in cmakeconfig.h.in reproduce exactly the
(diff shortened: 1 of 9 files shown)


infrastation:
The new revision runs out of memory, this requires more work.

infrastation:
It worked as expected in my manual test runs too, one of which was on the OpenBSD host that runs the worker. So there is an additional factor somewhere.

infrastation:
The OOM reproduces if IPv6 support is disabled (this is what the first build matrix setup does, which is what OOMed, and not what both build systems do by default, which is what I prototyped and tested). I forgot to run build matrix locally before opening the pull request, it seems.

infrastation:
Apparently, the first test that should be skipped when IPv6 is disabled and has aliases triggers the bug because `scalar ($test->{aliases})` returns a number large enough to fill the memory with test requests. `scalar (@{$test->{aliases}})` returns the number of aliases, as expected. The next revision will include the bug fix, I am still finishing the pending bug fix for old Linuxes.

infrastation:
Fixed a commit message only. Tested manually to various extents on Debian 6, Ubuntu 14.04, Solaris 10, Haiku and Hurd. This is going to be merged later today.

PR #1456, opened 2025-02-06
Factor TESTlib.pm out.

infrastation:
The purpose of these changes is to have as much shared code to reuse in tcpdump `TESTrun` as possible (that is going to be a separate pull request soon).

These changes in libpcap have been manually tested to varying extents on Solaris 10, Debian 6, Ubuntu 14.04 and Haiku. Let's test on the other OSes as well.


infrastation:
This revision gets the IGRP skip condition right.

PR #1458, opened 2025-02-09
Do not define offsetof().

infrastation:
In theory, this change is a no-op.


infrastation:
On a second thought, `pcap-int.h` includes `<stddef.h>`, so the diff is a bit larger than it needs to be.  Which also means other files that include `pcap-int.h` should not include `<stddef.h>`.  This will require a second revision, but let's see if CI finds any issues with this one.

guyharris:
Why is the cbpf-manfile change part of this change? It doesn't seem related.

guyharris:
The same applies to the pcap-filter man page changes.

guyharris:
And should the gencode.c whitespace changes also be done in a separate commit/pull request?

fmtutils.c
```diff
@@ -39,7 +39,6 @@
 
 #include "ftmacros.h"
 
-#include <stddef.h>
```
guyharris, line 42: On Windows, this file uses `wchar_t`, so it has a reason of its own to include stddef.h.

guyharris: ...unless we make an explicit commitment that pcap-int.h will always include stddef.h.

infrastation:
`offsetof()` is one of several small fixups I've been meaning to make, but I managed to address a few other to-do items as well. Each is a separate commit.

fmtutils.c
infrastation: So long as `pcap-int.h` continues to use `offsetof()`, it will include `<stddef.h>`. Arguably, each file that needs `offsetof()` ought to do the same instead of relying on another header. Let me see how it fits.

guyharris: stddef.h is for more than `offsetof()`; it's also for `ptrdiff_t`, `size_t`, `wchar_t`, and `NULL`, some of which are also defined in other headers.  Files that need one of those should include *some* header that defines it.

guyharris:
> `offsetof()` is one of several small fixups I've been meaning to make, but I managed to address a few other to-do items as well. Each is a separate commit.

So will those all show up as separate commits to the main repository when this is merged, or will they be squashed into a single commit?

infrastation:
The commits are going to stay separate, I am not going to squash them.

infrastation:
This revision does not rely on `pcap-int.h` to include `<stddef.h>`.

PR #1459, opened 2025-02-11
Added nix flake for ease of building

theredwiking:
I have been using nix for building projects ease and cross-platform
I added one to this project for making it ease for build and testing for my zig wrapper


infrastation:
The new file does not look specific to libpcap in any way, please confirm this pull request went to the correct project.

theredwiking:
It helps build libpcap using nix environment for consistency and easy on differentieret distros

guyharris:
Why does it need the Boost libraries?

infrastation:
These changes look as if they boil down to a list of dependencies in a particular packaging system (nix?). In this case a more appropriate place for such a file would be in the respective system. For example, Linux distributions use their own definition files to produce `.rpm` and `.deb` packages with the required contents and dependencies, *BSD systems and pkgsrc each use their own means (typically a `Makefile` with some patches), and so on. Keeping such files in the packager's repository rather than the upstream repository is a very common and the most reasonable arrangement. Closing.

PR #1460, opened 2025-02-12
assorted improvements to the build checks

infrastation:
The intent here is to make the single-threaded part of the build processes shorter.

This looks about the right number of commits and ready for a CI run. I have tested these changes manually on several OSes to varying extent. This may require some finish later.


infrastation:
This revision moves the `vsyslog()` check further away into `rpcapd`, makes Homebrew detection conditional, splits one combined commit into a few smaller ones and refines DPDK message printing.

PR #1465, opened 2025-02-15
external test props

infrastation:
These changes add initial filter tests for the `gateway` keyword. Manually tested on various systems and found to work as expected except on illumos (did not debug this yet). Let's see what CI results look like (most Buildbot worker hosts have been already been configured to have the new entries in `/etc/hosts` and `/etc/ethers`).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -101,6 +101,8 @@ DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
       Add kernel filtering, fix userland filtering.
     OpenBSD:
       Use getprotobyname_r() correctly on OpenBSD.
+    Linux:
+      Do not use ether_hostton() from musl libc.
     DAG:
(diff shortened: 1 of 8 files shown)


infrastation:
Also pcap-filter(7) may need an update in the `gateway` description, but I have not yet figured it out completely.

infrastation:
This revision is expected to pass all tests.

infrastation:
Tested manually to varying extents on hosts with props: OmniOS, Haiku, GNU/Hurd, Debian 6, Void Linux (musl), Solaris 11.4 CBE; without props: Solaris 10.  Seems to work, except illumos and Solaris 11.4 seem to take 5-15 minutes to pick up the changes to `/etc/ethers`, and Perl threads seem to crash quietly on Void Linux (which is not specific to these changes, same on the master branch).

PR #1468, opened 2025-02-17
TESTrun: more improvements

infrastation:


infrastation:
The `TESTRUN_PERL` change breaks CMake build on Solaris 10, looking...

PR #1472, opened 2025-02-20
test "net NAME"

infrastation:


infrastation:
Manual tests look as expected on: Haiku (with `/etc/networks`), Hurd (with `/etc/networks`), Solaris 10 (without `/etc/networks`).

infrastation:
One of the reasons the linux-s390x build failed is because there is a bug in libpcap, working on the bug fix now.

guyharris:
> One of the reasons the linux-s390x build failed is because there is a bug in libpcap, working on the bug fix now.

Is it code that was (explicitly or implicitly) assuming it's running on a little-endian machine?

infrastation:
The difference was not endianness, but the absence of `/etc/services` and `/etc/protocols` (the container did not have the `netbase` Debian package by default). The former caused `port NAME` and related syntax to fail correctly, but the latter caused `proto NAME` and related syntax generate the filter program with protocol 0 because the rare condition of an error from `getprotobyname_r()` was not handled properly.

infrastation:
(Force-pushed to reword the commit message, no changes to the code.)

Issue #863, opened 2019-10-15
Enhancement: parse special offsets in filter expression


fenner:
#861 talks about output, but input is also important.  I would much rather say "to look at skb_type you use `ether[type]=3`" instead of "`ether[-4092]=3`" and have to figure out if that is the value I really meant.

I'd be fine (it might be better, even) if it was introduced by a different value - `special[type]`?  `skf_ad[type]`?  (I'm bad at naming!)

guyharris:
I might vote for `skf_ad` to remind users that it's for Linux `SKF_AD_` values.

The compile would fail unless you're compiling for a live capture on a `PF_PACKET` socket.

infrastation:
That's an interesting idea. Please note that in BPF syntax all named offsets use the same namespace and it could not be clear in future which specific type the `type` means, so these offsets would need to use a prefix, let's say `ad_`. Then the index operation could use one of the following notations:
```
skf[ad_type]
linux[ad_type]
linuxskf[ad_type]
```
If done this way, it would be easier to relate the input filter expressions and the output of `bpf_dump()` if they used the same nomenclature, because currently it is:
```
(000) ldh      [type]
```
Right now `bpf_image()` recognizes the following named offsets:
```c
[SKF_AD_ALU_XOR_X] = "xor_x",
[SKF_AD_CPU] = "cpu",
[SKF_AD_HATYPE] = "hatype",
[SKF_AD_IFINDEX] = "ifidx",
[SKF_AD_MARK] = "mark",
[SKF_AD_NLATTR_NEST] = "nlan",
[SKF_AD_NLATTR] = "nla",
[SKF_AD_PAY_OFFSET] = "poff",
[SKF_AD_PKTTYPE] = "type",
[SKF_AD_PROTOCOL] = "proto",
[SKF_AD_QUEUE] = "queue",
[SKF_AD_RANDOM] = "random",
[SKF_AD_RXHASH] = "rxhash",
[SKF_AD_VLAN_TAG_PRESENT] = "vlanp",
[SKF_AD_VLAN_TAG] = "vlan_tci",
[SKF_AD_VLAN_TPID] = "vlan_tpid"
```
(Internally libpcap uses only `SKF_AD_VLAN_TAG`, `SKF_AD_VLAN_TAG_PRESENT`, `SKF_AD_IFINDEX`, `SKF_AD_PKTTYPE` and `SKF_AD_PROTOCOL`.)

Does it make sense?

PR #1473, opened 2025-02-21
Add VLAN tag packet filter support for Npcap when available

dmiller-nmap:
These changes are part of Npcap 1.81 and require Npcap SDK 1.15 to build, though I believe it will build safely either way, just without the new features. See nmap/npcap#171 for discussion and related changes in Npcap.


infrastation:
As far as I understand, the new code reuses `SKF_AD_VLAN_TAG_PRESENT` on Windows. Does it reimplement the lower (non-VLAN) auxiliary data items exactly the same as on Linux?

guyharris:
> As far as I understand, the new code reuses `SKF_AD_VLAN_TAG_PRESENT` on Windows. Does it reimplement the lower (non-VLAN) auxiliary data items exactly the same as on Linux?

The only `SKF_AD_` bits of metadata it implements are `SKF_AD_VLAN_TAG` and `SKF_AD_VLAN_TAG_PRESENT`; it doesn't implement any of the others.

infrastation:
Then at the very least this constant in Windows-specific code should have a comment to explain what it means in Npcap, because in Linux it means "everything below that is available".

guyharris:
> Then at the very least this constant in Windows-specific code should have a comment to explain what it means in Npcap, because in Linux it means "everything below that is available".

Which constant is that? `SKF_AD_MAX`? It means "everything numerically below that is available" in Npcap; see[ that part of npcap-bpf.h](https://github.com/nmap/npcap/blob/master/Common/npcap-bpf.h#L207).

dmiller-nmap:
That's correct: I tried to keep the same semantics as on Linux with the `SKF_AD_*` constants. Compare `pcap-linux.c`:
```c
#if defined(SO_BPF_EXTENSIONS) && defined(SKF_AD_VLAN_TAG_PRESENT)
	/*
	 * Can we generate special code for VLAN checks?
	 * (XXX - what if we need the special code but it's not supported
	 * by the OS?  Is that possible?)
	 */
	if (getsockopt(sock_fd, SOL_SOCKET, SO_BPF_EXTENSIONS,
	    &bpf_extensions, &len) == 0) {
		if (bpf_extensions >= SKF_AD_VLAN_TAG_PRESENT) {
			/*
			 * Yes, we can.  Request that we do so.
			 */
			handle->bpf_codegen_flags |= BPF_SPECIAL_VLAN_HANDLING;
		}
	}
#endif /* defined(SO_BPF_EXTENSIONS) && defined(SKF_AD_VLAN_TAG_PRESENT) */
```

vs `pcap-npf.c`:
```c
#if defined(HAVE_PACKET_GET_INFO) && defined(NPF_GETINFO_BPFEXT) && defined(SKF_AD_VLAN_TAG_PRESENT)

	/* Can we generate special code for VLAN checks? */
	oid_data_arg->Oid = NPF_GETINFO_BPFEXT;
	oid_data_arg->Length = sizeof(ULONG);
	if (PacketGetInfo(pw->adapter, oid_data_arg)) {
		if (*((ULONG *)oid_data_arg->Data) >= SKF_AD_VLAN_TAG_PRESENT) {
			/* Yes, we can.  Request that we do so. */
			p->bpf_codegen_flags |= BPF_SPECIAL_VLAN_HANDLING;
		}
	}
	else {
		pcapint_fmt_errmsg_for_win32_err(p->errbuf, PCAP_ERRBUF_SIZE,
		    GetLastError(), "Error calling PacketGetInfo");
	}
#endif /* HAVE_PACKET_GET_INFO */
```

This should work fine with the code in `bpf_image.c`. The implementation of `pcapint_filter_with_aux_data()` in `bpf_filter.c` won't work the same as in Linux, since Npcap doesn't offer the auxiliary data for user-mode filtering (apart from inserting the VLAN header into the packet data), but the only place that's invoked without an explicit `NULL` as the `aux_data` parameter is in `pcap-linux.c`, so it's irrelevant.

I'd like to consider supporting other `SKF_AD_*` metadata if it's useful, but what I saw in `gencode.c` was already gated with `#if defined(__linux__)`, so I determined it was fine to reorder the constants as long as code could test whether each one is supported. If needed, we could have a place in the handle to stash the return value of `SO_BPF_EXTENSIONS` or `NPF_GETINFO_BPFEXT`, since it has the same semantic meaning: anything less than that value is supported.

infrastation:
I missed the fact Npcap defines it as:
```c
#define SKF_AD_VLAN_TAG 0
#define SKF_AD_VLAN_TAG_PRESENT 4
#define SKF_AD_MAX 4
```
Linux defines it as:
```c
#define SKF_AD_PROTOCOL 0
#define SKF_AD_PKTTYPE  4
#define SKF_AD_IFINDEX  8
#define SKF_AD_NLATTR   12
#define SKF_AD_NLATTR_NEST      16
#define SKF_AD_MARK     20
#define SKF_AD_QUEUE    24
#define SKF_AD_HATYPE   28
#define SKF_AD_RXHASH   32
#define SKF_AD_CPU      36
#define SKF_AD_ALU_XOR_X        40
#define SKF_AD_VLAN_TAG 44
#define SKF_AD_VLAN_TAG_PRESENT 48
#define SKF_AD_PAY_OFFSET       52
#define SKF_AD_RANDOM   56
#define SKF_AD_VLAN_TPID        60
#define SKF_AD_MAX      64
```
So in Linux `SKF_AD_VLAN_TAG_PRESENT` implies 12 more data elements, but in Npcap it implies one, and the offsets are different.

This creates such a space for things to go wrong both in the code and in the user interpretation. For instance, for #863 the notation `ldh [skf_ad_vlanp]` or `ldh skf[ad_vlanp]` would mean different bytecode depending on which BPF implementation is supposed to execute it.

It may be not too late to avoid this namespace collision and to use something such as `NPCAP_AD_xxxxx`, which in #863 would unambiguously mean `ldh [npcap_ad_vlanp]`, or something along these lines.

infrastation:
Thank you for the comments. Perhaps a better way to express my concern is that we very much seem to be repeating the same type of mistake as the developers that in the past reused the same DLT name or value for completely different data on different OSes.

infrastation:
Yes, one of the things I had in mind are cBPF savefiles, as a means to capture the inputs and the outputs of a supposedly pure function, in this case `pcap_compile()`, and to verify/reproduce it at a different host and/or at a different time. To that end, the file format lacks yet another flag bit, "uses Linux BPF extensions", and since now yet another, "uses Npcap BPF extensions". I didn't add that because I was still thinking where `SKF_AD_MAX` would fit best.

Another thing is keeping human-readable BPF dumps and the corresponding numeric instructions an unambiguous two-way mapping (less the host endianness). E.g. if a user writes `ldb [vlanp]` loads an incorrect value, a developer knows this was compiled on Linux and is always the same as `{ 0x30, 0, 0, 0xfffff030 }`, etc.

dmiller-nmap:
Since libpcap is the primary consumer of the Packet.dll API and the only one making use of these special offset definitions, it is probably fine to change the name prefix from `SKF_AD_`. However, I was under the impression that the translation from filter expression to BPF bytecode (and vice versa) was not guaranteed to be identical in all circumstances, since different contexts result in different bytecode. For example, `ip[0]` results in different addresses loaded depending on data link type, and 'ip broadcast' results in different bytecode depending on the mask passed to `pcap_compile()`. I'll have to learn more about cBPF savefiles to be able to have that conversation.

guyharris:
> However, I was under the impression that the translation from filter expression to BPF bytecode (and vice versa) was not guaranteed to be identical in all circumstances

It's definitely link-layer-type-dependent.

It's *also* OS-dependent, as, for example, not all OS's networking stack extract VLAN tags and put them in per-packet metadata, so not all OSes require special handling of VLAN packets in filter code.

mcr:
Some of this work seems like it ought to involve the IETF BPF WG.
(It's really the eBPF WG)
Note everyone knows about that effort.

infrastation:
As far as I see it, the mnemonic offset names are irrelevant in the sense of a cbpf-savefile(5) because it records the bytecode as numbers. To interpret/validate/use these numbers correctly it is necessary to record which dialect is in use. `bpf_compile()` and `bpf_setfilter()` have this as an implicit discrete input via the C preprocessor, cbpf-savefile(5) tries to capture that in the flags. Whether `bpf_dump()` uses mnemonic names for that dialect or not in its output is a matter of convenience. So in this scope instead of the namespace string it would be a better fit to have a discrete indicator such as a flag bit or the likes.

In the sense of the filter expression string the namespace is indeed a relevant factor, at the very least to be able to tell which names are valid. One potential approach could be prefixing the namespace (e.g. `link[skf_ad_vlan]` and `link[npcap_ad_vlan]`). In any case, for consistency it would be best if `bpf_dump()` used the same names for its output.

Some research still remains to be done in this direction and not all of these possibilities will necessarily end up implemented as discussed or at all, but in any case the best way not to paint yourself into a corner seems to be avoiding the reuse of Linux SKF names on Windows.

PR #1476, opened 2025-02-25
DPDK: don't enable it by default.

guyharris:
The code is immensely hard to keep compiling for every random API change the DPDK folks make. If somebody wants DPDK support, it's on them to make it work.


infrastation:
On the other hand, if you let it try and fail to compile, someone else may eventually get motivated to fix it.

PR #1478, opened 2025-02-26
tPrefer NPCAP_AD prefix for Npcap over SKF_AD prefix

dmiller-nmap:
As discussed in #1473, to avoid namespace collision and confusion, use separate namespaces for Npcap's and Linux's BPF extensions. In gencode.c, known-good tested code continues to use SKF_AD_ for both, but a comment and conditional definition of those constants should make things clearer.

### DIFF
--- a/bpf_filter.c
+++ b/bpf_filter.c
@@ -82,7 +82,8 @@ enum {
  *
  * Thanks to Ani Sinha <ani@arista.com> for providing initial implementation
  */
-#if defined(SKF_AD_VLAN_TAG_PRESENT)
+/* Npcap does not provide auxiliary data */
+#if defined(SKF_AD_VLAN_TAG_PRESENT) && !defined(NPCAP_AD_VLAN_TAG_PRESENT)
 u_int
 pcapint_filter_with_aux_data(const struct bpf_insn *pc, const u_char *p,
     u_int wirelen, u_int buflen, const struct pcap_bpf_aux_data *aux_data)
@@ -142,7 +143,7 @@ pcapint_filter_with_aux_data(const struct bpf_insn *pc, const u_char *p,
 DIAG_OFF_DEFAULT_ONLY_SWITCH
 			switch (pc->k) {
 
-#if defined(SKF_AD_VLAN_TAG_PRESENT)
+#if defined(SKF_AD_VLAN_TAG_PRESENT) && !defined(NPCAP_AD_VLAN_TAG_PRESENT)
 			case SKF_AD_OFF + SKF_AD_VLAN_TAG:
 				if (!aux_data)
 					return 0;
--- a/bpf_image.c
+++ b/bpf_image.c
@@ -48,11 +48,18 @@
 #include "os-proto.h"
 #endif
 
-#ifdef SKF_AD_OFF
+#if defined(SKF_AD_OFF) || defined(NPCAP_AD_OFF)
 /*
- * Symbolic names for offsets that refer to the special Linux BPF locations.
+ * Symbolic names for special offsets that refer to platform-specific BPF extensions.
  */
-static const char *offsets[SKF_AD_MAX] = {
+#ifdef (NPCAP_AD_MAX)
+#define BPFEXT_AD_OFF NPCAP_AD_OFF
+#define BPFEXT_AD_MAX NPCAP_AD_MAX
+#else
+#define BPFEXT_AD_OFF SKF_AD_OFF
+#define BPFEXT_AD_MAX SKF_AD_MAX
+#endif
+static const char *offsets[BPFEXT_AD_MAX] = {
 #ifdef SKF_AD_PROTOCOL
 	[SKF_AD_PROTOCOL] = "proto",
 #endif
@@ -86,10 +93,14 @@ static const char *offsets[SKF_AD_MAX] = {
 #ifdef SKF_AD_ALU_XOR_X
 	[SKF_AD_ALU_XOR_X] = "xor_x",
 #endif
-#ifdef SKF_AD_VLAN_TAG
+#ifdef NPCAP_AD_VLAN_TAG
+	[NPCAP_AD_VLAN_TAG] = "vlan_tci",
+#elif defined(SKF_AD_VLAN_TAG)
 	[SKF_AD_VLAN_TAG] = "vlan_tci",
 #endif
-#ifdef SKF_AD_VLAN_TAG_PRESENT
+#ifdef NPCAP_AD_VLAN_TAG_PRESENT
+	[NPCAP_AD_VLAN_TAG_PRESENT] = "vlanp",
+#elif defined(SKF_AD_VLAN_TAG_PRESENT)
 	[SKF_AD_VLAN_TAG_PRESENT] = "vlanp",
 #endif
 #ifdef SKF_AD_PAY_OFFSET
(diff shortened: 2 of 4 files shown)


bpf_filter.c
```diff
@@ -82,7 +82,8 @@ enum {
  *
  * Thanks to Ani Sinha <ani@arista.com> for providing initial implementation
  */
-#if defined(SKF_AD_VLAN_TAG_PRESENT)
+/* Npcap does not provide auxiliary data */
+#if defined(SKF_AD_VLAN_TAG_PRESENT) && !defined(NPCAP_AD_VLAN_TAG_PRESENT)
```
infrastation, line 86: Is it true in this context that so long as `SKF_AD_VLAN_TAG_PRESENT` is defined, `NPCAP_AD_VLAN_TAG_PRESENT` is never defined?

gencode.c
```diff
@@ -9418,7 +9425,7 @@ gen_vlan(compiler_state_t *cstate, bpf_u_int32 vlan_num, int has_vlan_tag)
 		 *
 		 * This requires special treatment.
 		 */
-#if defined(SKF_AD_VLAN_TAG_PRESENT)
+#if defined(SKF_AD_VLAN_TAG_PRESENT) || defined(NPCAP_AD_VLAN_TAG_PRESENT)
```
infrastation, line 9428: Is it true in this context that so long as `SKF_AD_VLAN_TAG_PRESENT` is defined, it does not matter if `NPCAP_AD_VLAN_TAG_PRESENT` is defined, and so long as `SKF_AD_VLAN_TAG_PRESENT` is not defined, `NPCAP_AD_VLAN_TAG_PRESENT` is certainly not defined? In other words, changing this line seems to be a no-op.

bpf_image.c
```diff
@@ -86,10 +93,14 @@ static const char *offsets[SKF_AD_MAX] = {
 #ifdef SKF_AD_ALU_XOR_X
 	[SKF_AD_ALU_XOR_X] = "xor_x",
 #endif
-#ifdef SKF_AD_VLAN_TAG
+#ifdef NPCAP_AD_VLAN_TAG
+	[NPCAP_AD_VLAN_TAG] = "vlan_tci",
+#elif defined(SKF_AD_VLAN_TAG)
```
infrastation, line 98: If at most one of {`NPCAP_AD_VLAN_TAG`, `SKF_AD_VLAN_TAG`} is defined in any one build, using `#elif defined(NPCAP_AD_VLAN_TAG)` would keep the diff cleaner.

infrastation:
Thank you, this generally makes sense, there is space for potential minor clean-ups as commented.

bpf_filter.c
dmiller-nmap: As it currently stands, Npcap's `npcap-bpf.h` conditionally defines the same `SKF_AD_*` constants it did before (https://github.com/nmap/npcap/blob/a4e1690925da1c0086400a1dab7688567818ec5f/Common/npcap-bpf.h#L230-L238):

```c
/* Npcap SDK 1.15 defined these names instead, so they are preserved here for
 * compatibility and to simplify integration with existing Linux-targeted code.
 * However, it is preferred to use the NPCAP_AD_* prefix in new code. */
#ifndef SKF_AD_OFF
#define SKF_AD_OFF              NPCAP_AD_OFF
#define SKF_AD_VLAN_TAG         NPCAP_AD_VLAN_TAG
#define SKF_AD_VLAN_TAG_PRESENT NPCAP_AD_VLAN_TAG_PRESENT
#define SKF_AD_MAX              NPCAP_AD_MAX
#endif
```

So in a Npcap build, both constants are defined. We can still change this if necessary.

gencode.c
dmiller-nmap: I intended this to draw attention to the fact that this block is active in a Npcap build as well, clarifying intent of the preprocessor directive, especially in contrast to other places in this file where `SKF_AD_*` constants are instead guarded with `#if defined(__linux__)`.

bpf_image.c
dmiller-nmap: As mentioned above, both may be defined currently for a Npcap build. That is why the `NPCAP_AD_*` check must come first. However, we have not made a SDK release for that yet, so it is still open to change.

infrastation: Thank you, let me check again.

bpf_filter.c
infrastation: Based on what you write, as far as I see, if there is a good way to remove `SKF_AD_VLAN_TAG_PRESENT` from Npcap scope now, it would be best to do that, then this specific change will not be required and things will be a bit simpler; if it is too late for that, this comment needs to explain this context a bit better:
```c
/*
 * Npcap does not provide auxiliary data.  However, it can define
 * SKF_AD_VLAN_TAG_PRESENT if it defines NPCAP_AD_VLAN_TAG_PRESENT.
 */
```

gencode.c
infrastation: In this context it seems much easier to follow if this specific change was just:
```patch
--- a/gencode.c
+++ b/gencode.c
@@ -8871,6 +8871,12 @@ gen_vlan(compiler_state_t *cstate, bpf_u_int32 vlan_num, int has_vlan_tag)
                 * Newer version of the Linux kernel pass around
                 * packets in which the VLAN tag has been removed
                 * from the packet data and put into metadata.
+                * (Specifically, the kernel started defining the constant
+                * in release 3.8.13 and started using it for packet
+                * filtering in release 3.15.8.)
+                *
+                * Npcap since version 1.81 supports the same feature, in
+                * which case the same constant has been defined by now.
                 *
                 * This requires special treatment.
                 */
```

bpf_image.c
```diff
@@ -107,16 +118,16 @@ static const char *offsets[SKF_AD_MAX] = {
 static void
 bpf_print_abs_load_operand(char *buf, size_t bufsize, const struct bpf_insn *p)
 {
-#ifdef SKF_AD_OFF
+#ifdef BPFEXT_AD_OFF
 	const char *sym;
 
 	/*
 	 * It's an absolute load.
 	 * Is the offset a special Linux offset that we know about?
```
infrastation, line 126: If this block is no longer Linux-specific, the comment needs to mention either both possibilities, or none.

gencode.c
```diff
@@ -9222,6 +9222,13 @@ gen_vlan_no_bpf_extensions(compiler_state_t *cstate, bpf_u_int32 vlan_num,
 	return b0;
 }
 
+/* Npcap defines these constants in a way that is compatible with Linux, as
```
infrastation, line 9225: "...compatible with Linux at source code level"

bpf_image.c
infrastation: Alright, this looks correct then.

infrastation:
Thank you for waiting. I understand the proposed change better now, please address as many of today's comments as possible and rebase on the current master branch.

infrastation:
On a related note, it seems in commit aacb928 the new code in `pcap-npf.c` would be easier to follow if the two new variables were a part of the new conditional code block that uses them.

dmiller-nmap:
Thanks for the feedback. I'll look at this.

dmiller-nmap:
> On a related note, it seems in commit [aacb928](https://github.com/the-tcpdump-group/libpcap/commit/aacb92863d313aa40ab041a62a7fcc744d48100b) the new code in `pcap-npf.c` would be easier to follow if the two new variables were a part of the new conditional code block that uses them.

I agree that would be nice, but I wasn't sure if I could expect a C99-compliant compiler.

infrastation:
libpcap started requiring C99 in release 1.10.0 a few years ago.

PR #1479, opened 2025-03-01
Add support for filtering VNTag frames

### LINKED ISSUE #770: vn-tag support required for bpf

stefan-baranoff:
Add a "vntag" keyword similar to the existing "vlan" and "mpls" keywords to filter on VNTag frames. VNTag was proposed as 802.1Qbh but has been replaced in the standards by 802.1Qbr although some vendors are still using VNTag.

Because VNTag is always followed by VLAN and the following VLAN has the next EtherType, both the VNTag and VLAN have to be present to match and both have to be skipped in processing. So, this commit also adds an optional `vlan [vlan_id]` after the `vntag` keyword to easily filter on the VLAN ID following the VNTag.

Fixes #770


infrastation:
Please cover the new code with tests as much as you can (see "accept", "apply" and "reject" tests types in `TESTrun`) and add a test packet capture with the encodings that exercise the new code paths.

stefan-baranoff:
Understood -- I opened the PR a little pre-maturely anyway; my branch name was off (`nvtag` vs `vntag`) and I realized I didn't have tests so closed right as you were leaving the comment. I'll re-open a new PR with tests!

Issue #1247, opened 2023-12-06
Linux SocketCAN byte order problems


CWNE88:
I’d like to set capture filter for CAN ID.
Either specific CAN ID variable, or being able to specify which bytes of the payload to match.
Similar to this for selecting ID 5c2:

   ether[2:2] == 0x05c2

That “ether” filter will not work, because the device type is SocketCAN.
However, that above filter does actually work if reading from a captured file.
I would like to be able to filter at the capture stage.

From current code:

case DLT_CAN_SOCKETCAN:
		bpf_error(cstate, "CAN link-layer type filtering not implemented");

CAN only has an ID, length, and payload (of up to 8 bytes).
Even if we can only filter with specific bytes, that would be enough.

For instance, something like this maybe:

   can[2:2] == 0x05c2

At least that way, we could define what bytes in the packet to filter for the CAN type.

Thanks
Paul.

infrastation:
`ether[2:2]` is the same as `link[2:2]`, although it is not immediately obvious why offline filtering works and online filtering does not. What would be the cheapest piece of hardware to test these code paths and to reproduce the problem?

CWNE88:
I did originally try link[2:2] as well, but nothing got captured.
The cheap device I'm using is the isolated (although you probably don't need isolation for a desk test) version of this:
https://www.aliexpress.com/item/1005006032351087.html

CWNE88:
[can-request.pcapng.gz](https://github.com/the-tcpdump-group/libpcap/files/13582363/can-request.pcapng.gz)

CWNE88:
Okay, it seems the can ID must be a value that fits within 7 bits, so that explains the Wireshark information.

To set this up, you don't need any other equipment, just the adapter.
Install can-utils to get the cansend command.
When you connect the device, it will come up as can0 (down)

Bring it up with:
ip link set can0 type can bitrate 500000
ip link set up dev can0

Send a message with:
cansend can0 123#11223344
where 123 is the ID and 11223344 is the payload (maximum 8 bytes for the payload)

You can run the capture on a different CLI to test filtering.

infrastation:
Thank you for the comments, I have ordered a very similar USB dongle and hopefully will be able to reproduce this problem soonish. Meanwhile other developers may want to make additional input.

infrastation:
This is how my USB CAN device identifies itself:
```
ID 1d50:606f OpenMoko, Inc. Geschwister Schneider CAN adapter
[...]

PR #1483, opened 2025-03-10
Filter Linux SocketCAN frames in userland if necessary.

infrastation:
To the best of my understanding, this is a correct workaround for #1247. It works in my tests with CAN CC. Tests for this code path do not exist, at least not yet, so please proof-read if you can.


infrastation:
Addressed the compiler warnings and for now removed the other, non-essential refactoring change, which requires a bit more attention to audit the handling of `cooked`.

PR #1484, opened 2025-03-13
Autoconf: Use AC_SYS_YEAR2038_RECOMMENDED when possible

fxlb:
On Linux, if Autoconf version >= 2.72 and GNU C Library version >= 2.34, 
s/AC_SYS_LARGEFILE/AC_SYS_YEAR2038_RECOMMENDED/ to ensure time_t is Y2038-safe.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -23,6 +23,27 @@ fi
 
 echo "$AUTORECONF identification: $AUTORECONFVERSION"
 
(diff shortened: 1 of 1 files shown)


infrastation:
Perhaps this should be specfic to Linux (and glibc if possible) to eliminate surprises in other cases.

infrastation:
That's a good point.

Issue #125, opened 2013-04-15
pcap-filter man, better libpcap filter expressions intro


guyharris:
Converted from [SourceForge issue 2823520](http://sourceforge.net/support/tracker.php?aid=2823520), submitted by doru001

For clarity and ease of learning reasons, please change the filter expressions presentation paragraphs to include: 
- rename all qualifiers as modifiers, or otherwise make the manual compatible with error messages. For example, in the expression 'tcpdump tcp host 2', the man says that tcp is a qualifier, while the error message says that tcp is a modifier. Also, the man does not justify the use of both terms, modifier and qualifier, or at least I did not find the justification. 
- the mention that a primitive only operates within one protocol level header, and that one primitive can not completely define one protocol level header, that is, one primitive can not contain all conditions applicable to one protocol level header. 
- the right order of the three different kinds of modifiers/qualifiers, formally: [proto][dir][type][id]. Now they are presented in this order: type, dir, proto. 
- emphasize the statement that [proto][dir][type][id] is only a general structure, not a definition of primitives, and that not any primitive accepts all terms in [proto][dir][type][id].
- emphasize the statement that the list of allowable primitives actually defines primitives; primitives are not defined by [proto][dir][type][id]. 
- complete the list of primitives with 'ip src host [id]' and others, such that it becomes comprehensive. 
- explain when is 'ip' an alias for 'ether proto \ip' and when is it a modifier. Same for other aliases.

infrastation:
The original issue was opened on 2009-07-18. The only immediately actionable work item I can tell here is to make the error messages consistent with the man page.

infrastation:
On the note of Linux BPF placed "outside of" iptables, on FreeBSD this used to be exactly the opposite: tcpdump never sees incoming packets rejected by `ipfw`.

infrastation:
After a number of changes this problem space has been addressed as follows:
* **implemented**: _rename all qualifiers as modifiers, or otherwise make the manual compatible with error messages_ — the man page now uses "qualifier" in all instances (it used to be prevalent, but not the only) and the error messages now use "qualifier" to match the man page.
* **rejected**: _the mention that a primitive only operates within one protocol level header, and that one primitive can not completely define one protocol level header, that is, one primitive can not contain all conditions applicable to one protocol level header_ — in fact, some primitives do not access any protocol headers (`greater`) and some access more than one (`gateway` or even `ip host`); also it is obvious that one primitive does not test all possible protocol fields.
[...]

PR #1490, opened 2025-03-21
Make pcap_compile() error messages more uniform and consistent.

infrastation:
The original motivation for these changes was to address #125 as far as reasonably practicable, but there is assorted other refactoring as well.


infrastation:
This revision has redundant checks in `assert_atm()` and `assert_ss7()` and no longer includes the commits that have no effect on the error messages. Those are going to be a separate pull request later together with some other changes when I have time to complete that lot of work properly (i.e. with tests).

This pull request is going to be merged soon. To the best of my understanding, its only user-visible change is the error messages generated by `pcap_compile()`.

PR #1493, opened 2025-03-27
helper functions and other clean-ups

infrastation:
This series of changes deduplicates some C code in `gencode.c` without affecting the resulting filter program.


infrastation:
This needs a few minor fixups to look better.

infrastation:
This revision seems right.

infrastation:
I have another WIP change along the same lines, but it is not ready yet.

Issue #1375, opened 2024-10-22
do NOT close fp in pcap_close call when the fp obtained via pcap_fopen_offline


crackevil:
- [ ] This is not a security issue (See first line).

libpcap version: 1.10.4 on windows.

calling pcap_close will close the fp which opened before pcap_fopen_offline.
the pcap_dump_close routine has the same issue.

guyharris:
> calling pcap_close will close the fp which opened before pcap_fopen_offline.

Yes, it will, and that behavior dates back to libpcap 0.9, when `pcap_fopen_offline()` was introduced.

Changing it would break the API and ABI contracts that libpcap offers to its callers; that could, for example, leak the `FILE *` and its FD in programs that expect `pcap_close()` to close the `FILE *`.

If there is code that uses libpcap and nees to use the `FILE *` after `pcap_close()` is called, we would have to introduce another routine that will flag the `pcap_t` as "don't close the `FILE *`".

(When a `FILE *` opened with `fdopen()` is closed, [the FD that was provided to `fdopen()` is closed](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fclose.html) - "The _fclose_() function shall perform the equivalent of a [_close_()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/close.html) on the file descriptor that is associated with the stream pointed to by _stream_." - which is an analogous behavior.)

guyharris:
I will update the `pcap_close()` man page to explicitly state that the `FILE *` will be closed if the `pcap_t` corresponds to a savefile.

crackevil:
thanks~
indeed, this behavior will impact old code. it's fine if the behavior keep consistent.

guyharris:
I've updated the documentation.

I've also changed `pcap_dump_close()` to flush, but not close, the `FILE  *` if it's `stdout`, just as `pcap_close()` doesn't close a savefile's `FILE *` if it's `stdin`.

crackevil:
maybe new routines like **pcap_close_without_fp** will be more helpful

infrastation:
One potential way to do this would be using a new function generalised from #1374 such as `pcap_set_shutdown_mode()` or even something such as `pcap_set_flags(p, pcap_get_flags(p) | PCAP_CLOSE_LEAVE_FILES_OPEN)`.

guyharris:
> maybe new routines like **pcap_close_without_fp** will be more helpful

Where is this useful?

[...]

Issue #1374, opened 2024-10-20
Libpcap API for stopping delivering new packets


garrymar:
There are cases when it is desirable to fully drain the OS-level input packet buffer before terminating programs using libpcap like tcpdump, Wireshark, and others. For example, it might be necessary when large input buffers are in use to accommodate faster input packet streams with slower storage devices.

Currently, when programs like tcpdump terminate, they cannot drain the input buffer to a dump file as there is no way to instruct libpcap to stop delivering new packets to the capturing program first. As a result, there is no way to prevent loss of packets stored in large input buffers when the programs stop. If libpcap provides such an API, consuming program will be able to ask for no more new packets and safely drain the input buffer before quitting.

More details on this use case can be found in the following mailing list discussion:

https://seclists.org/tcpdump/2024/q4/10

Thank you.

Regards,
Garri

guyharris:
My first question to the audience:

What should the routine be called? It could be `pcap_stop_capture()`, but that name seems a bit vague and its purpose could be misinformed.

guyharris:
Note that `pcap_stop_capture()` would cause all packets received by the machine after it's called not to be provided the the program, so, unless you only call it when you know you don't want any more packets, you're going to end up with... all subsequent packets being dropped by the capturing program.

GD1200:
> My first question to the audience:
> 
> What should the routine be called? It could be `pcap_stop_capture()`, but that name seems a bit vague and its purpose could be misinformed.

@guyharris, what about `pcap_deprive_buffer()` or `pcap_deprive_input_buffer()`? I think it would be in line with what will happen with the buffer.

I do not know if the input buffer is also used for offline reading, but if it is, it can help avoid focus on the "capture" part as it is connoted with live reading.

My 2 cents.

Garri

guyharris:
> I think it would be in line with what will happen with the buffer.

I.e., the input buffer will be deprived of any further packets, as in "[further packets will be withheld from it](https://www.merriam-webster.com/dictionary/deprive)"? I guess that works, although 1) it's a bit of  a complicated explanation of the deprivation and 2) people might not be thinking of the input buffer, they may be thinking of the enitre `pcap_t`.

> I do not know if the input buffer is also used for offline reading

[...]

Issue #1050, opened 2021-09-15
Can not generate BPF extensions on dead handle.


shemminger:
Is there a way to generate BPF code to use extensions around vlan's when compiling on a non-live pcap handle?

I am adding support for pcap filters to DPDK.
The code works by doing:

   	pcap = pcap_open_dead(DLT_EN10MB, snaplen);
	pcap_compile(pcap, &bf, filter_str,
			 1, PCAP_NETMASK_UNKNOWN);

And then the resulting BPF code is converted to eBPF interpreted by DPDK.

This works for most cases, but the DPDK has options to strip VLAN's like Linux kernel. 
The converter code could handle that BUT the pcap_compile won't generate required BPF extensions on a handle created in this way.

Don't want to create a live handle because don't want to be capturingpackets in the kernel environment.

Maybe an extension like:
       pcap_compile_with_bpf_extensions(...)

mcr:
Yes, that sounds reasonable, but we need to avoid more _with_foo_bar() stuff, and do something that is more easily turned into an ABI.
The short is that there isn't a way.

guyharris:
> I am adding support for pcap filters to DPDK.

Do you mean that pcap filters don't work when you capture on a DPDK device *using libpcap* (i.e., opening a DPDK device, and capturing on it, using libpcap APIs), and you're fixing that, or do you mean you're adding support for pcap filter expressions when *not* using libpcap to do DPDK captures?

If it's the latter, then that seems to indicate that what's needed is a separate library to compile pcap filters into various machine languages.

> Don't want to create a live handle because don't want to be capturingpackets in the kernel environment.

If you capture on a DPDK device using libpcap, you're not, as far as I know, capturing in the kernel environment.

shemminger:
To be more clear about what this is trying to do:

- The network processing application (for example OpenVswitch) is a DPDK primary process which interacts with usermode network drivers.
- Application in DPDK (dpdk-dumpcap) runs as a secondary process (communicating over IPC/shared memory to primary):
- It compiles a filter using pcap_open_dead/pcap_compile and gets BPF result.
- It passes the BPF program off to function rte_bpf_convert() in DPDK that converts BPF to eBPF
- Finally the application  then communicates the request  then makes a request to the primary process
- Resulting eBPF is handled by DPDK as part of the capture process.

The key point is at the compile step.  Since the application is using pcap_open_dead, the code generation (gencode.c) will not generate BPF extensions to check for VLAN.

[...]

PR #1495, opened 2025-03-27
Add pcap_breakloop_flush API

benzea:
I was a it annoyed because tcpdump on Linux can loose a second worth of packets (if there is little traffic and the mmap_v3 API is used). I had this code already, before reading that `pcap_breakloop` actually even documents this behaviour in a way.

I have *not* tested this properly. Even the first commit can probably be improved to clear break_loop for all errors.

So for now, more of a request for comments. Are you open to such an API, or do you think it is preferable to stick with the current limitations and let tcpdump work around the problem?

### DIFF
--- a/dlpisubs.c
+++ b/dlpisubs.c
@@ -166,8 +166,7 @@ pcap_process_pkts(pcap_t *p, pcap_handler callback, u_char *user,
 		 */
 		if (p->break_loop) {
 			if (n == 0) {
-				p->break_loop = 0;
-				return (-2);
+				return PCAP_ERROR_BREAK;
 			} else {
 				p->bp = bufp;
 				p->cc = ep - bufp;
--- a/pcap-bpf.c
+++ b/pcap-bpf.c
@@ -1190,15 +1190,9 @@ pcap_read_bpf(pcap_t *p, int cnt, pcap_handler callback, u_char *user)
 	/*
 	 * Has "pcap_breakloop()" been called?
 	 */
-	if (p->break_loop) {
-		/*
-		 * Yes - clear the flag that indicates that it
-		 * has, and return PCAP_ERROR_BREAK to indicate
-		 * that we were told to break out of the loop.
-		 */
-		p->break_loop = 0;
+	if (p->break_loop)
 		return (PCAP_ERROR_BREAK);
-	}
+
 	cc = p->cc;
 	if (p->cc == 0) {
 		/*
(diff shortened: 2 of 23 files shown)


infrastation:
It would help to state the problem in more detail, e.g. to give the steps to reproduce.

benzea:
So, the problem here are automated tests that inspect packets after a certain scenario was run.

This could for example be:
* Start tcpdump on a monitor interface (writing to a file)
* Ask wpa_supplicant to connect to an AP
* Wait until wpa_supplicant reports it is connected
* Stop tcpdump by sending a SIGTERM
* Inspect whether the captured association and 4-way handshake is well formed

The simple work-around is to add a one or two second delay, but if you forget that you will likely get random failures. But, I feel like such a workaround should not be needed as we are certain that the packet has been received by the kernel already. It is purely missing because tcpdump/libpcap is not flushing them out to the file.

To be more exact, part of the problem is that tcpdump sets a 1000ms timeout. When TPACKET3 is used, then libpcap configures the kernel with this timeout (`tp_retire_blk_tov`). And that results in the kernel waiting for up to one second before notifying userspace (if the next block is not fully filled with packets). When tcpdump receives the SIGTERM, it simply calls `pcap_breakloop()` and libpcap never processes these packets even though they must be available already.

infrastation:
Thank you. Does this look the same as the problem in #1374?

benzea:
Yes, same problem, just worded differently. The proposal from the [(currently) last comment ](https://github.com/the-tcpdump-group/libpcap/issues/1374#issuecomment-2427819179) is actually quite similar to the one used in my patch.

infrastation:
In this case a better solution may be introducing a more generic function for setting flags/options on the `pcap_t` handle (see the middle of #1375 and the end of #1050).

benzea:
> In this case a better solution may be introducing a more generic function for setting flags/options on the pcap_t handle …
To me this feels like this is a property of the "break"/"abort" operation rather than the `pcap_t`. So setting it separately (and in advanced due to signal-safety) seems a bit weird to me. It just isn't as obvious anymore what the `pcap_breakloop` call will do.

Maybe the internal API might be more interesting to talk about for a start? My approach was to pass a flag to `breakloop_op` and also store it in the `break_loop`. One the one hand, that should give the backend the information about what the user intended here. But, with the input stream open, it is hard to define where exactly it should stop.

The idea of closing the input seem nice. I imagine that in the case of the MMAP interface on Linux, this would work perfectly. You close the stream and then read all packets that the kernel has previously prepared in the buffer memory.

cicku:
Do we have any plan to make this happen in 1.11.x?

infrastation:
Not immediately. Right now bug fixes have priority over new features.

PR #1498, opened 2025-03-31
assorted refactoring and all-zero tetras

infrastation:
For some time I've been trying to address a number of imperfections, this is a series of changes that look ready. The first part reorganizes quite a bit of C code in `gencode.c` with the intent to make it easier to follow, but not to change any filter program outputs. The second part does not change much code, but it does reduce the unoptimized filter programs for IPv4 and IPv6 netmasks.

Due to the amount of changes this pull request is going to stay open for at least a week before merging.

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -5236,6 +5236,12 @@ gen_host6(compiler_state_t *cstate, struct in6_addr *addr,
 	case Q_DEFAULT:
 	case Q_IPV6:
 		b0 = gen_linktype(cstate, ETHERTYPE_IPV6);
+		// Same as the Q_IP case in gen_host().
+		if (
+			! memcmp(mask, &in6addr_any, sizeof(struct in6_addr)) &&
+			! memcmp(addr, &in6addr_any, sizeof(struct in6_addr))
+		)
+			return b0;
 		b1 = gen_hostop6(cstate, addr, mask, dir, 8, 24);
 		gen_and(b0, b1);
 		return b1;
--- a/gencode.c
+++ b/gencode.c
@@ -5172,16 +5172,16 @@ gen_mpls_linktype(compiler_state_t *cstate, bpf_u_int32 ll_proto)
 	case ETHERTYPE_IPV6:
 		/* match the bottom-of-stack bit */
 		b0 = gen_mcmp(cstate, OR_LINKPL, (u_int)-2, BPF_B, 0x01, 0x01);
-		/* match the IPv4 version number */
+		/* match the IPv6 version number */
 		b1 = gen_mcmp(cstate, OR_LINKPL, 0, BPF_B, 0x60, 0xf0);
 		gen_and(b0, b1);
 		return b1;
 
- default:
-	 /* FIXME add other L3 proto IDs */
-	 bpf_error(cstate, "unsupported protocol over mpls");
-	 /*NOTREACHED*/
- }
+	default:
+		/* FIXME add other L3 proto IDs */
+		bpf_error(cstate, "unsupported protocol over mpls");
+		/*NOTREACHED*/
+	}
 }
 
 static struct block *
--- a/gencode.c
+++ b/gencode.c
@@ -679,6 +679,8 @@ static struct block *gen_hostop6(compiler_state_t *, struct in6_addr *,
     struct in6_addr *, int, u_int, u_int);
 #endif
 static struct block *gen_ahostop(compiler_state_t *, const uint8_t, int);
+static struct block *gen_mac48hostop(compiler_state_t *, const u_char *,
+    const int, const u_int, const u_int);
 static struct block *gen_ehostop(compiler_state_t *, const u_char *, int);
 static struct block *gen_fhostop(compiler_state_t *, const u_char *, int);
 static struct block *gen_thostop(compiler_state_t *, const u_char *, int);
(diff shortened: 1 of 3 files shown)


mcr (approved):
This seems to consistently return a block from all gen_*(). Good.
It seems to replace b0=.. b1=.. gen_and() which seems to write to b1 with a gen_and() that returns a new block, which is actually always b1.  One advantage of the two intermediate variables is that they are much easier to inspect in a debugger, assuming the compiler hasn't optimized them out, while the inline gen_and(thing1,thing2) is harder.
This also would seem to preclude checking b0 or b1 for NULL.
The code is certainly not worse, and probably is better.
Can none of the gen_*() fail and return NULL?
But, please go ahead and merge.

guyharris:
> assuming the compiler hasn't optimized them out

Compilers do lots of stuff these days. Debuggers haven't always kept up, and sometimes they *can't*. Environments wherein compilers generate intermediate code that later gets translated to machine code, so that, when debugging, certain optimizations can be omitted (or done in such a way as to stash away stuff visible in the source code but not in the generated object code), might be a direction in which systems should move.  (Dunno whether IBM's System/38 and IBM i did that for debugging.)

But I haven't used debuggers in a while. Some of this is Apple's LLDB, the user interface to which is a vast unimprovement over that of GDB. (Who on earth thought that specifying command arguments with old-fashioned single-letter UNIX-command-style arguments was a good idea? Heck, even most UN\*Xes don't think that any more, as they provide `getopt_long()`.)

guyharris:
> all-zero tetras

Is that anything like a [neon tetra](https://en.wikipedia.org/wiki/Neon_tetra)?

(I'm guessing, from the new tests, that it refers to 0.0.0.0 as an IPv4 address.)

infrastation:
Thank you for checking. "All-zero tetra" is the best term I could find for `0x00000000` (the entirety of an IPv4 netmask or a quarter of an IPv6 netmask).

The and/or/not chaining commit is indeed more debatable than other. The temporary variables indeed could potentially be useful for debugging, given the debugger can follow the pointers and print the structures correctly. **If anyone actually uses this for debugging or sees any other good reason to exclude this commit, please make it clear.** Otherwise:
* Chaining the functions eliminates the need to remember which one of the arguments to `gen_and()` and `gen_or()` can be used after the function call (a few times I used the incorrect variable in a prototype and caused a segfault, which arguably became a smaller problem since the tests made it detectable much earlier for code that is covered with tests).
* If anybody needs an occasional temporary variable once in a long time to debug something particular, it would not be difficult to add it in their working copy. (I add temporary `printf()` lines instead.)
* Indentation and code style permitting, more complex tests could potentially be converted to trees of conditions where that would make it easier to follow the logic.

All three functions cannot be made static because the grammar uses them.

The check for the argument(s) being NULL is unrelated, but is a good idea (there is a few `gen_something()` functions that return a NULL to mean "always true" or "not applicable", perhaps these could be suffixed with `_or_null` to make it easier to see). In this case what would be the correct way to handle it? I've been thinking that the `abort()` calls in `gencode.c` ought to be a `bpf_error()` quoting the file name and line number — if a particular corner of the generator has a bug, there is no reason to terminate the program, instead a normal failure with a sufficiently detailed reason would be much more useful and the program could decide how to handle it according to the use case.

infrastation:
I have spent some time rebasing these changes on the current commits, and it turns out the and/or/not commit needs more work because it makes some filter programs non-deterministic. Specifically, this change
```patch
@@ -3604,21 +3592,18 @@ gen_linktype(compiler_state_t *cstate, bpf_u_int32 ll_proto)
 	case DLT_IEEE802_11_RADIO_AVS:
 	case DLT_IEEE802_11_RADIO:
 	case DLT_PPI:
 		/*
 		 * Check that we have a data frame.
-		 */
-		b0 = gen_mcmp(cstate, OR_LINKHDR, 0, BPF_B,
-			IEEE80211_FC0_TYPE_DATA,
-			IEEE80211_FC0_TYPE_MASK);
-
-		/*
+		 *
 		 * Now check for the specified link-layer type.
 		 */
-		b1 = gen_llc_linktype(cstate, ll_proto);
-		gen_and(b0, b1);
-		return b1;
+		return gen_and(
+		    gen_mcmp(cstate, OR_LINKHDR, 0, BPF_B,
+		        IEEE80211_FC0_TYPE_DATA,
+		        IEEE80211_FC0_TYPE_MASK),
+		    gen_llc_linktype(cstate, ll_proto));
 		/*NOTREACHED*/
 
 	case DLT_FDDI:
 		/*
 		 * XXX - check for LLC frames.
```
made the associated tests fail because the memory register numbers have changed:
```text
    accept_link_proto_ip_PRISM_HEADER_unopt_0: diff error
--- /tmp/libpcap_TESTrun_iOswpwwB/job001-expected.txt	2025-04-07 16:02:05.164265624 +0100
+++ /tmp/libpcap_TESTrun_iOswpwwB/job001-stdout.txt	2025-04-07 16:02:05.168265530 +0100
@@ -4,23 +4,23 @@
 (003) ld       [4]
 (004) ja       6
 (005) ld       #0x90
-(006) st       M[0]
+(006) st       M[1]
 (007) tax
 (008) txa
 (009) add      #24
-(010) st       M[1]
+(010) st       M[0]
 (011) ldb      [x + 0]
 (012) jset     #0x8             jt 13	jf 18
 (013) jset     #0x4             jt 18	jf 14
 (014) jset     #0x80            jt 15	jf 18
-(015) ld       M[1]
+(015) ld       M[0]
 (016) add      #2
-(017) st       M[1]
-(018) ldx      M[0]
+(017) st       M[0]
+(018) ldx      M[1]
 (019) ldb      [x + 0]
 (020) and      #0xc
 (021) jeq      #0x8             jt 22	jf 26
-(022) ldx      M[1]
+(022) ldx      M[0]
 (023) ldh      [x + 6]
 (024) jeq      #0x800           jt 25	jf 26
 (025) ret      #262144
```
This is because it is up to the compiler in which order to evaluate the arguments of a function call, and many functions cause direct and indirect invocations of `alloc_reg()` and `free_reg()`, which in this case mean a visible side effect. Assigning the results to temporary variables defines the sequence of these side effects.

I am going to remove this commit from this pull request for now. The `gen_not()` part is safe because there is only one argument, but it is small and the other changes need more attention.

infrastation:
This is the safe revision, it is going to be merged later today.

infrastation:
> I've been thinking that the abort() calls in gencode.c ought to be a bpf_error() quoting the file name and line number — if a particular corner of the generator has a bug, there is no reason to terminate the program, instead a normal failure with a sufficiently detailed reason would be much more useful and the program could decide how to handle it according to the use case.

For posterity, pull request #1579 implements approximately that.

PR #1500, opened 2025-04-04
protocol numbers and ports

infrastation:
This implements better validation of some `protocol` forms and cleaner unoptimized filter programs for `port` and `portrange` forms without a protocol qualifier.


infrastation:
Requires a few explicit type casts, as it turns out.

infrastation:
Some of the type casts were trivial, but the others required more attention, so I ended up rewriting these changes a few times. This revision has a bit better comments and removes some duplicate code as well.

infrastation:
This revision should be warning-free. Rebased on top of #1502. This is going to be merged later today.

PR #1504, opened 2025-04-07
Always enable IPv6 and the "gateway" primitive.

infrastation:
Please proof-read. This change is necessary to unblock the decommission of `INET6` and the making of all IPv6-related features always-on.

PR #1507, opened 2025-04-16
Consolidate MAC-48 address matching code.

infrastation:
This is the less straightforward change I mentioned in #1505 earlier.

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -6739,7 +6739,8 @@ gen_scode(compiler_state_t *cstate, const char *name, struct qual q)
 	int proto = q.proto;
 	int dir = q.dir;
 	int tproto;
-	u_char *eaddr;
+	u_char *eaddrp;
+	u_char eaddr[6];
 	bpf_u_int32 mask, addr;
 	struct addrinfo *res, *res0;
 	struct sockaddr_in *sin4;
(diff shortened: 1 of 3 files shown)


infrastation:
For this change I have prepared 5 more commits to get some details right, these may require minor clean-ups. If everything goes well, I am going to declare this lot ready for review next week.

infrastation:
The failure on `linux-armv7l` requires more attention:
```text
$ ./testprogs/filtertest 
Segmentation fault (core dumped)
```

infrastation:
That's not specific to the new TinyCC, please see [this message](https://lists.nongnu.org/archive/html/tinycc-devel/2025-04/msg00003.html). The next revision will not add `atexit()` to that test program. However, `pcap-linux.c` uses `atexit()` already, so perhaps the build systems ought to test whether the function actually works, and either fail the build early, or not define `HAVE_ATEXIT` and cause `pcap-linux.c` to disable the code that depends on `atexit()`. Perhaps this could be a separate bug report.

infrastation:
This is the next revision. As before, it does not change anything with regard to `WSAStartup()` and `WSACleanup()`, which are outside of my area of expertise. I guess the Windows-specific case is not becoming worse than it is now.

The "Plug memory leaks from pcap_ether_hostton()." commit is meant to be cherry-picked into the stable branch, hence it changes code that the next commit replaces.

Please proof-read if you can, I am going to merge this soon in order to unblock #1504.

Issue #1358, opened 2024-09-28
When using CMake to build, pkg-config finds libnl but libpcap does not compile with libnl support.


johnthacker:
- [X] This is not a security issue (See first line).

When using CMake instead of configure on Linux, the code to set `HAVE_LIBNL` and the libnl libraries does not work, at least in the path where `pkg-config` finds libnl. `HAVE_LIBNL` is not set in that path, and the include directory path is not set either. The result is linking against libnl, but not actually compiling the code into libpcap that would use libnl. The path that does not use pkg-config may not set the proper include directories if libnl is in a nonstandard locations (producing similar behavior to #172.)

PR #1510, opened 2025-04-26
Linux: Always request and wait for an ACK from nl80211

### LINKED ISSUE #1508: monitor-less 802.11 interfaces now fail with an SIOCGIFINDEX error message
This is a regression after pull request #1362. @johnthacker, please review if you can. libpcap since commit 0ac887b returns a cryptic error if the 802.11 interface does not support monitor mode. Steps to reproduce:
1. Take a Raspberry Pi with a built-in 802.11 interface and a recent RaspiOS flavour of Debian.
2. Confirm that it does not support monitor mode (applies to at least models 3B and 5):
   ```
   # iw phy phy0 interface add mon0 type monitor
   command failed: Operation not supported (-95)
   ```
3. Note how the packaged tcpdump/libpcap fail enabling monitor mode:
   ```
   # tcpdump --version
   tcpdump version 4.99.3
   libpcap version 1.10.3 (with TPACKET_V3)
   OpenSSL 3.0.15 3 Sep 2024
   # tcpdump -i wlan0 -I
   tcpdump: wlan0: That device doesn't support monitor mode
   ```
4. Install `libnl-genl-3-dev` and build tcpdump/libpcap from the current master branches.
5. Note how the snapshot fails:
   ```
   # tcpdump --version
   tcpdump version 5.0.0-PRE-GIT
   64-bit build, 64-bit time_t
   # tcpdump -i wlan0 -I
   tcpdump: SIOCGIFINDEX: No such device
   # opentest -i wlan0 -I
   opentest: wlan0: Generic error
   (SIOCGIFINDEX: No such device)
   ```
6. Optionally connect a USB 802.11 adapter that supports monitor mode and see that the happy path seems to work as expected.

The immediate reason why the message is different is because given two interfaces (one without monitor mode support and one with), the code paths begin in `pcap_activate_linux()` and remain identical until the "Now configure the monitor interface up." block in `enter_rfmon_mode()`, there the `ioctl(sock_fd, SIOCGIFFLAGS, &ifr)` call fails if `mon0` does not exist, in this case `pcapint_fmt_errmsg_for_errno()` sets the error message, then `del_mon_if()` invokes `iface_get_id()`, which attempts `ioctl(fd, SIOCGIFINDEX, &ifr)`, which fails and overwrites the error message with what the user is seeing.

The root cause seems to be elsewhere because the monitor-less code path should not have reached that block and the clean-up of non-existent monitor interface should not be necessary in the first place. It seems wrong that the `add_mon_if()` call just before reports success in both cases. In both cases `ret == 1` and `type == NL80211_IFTYPE_STATION` after the call to `get_if_type()`. Commenting the call to `get_if_type()` and instead setting the two variables directly (which is almost the same as reverting the commit in question) causes `add_mon_if()` to fail with the correct error message. So it seems that `get_if_type()` does not communicate with the netlink socket correctly and leaves it in a state that causes the failure to fail in `add_mon_if()`.

johnthacker:
The Linux kernel documentation for Netlink says:

"Note that unless the NLM_F_ACK flag is set on the request Netlink will not respond with NLMSG_ERROR if there is no error. To avoid having to special-case this quirk it is recommended to always set NLM_F_ACK." (https://docs.kernel.org/userspace-api/netlink/intro.html#nl-msg-type)

In other places, it suggests that sockets do have Auto-ACK enabled by default, however.

Regardless, follow instructions and always set NLM_F_ACK, and receive messages until we get the ACK.

Fix #1508

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -756,7 +756,10 @@ DIAG_ON_NARROWING
 
 	err = nl_send_sync(state->nl_sock, msg); // calls nlmsg_free()
 	if (err < 0) {
-		if (err == -NLE_FAILURE) {
+		switch (err) {
+
+		case -NLE_FAILURE:
+		case -NLE_AGAIN:
 			/*
 			 * Device not available; our caller should just
 			 * keep trying.  (libnl 2.x maps ENFILE to
@@ -765,10 +768,20 @@ DIAG_ON_NARROWING
 			 * about that.)
 			 */
 			return 0;
-		} else {
+
+		case -NLE_OPNOTSUPP:
+			/*
+			 * Device is a mac80211 device but adding it as a
+			 * monitor mode device isn't supported.  Report our
+			 * error.
+			 */
+			return PCAP_ERROR_RFMON_NOTSUP;
+
+		default:
 			/*
 			 * Real failure, not just "that device is not
-			 * available.
+			 * available."  Report a generic error, using the
+			 * error message from libnl.
 			 */
 			snprintf(handle->errbuf, PCAP_ERRBUF_SIZE,
 			    "%s: nl_send_sync failed adding %s interface: %s",
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -644,9 +644,6 @@ get_if_type(pcap_t *handle, int sock_fd, struct nl80211_state *state,
 	if (ifindex == -1)
 		return PCAP_ERROR;
 
-	struct nl_cb *cb = nl_cb_alloc(NL_CB_DEFAULT);
-	nl_cb_set(cb, NL_CB_VALID, NL_CB_CUSTOM, if_type_cb, (void*)type);
-
 	msg = nlmsg_alloc();
 	if (!msg) {
 		snprintf(handle->errbuf, PCAP_ERRBUF_SIZE,
@@ -683,7 +680,10 @@ get_if_type(pcap_t *handle, int sock_fd, struct nl80211_state *state,
 		}
 	}
 
+	struct nl_cb *cb = nl_cb_alloc(NL_CB_DEFAULT);
+	nl_cb_set(cb, NL_CB_VALID, NL_CB_CUSTOM, if_type_cb, (void*)type);
 	nl_recvmsgs(state->nl_sock, cb);
+	nl_cb_put(cb);
 
 	/*
 	 * Success.
@@ -695,6 +695,7 @@ nla_put_failure:
 	    "%s: nl_put failed getting interface type",
 	    device);
 	nlmsg_free(msg);
+	// Do not call nl_cb_put(): nl_cb_alloc() has not been called.
 	return PCAP_ERROR;
 }
 
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -654,11 +654,13 @@ get_if_type(pcap_t *handle, int sock_fd, struct nl80211_state *state,
 		return PCAP_ERROR;
 	}
 
-	genlmsg_put(msg, 0, 0, genl_family_get_id(state->nl80211), 0,
+	genlmsg_put(msg, NL_AUTO_PORT, NL_AUTO_SEQ,
+		    genl_family_get_id(state->nl80211), 0,
 		    0, NL80211_CMD_GET_INTERFACE, 0);
 	NLA_PUT_U32(msg, NL80211_ATTR_IFINDEX, ifindex);
 
-	err = nl_send_auto_complete(state->nl_sock, msg);
+	err = nl_send_auto(state->nl_sock, msg);
+	nlmsg_free(msg);
 	if (err < 0) {
 		if (err == -NLE_FAILURE) {
 			/*
@@ -668,7 +670,6 @@ get_if_type(pcap_t *handle, int sock_fd, struct nl80211_state *state,
 			 * to that, but there's not much we can do
 			 * about that.)
 			 */
-			nlmsg_free(msg);
 			return 0;
 		} else {
 			/*
@@ -676,9 +677,8 @@ get_if_type(pcap_t *handle, int sock_fd, struct nl80211_state *state,
 			 * available.
 			 */
 			snprintf(handle->errbuf, PCAP_ERRBUF_SIZE,
-			    "%s: nl_send_auto_complete failed getting interface type: %s",
+			    "%s: nl_send_auto failed getting interface type: %s",
 			    device, nl_geterror(-err));
-			nlmsg_free(msg);
 			return PCAP_ERROR;
 		}
 	}
@@ -688,8 +688,6 @@ get_if_type(pcap_t *handle, int sock_fd, struct nl80211_state *state,
 	/*
 	 * Success.
 	 */
-	nlmsg_free(msg);
-
 	return 1;
 
 nla_put_failure:
@@ -720,7 +718,8 @@ add_mon_if(pcap_t *handle, int sock_fd, struct nl80211_state *state,
 		return PCAP_ERROR;
 	}
 
-	genlmsg_put(msg, 0, 0, genl_family_get_id(state->nl80211), 0,
+	genlmsg_put(msg, NL_AUTO_PORT, NL_AUTO_SEQ,
+		    genl_family_get_id(state->nl80211), 0,
 		    0, NL80211_CMD_NEW_INTERFACE, 0);
 	NLA_PUT_U32(msg, NL80211_ATTR_IFINDEX, ifindex);
 DIAG_OFF_NARROWING
@@ -728,7 +727,7 @@ DIAG_OFF_NARROWING
 DIAG_ON_NARROWING
 	NLA_PUT_U32(msg, NL80211_ATTR_IFTYPE, NL80211_IFTYPE_MONITOR);
 
-	err = nl_send_auto_complete(state->nl_sock, msg);
+	err = nl_send_sync(state->nl_sock, msg); // calls nlmsg_free()
 	if (err < 0) {
 		if (err == -NLE_FAILURE) {
 			/*
@@ -738,7 +737,6 @@ DIAG_ON_NARROWING
 			 * to that, but there's not much we can do
 			 * about that.)
 			 */
-			nlmsg_free(msg);
 			return 0;
 		} else {
 			/*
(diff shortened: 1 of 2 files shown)


infrastation:
Thank you for waiting. In my test with `wlan0` on an RPi these changes do not restore the expected error message, could you double-check?

Without these changes:
```
tcpdump: SIOCGIFINDEX: No such device
```
With these changes:
```
tcpdump: wlan0: nl_wait_for_ack failed adding mon0 interface: Operation not supported
```
The expected error message is:
```
tcpdump: wlan0: That device doesn't support monitor mode
```

johnthacker:
I agree I don't see that error message, but I don't see that error message by checking out the commit before 0ac887b either. I also don't see it before 24cf361c4d843d620acadc0ff0977dc2364ee81e either. In fact, I checked out libpcap-1.10.4 and built that, and linked tcpdump against it, and saw the same:

```
tcpdump: wlan0: nl_wait_for_ack failed adding mon0 interface: Operation not supported
```

message. I also saw it when checking out libpcap-1.10.3, which surprised me somewhat as the RPi has libpcap-1.10.3 installed on it from Debian Bookworm, and the libpcap on the RPi did give the error message you're looking for.

Then a lightbulb went off:

```
$ ldd /usr/lib/aarch64-linux-gnu/libpcap.so.0.8
	linux-vdso.so.1 (0x0000007fabcc8000)
	libc.so.6 => /lib/aarch64-linux-gnu/libc.so.6 (0x0000007faba60000)
	/lib/ld-linux-aarch64.so.1 (0x0000007fabc80000)
	libdbus-1.so.3 => /lib/aarch64-linux-gnu/libdbus-1.so.3 (0x0000007fab9e0000)
	libsystemd.so.0 => /lib/aarch64-linux-gnu/libsystemd.so.0 (0x0000007fab8f0000)
	libcap.so.2 => /lib/aarch64-linux-gnu/libcap.so.2 (0x0000007fab8c0000)
	libgcrypt.so.20 => /lib/aarch64-linux-gnu/libgcrypt.so.20 (0x0000007fab7b0000)
	liblzma.so.5 => /lib/aarch64-linux-gnu/liblzma.so.5 (0x0000007fab760000)
	libzstd.so.1 => /lib/aarch64-linux-gnu/libzstd.so.1 (0x0000007fab6a0000)
	liblz4.so.1 => /lib/aarch64-linux-gnu/liblz4.so.1 (0x0000007fab650000)
	libgpg-error.so.0 => /lib/aarch64-linux-gnu/libgpg-error.so.0 (0x0000007fab600000)
$ ldd ../libpcap/libpcap.so.1.10.3
	linux-vdso.so.1 (0x0000007f81844000)
	libnl-genl-3.so.200 => /lib/aarch64-linux-gnu/libnl-genl-3.so.200 (0x0000007f81760000)
	libnl-3.so.200 => /lib/aarch64-linux-gnu/libnl-3.so.200 (0x0000007f81710000)
	libdbus-1.so.3 => /lib/aarch64-linux-gnu/libdbus-1.so.3 (0x0000007f81690000)
	libc.so.6 => /lib/aarch64-linux-gnu/libc.so.6 (0x0000007f814e0000)
	/lib/ld-linux-aarch64.so.1 (0x0000007f81800000)
	libsystemd.so.0 => /lib/aarch64-linux-gnu/libsystemd.so.0 (0x0000007f813f0000)
	libcap.so.2 => /lib/aarch64-linux-gnu/libcap.so.2 (0x0000007f813c0000)
	libgcrypt.so.20 => /lib/aarch64-linux-gnu/libgcrypt.so.20 (0x0000007f812b0000)
	liblzma.so.5 => /lib/aarch64-linux-gnu/liblzma.so.5 (0x0000007f81260000)
	libzstd.so.1 => /lib/aarch64-linux-gnu/libzstd.so.1 (0x0000007f811a0000)
	liblz4.so.1 => /lib/aarch64-linux-gnu/liblz4.so.1 (0x0000007f81150000)
	libgpg-error.so.0 => /lib/aarch64-linux-gnu/libgpg-error.so.0 (0x0000007f81100000)
```

**The libpcap installed by Debian isn't built against libnl**, so it's unable to set monitor mode for *any* interface, and libpcap returns `PCAP_ERROR_RFMON_NOTSUP` for the attempt.

So I think that this PR does fix the issue that was introduced by my earlier commit. I do agree that the error message you mention above is the correct one to give, but that's a separate pre-existing issue. I am willing to work on it as well, but perhaps we can commit this fix first?

johnthacker:
It probably wasn't being built against libnl because of #1358.

johnthacker:
I added another commit to return `PCAP_ERROR_RFMON_NOTSUP` instead of the generic error in the case where Netlink reports that adding a device (already successfully queried as a mac80211 device not in monitor mode) in monitor mode is unsupported.

infrastation:
You are right, the user-visible message is not a new problem and addressing one thing at a time would be best.

I tried comprehending the first proposed commit, among other things it seems to plug a memory leak on error. Specifically, in `get_if_type()` it adds calls to `nl_cb_put()` after `nlmsg_free()` everywhere except in the very last instance (where indeed it should not be done) and the branch where `nlmsg_alloc()` fails (where it seems it should). Would the following be easier to follow, if I understand this logic correctly? Would it be better to make the leak plug a separate change?
```patch
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -648,9 +648,6 @@ get_if_type(pcap_t *handle, int sock_fd, struct nl80211_state *state,
 	if (ifindex == -1)
 		return PCAP_ERROR;
 
-	struct nl_cb *cb = nl_cb_alloc(NL_CB_DEFAULT);
-	nl_cb_set(cb, NL_CB_VALID, NL_CB_CUSTOM, if_type_cb, (void*)type);
-
 	msg = nlmsg_alloc();
 	if (!msg) {
 		snprintf(handle->errbuf, PCAP_ERRBUF_SIZE,
@@ -674,7 +671,6 @@ get_if_type(pcap_t *handle, int sock_fd, struct nl80211_state *state,
 			 * about that.)
 			 */
 			nlmsg_free(msg);
-			nl_cb_put(cb);
 			return 0;
 		} else {
 			/*
@@ -685,11 +681,13 @@ get_if_type(pcap_t *handle, int sock_fd, struct nl80211_state *state,
 			    "%s: nl_send_auto_complete failed getting interface type: %s",
 			    device, nl_geterror(-err));
 			nlmsg_free(msg);
-			nl_cb_put(cb);
 			return PCAP_ERROR;
 		}
 	}
 
+	struct nl_cb *cb = nl_cb_alloc(NL_CB_DEFAULT);
+	nl_cb_set(cb, NL_CB_VALID, NL_CB_CUSTOM, if_type_cb, (void*)type);
+
 	err = nl_recvmsgs(state->nl_sock, cb);
 	if (err < 0) {
 		snprintf(handle->errbuf, PCAP_ERRBUF_SIZE,
@@ -723,6 +721,7 @@ nla_put_failure:
 	    "%s: nl_put failed getting interface type",
 	    device);
 	nlmsg_free(msg);
+	// Do not call nl_cb_put(): nl_cb_alloc() has not been called.
 	return PCAP_ERROR;
 }
 
```

Another change is replacing `nl_send_auto_complete()` with `nl_send_auto()`, which is an unrelated change:
```c
/**
 * @deprecated Please use nl_send_auto()
 */
int nl_send_auto_complete(struct nl_sock *sk, struct nl_msg *msg)
{
	return nl_send_auto(sk, msg);
}
```

One other thing that looks not quite right is that (notwithstanding the logic explained for the kernel end of a netlink socket) libnl is supposed to set `NLM_F_ACK` automatically: according to the documentation, this is the default (can be switched off using `nl_socket_disable_auto_ack()`), and the current source code looks consistent with that:
```c
/**
 * Disable automatic request for ACK
 * @arg sk		Netlink socket.
 *
 * The default behaviour of a socket is to request an ACK for
 * each message sent to allow for the caller to synchronize to
 * the completion of the netlink operation. This function
 * disables this behaviour and will result in requests being
 * sent which will not have the NLM_F_ACK flag set automatically.
 * However, it is still possible for the caller to set the
 * NLM_F_ACK flag explicitely.
 */
void nl_socket_disable_auto_ack(struct nl_sock *sk)
{
	sk->s_flags |= NL_NO_AUTO_ACK;
}
```
If this is the case, the change to `genlmsg_put()` does not affect the logic because `NLM_F_ACK` will be set automatically and both `NL_AUTO_PORT` and `NL_AUTO_SEQ` are defined to 0, is this correct? In this case the latter could be a part of a separate style-focused commit to keep the actual bug fix focused.

johnthacker:
Sounds good, I went with your proposed changes.

pcap-linux.c
```diff
@@ -754,49 +754,73 @@ DIAG_ON_NARROWING
 
 	err = nl_send_auto_complete(state->nl_sock, msg);
```
infrastation: As far as I understand, here `nlmsg_free()` could be called exactly once; furthermore, it could be called implicitly from `nl_send_sync()`, in which case this code block would be reduced to roughly this:
```c
err = nl_send_sync(state->nl_sock, msg);
switch (err) {
// return one of the error codes
}
// success
```
Arguably, the latter would only make perfect sense if `nl_wait_for_ack()` indeed returned `-NLE_OPNOTSUPP` etc, which may be not the case.

infrastation:
Thank you, it is becoming less convoluted now.

pcap-linux.c
johnthacker: I am a bit surprised too that `nl_wait_for_ack` returns `-NLE_OPNOTSUPP` instead of the preceding send command, but apparently at least on the RPi that is the case, as seen in the error:

```
tcpdump: wlan0: nl_wait_for_ack failed adding mon0 interface: Operation not supported
```

johnthacker:
Thanks for all the review. I think I'm happy with the current state now; let me know if you want me to squash any commits or anything.

infrastation:
Thank you for preparing solutions to the following problems:
* an incorrect reference to `SIOCGIFINDEX` (a relatively recent bug)
* a not-sufficiently specific reference to `nl_wait_for_ack` instead of returning `PCAP_ERROR_RFMON_NOTSUP` (an older bug)
* a memory leak in `get_if_type()`:
  ```
  opentest: wlan0: Generic error
  (SIOCGIFINDEX: No such device)
  ==80869== 224 bytes in 1 blocks are definitely lost in loss record 2 of 3
  ==80869==    at 0x4889F94: calloc (vg_replace_malloc.c:1328)
  ==80869==    by 0x48FF9BB: nl_cb_alloc (in /usr/lib/aarch64-linux-gnu/libnl-3.so.200.26.0)
  ==80869==    by 0x12C69B: get_if_type (pcap-linux.c:647)
  ==80869==    by 0x12C69B: enter_rfmon_mode (pcap-linux.c:4934)
  ==80869==    by 0x12D127: setup_socket (pcap-linux.c:2505)
  ==80869==    by 0x12D127: pcap_activate_linux (pcap-linux.c:1249)
  ==80869==    by 0x110DAF: pcap_activate (pcap.c:2708)
  ==80869==    by 0x10DE87: main (opentest.c:174)
  ```

On the one hand, some of these changes overwrite each other, but on the other hand it is desirable to make the commits easy to cherry-pick into the libpcap-1.10 branch. Let me see if there is a better way to arrange these commits.

infrastation:
Thank you for waiting. Please see the "nl80211_no_mon_rejig" branch in my libpcap repository, this represents the sum of your changes almost exactly less a few trivial style fixups, but the order and contents of commits is different (one lot of clean-ups followed by three minimal bug fixes), also the bug fix commit messages now include "before" and "after". This should be much easier to follow later if necessary.

I have tested that each of these commits works as expected both if applied to the master branch and if cherry-picked into the libpcap-1.10 branch. Please have a look and use for this pull request as much as it fits.

One thing that is still missing is the rationale in the commit that fixes the original `SIOCGIFINDEX` problem. I did not examine the the netlink message exchange closely, my understanding of the diff is that returning of `NL_STOP` somehow prevented consuming of the `NLMSG_ERROR` part of the response and returning `NL_SKIP` unblocked that, but the call to `nl_wait_for_ack()` is still required. "Receive messages until we get the ACK." does not quite explain the root cause and the solution. A few more comments in the source code would be very appropriate to explain the context.

infrastation:
Thank you, this revision looks good. If anyone else sees a problem with the current revision, please comment, otherwise this is going to be merged in a couple days.

PR #1514, opened 2025-05-08
add edk2 mm request serialization protocol

kraxel:


kraxel:
Ping.

(mail to tcpdump-workers with more details is here: https://seclists.org/tcpdump/2025/q2/22)

guyharris:
@mcr, what should we do about new link-layer type requests while we're waiting for [draft-ietf-opsawg-pcaplinktype](https://datatracker.ietf.org/doc/html/draft-ietf-opsawg-pcaplinktype) to burst out of its chrysalis as a ~~butterfly~~*RFC*?

Once it's an RFC, presumably we'll do our Expert Review and, once that's done, submit a new I-D for one or more new link-layer types, but what do we do in the interim?

mcr:
I think, since you and I will become the Designated Experts, we should just allocated the next number.
Once it becomes an RFC, we will then remember and clue IANA in.
No new I-D is needed. IANA just keeps a list.  The RFC is needed to set the rules for the registry, and to initialize it.

mcr:
> I think, since you and I will become the Designated Experts, we should just allocated the next number. Once it becomes an RFC, we will then remember and clue IANA in. No new I-D is needed. IANA just keeps a list. The RFC is needed to set the rules for the registry, and to initialize it.

I created a new label.
So, actually, let's tell @kraxel that they should go ahead with 302.  I would rather *not* merge this until we can update IANA, but I'm of two minds here.

infrastation:
There is already a label for it.

infrastation:
This would need rebasing on the current master branch and resolving a trivial merge conflict.

infrastation:
Rebased.

infrastation:
Merged, thank you for waiting.

infrastation:
For posterity, `LINKTYPE_EDK2_MM` became available in libpcap 1.10.6.

PR #1517, opened 2025-05-21
Define ETHERTYPE_{SLOW,LLDP} and accept them in pcap_nametoeproto()

nabijaczleweli:
Linux has
```c
#define ETH_P_SLOW	0x8809		/* Slow Protocol. See 802.3ad 43B */
#define ETH_P_LLDP	0x88CC		/* Link Layer Discovery Protocol */
```
for these.

SLOW carries IEEE 802.3 LACP, both defined in annex 57A of [802.3-2018](https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=8457469) and LLDP is defined in sexion 79 of the same.

### DIFF
--- a/ethertype.h
+++ b/ethertype.h
@@ -100,6 +100,9 @@
 #ifndef ETHERTYPE_IPV6
 #define ETHERTYPE_IPV6		0x86dd
 #endif
+#ifndef ETHERTYPE_SLOW
+#define ETHERTYPE_SLOW		0x8809
+#endif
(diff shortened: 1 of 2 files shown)


fxlb:
`ether proto lldp`, `'ether proto \lldp'`, `ether proto slow`and `'ether proto \slow'` are now valid filters. Thus the `pcap-filter` man page should be updated, section `ether proto protocol`.

nabijaczleweli:
#1557

Issue #1512, opened 2025-04-30
"arp host NAME" and "rarp host NAME" reject hosts with IPv6 addresses


infrastation:
I've been attempting some seemingly simple deduplication of the code that generates IPv4 and IPv6 checks for the `host` and `gateway` primitives, and it turned out to be not as simple, and until recently I could not understand why. As it turns out, the current implementation does not handle IPv6 addresses consistently in all cases, as shown below. An immediate way to address this would be to document and to test the current behaviour well; it seems best at some later point to implement a more consistent (i.e. different from the current) behaviour. The problem stands when in a primitive the ID is a string, in which case the code path goes through `gen_scode()`.

There the `Q_GATEWAY` case resolves the string as an Internet hostname, uses all IPv4 addresses and ignores any other (i.e. IPv6) addresses. The latter behaviour arguably is a missing minor feature, but until it gets implemented, ignoring of the IPv6 addresses seems the most reasonable thing to do for an IPv4-only primitive that has been a part of the solution space way before IPv6. For example:
```text
$ filtertest EN10MB 'gateway eth-ipv4-ipv6.host123.libpcap.test'
(000) ld       [8]
(001) jeq      #0x400140e       jt 2	jf 4
(002) ldh      [6]
(003) jeq      #0xaa00          jt 8	jf 4
(004) ld       [2]
(005) jeq      #0x400140e       jt 6	jf 21
(006) ldh      [0]
(007) jeq      #0xaa00          jt 8	jf 21
(008) ldh      [12]
(009) jeq      #0x800           jt 10	jf 14
(010) ld       [26]
(011) jeq      #0xa141e28       jt 21	jf 12
(012) ld       [30]
(013) jeq      #0xa141e28       jt 21	jf 20
(014) jeq      #0x806           jt 16	jf 15
(015) jeq      #0x8035          jt 16	jf 20
(016) ld       [28]
(017) jeq      #0xa141e28       jt 21	jf 18
(018) ld       [38]
(019) jeq      #0xa141e28       jt 21	jf 20
(020) ret      #262144
(021) ret      #0
```
The above is a logical OR of `ip gateway`, `arp gateway` and `rarp gateway`, each of which accepts the same hostname as above, ignores the IPv6 part of the implicit input and generates the protocol-specific subset of the same output. As far as IPv4 goes, this semantics makes complete sense.

[...]

PR #1519, opened 2025-05-22
DRAFT: Factor gen_host46_byname() out.

infrastation:
This is the more difficult part of a prototype I started, but could not quickly complete earlier. The intent is simple: since `gateway X` nominally includes `not host X`, and since both `gen_gateway()` and `gen_scode()` call `pcap_addrinfo()` and process the result(s) in a very similar manner, having a slightly different code block in each function makes the code harder to maintain. However, the complexity turned out to be in the detail (including #1512), so this work-in-progress revision fails many tests. Also before this can be progressed any further, the reject filter tests must cover all code paths that go through `pcap_addrinfo()` and fail to resolve the hostname.

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -7112,10 +7112,8 @@ struct block *
 gen_mcode6(compiler_state_t *cstate, const char *s, bpf_u_int32 masklen,
     struct qual q)
 {
-	struct addrinfo *res;
-	struct in6_addr *addr;
+	struct in6_addr addr;
 	struct in6_addr mask;
-	struct block *b;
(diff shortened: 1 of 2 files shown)


infrastation:
The previous revision failed 80 tests, this revision passes all tests, including the new tests it adds. It also includes a bug fix for the bug #1512. It passes a build matrix with Valgrind on Linux.

infrastation:
This revision handles integer types a bit better, let's see if that clears the Windows CI failure.

infrastation:
This revision spells an assignment in a conditional expression in a way that hopefully won't upset MSVC, let's see if that worked.

PR #1520, opened 2025-05-22
WIP/RFC possible use of va_list in gencode.c

infrastation:
This is another unfinished piece that didn't quite work out the way I originally expected, but took some time to produce anyway. Potentially this could be developed further, please have a look when you have time.

This is one of the optimizations that avoid generating duplicate instructions in the first place. On this occasion it is the code produced by a series of ORed `gen_cmp()` calls, all of which load the same exact packet data, in which case, obviously, only the first load is required. The draft demonstrates how this can be done for PPP, ATM and ARCnet (the tests pass), all of which hard-code the values.


infrastation:
This minor optimisation method would require more work to become useful, and there is no time for that, so let's close it until time is good to retry.

PR #1522, opened 2025-05-27
DRAFT Split fix_program() and fix_offset() into smaller parts.

infrastation:
Yet another unfinished piece from my working copy. This one looks mostly complete, but I could not quickly verify that the new revision works as expected in every case. Requires a bit more work.


infrastation:
This can be retried later if necessary.

Issue #1022, opened 2021-06-08
Redundant/unnecessary BPF instructions generated for some capture filters


cjmaynard:
Running dumpcap 3.4.5 on Windows 10 with Npcap version 1.31, based on libpcap version 1.10.1-PRE-GIT.

I recently discovered a few capture filters that cause some redundant and/or unnecessary BPF instructions to be generated.  For example:

1. Capture TCP/IPv6 traffic over port 100: Capture Filter `ip6[6]=6 and tcp port 100` generates redundant instructions 003 and 004:

```
    (000) ldh      [12]
    (001) jeq      #0x86dd          jt 2    jf 10
    (002) ldb      [20]
    (003) jeq      #0x6             jt 4    jf 10
    (004) jeq      #0x6             jt 5    jf 10
    (005) ldh      [54]
    (006) jeq      #0x64            jt 9    jf 7
    (007) ldh      [56]
    (008) jeq      #0x64            jt 9    jf 10
    (009) ret      #262144
    (010) ret      #0
```

2. Reversing the order doesn't help, but adding a redundant _clause_ isn't ignored as one might expect:  Capture filter `tcp port 100 and ip6[6]=6 and tcp port 100` generates redundant instructions 008-013:

```
    (000) ldh      [12]
    (001) jeq      #0x86dd          jt 2    jf 15
    (002) ldb      [20]
    (003) jeq      #0x6             jt 4    jf 15
    (004) ldh      [54]
    (005) jeq      #0x64            jt 8    jf 6
    (006) ldh      [56]
    (007) jeq      #0x64            jt 8    jf 15
    (008) ldb      [20]
    (009) jeq      #0x6             jt 10   jf 15
    (010) ldh      [54]
    (011) jeq      #0x64            jt 14   jf 12
    (012) ldh      [56]
    (013) jeq      #0x64            jt 14   jf 15
    (014) ret      #262144
    (015) ret      #0
```

3. Capture traffic from source port 100 over IP:  Capture filter `ip and src port 100` generates BPF code that prioritizes SCTP (132) traffic, but why is SCTP (132) prioritized over TCP (6) or UDP (17)?  Surely it's not as widely used in most applications, right?

```
    (000) ldh      [12]
    (001) jeq      #0x800           jt 2    jf 12
    (002) ldb      [23]
    (003) jeq      #0x84            jt 6    jf 4
    (004) jeq      #0x6             jt 6    jf 5
[...]

PR #1525, opened 2025-06-03
Fix IPv4 multicast upper bound

codido:
IPv4 multicast addresses are defined to be in 224.0.0.0/4.

While the generated code for multicast checked the lower bound, it
failed to verify the upper bound, mistakenly including the reserved
range (240.0.0.0/4) and broadcast (255.255.255.255).

Fix this by explicitly comparing the first four bits of the address to
the multicast ones.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -62,6 +62,7 @@ DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
       Require "(ip|ip6) proto" value to be within range.
       Require "protochain" value to be within range.
       Always enable the "gateway" primitive.
+      Fix IPv4 multicast filtering to only include 224.0.0.0/4.
     rpcap:
       Support user names and passwords in rpcap:// and rpcaps:// URLs.
       Add a -t flag to rpcapd to specify the data channel port; from
--- a/gencode.c
+++ b/gencode.c
@@ -8095,7 +8095,12 @@ gen_multicast(compiler_state_t *cstate, int proto)
 
 	case Q_IP:
 		b0 = gen_linktype(cstate, ETHERTYPE_IP);
-		b1 = gen_cmp_ge(cstate, OR_LINKPL, 16, BPF_B, 224);
+
+		/*
+		 * Compare address with 224.0.0.0/4
+		 */
+		b1 = gen_mcmp(cstate, OR_LINKPL, 16, BPF_B, 0xe0, 0xf0);
+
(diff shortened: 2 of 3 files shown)


infrastation:
This is well spotted, thank you. Please note the proposed change produces a filter program that leaves space for optimization due to the additional load instruction from the same offset:
```text
$ ./testprogs/filtertest EN10MB 'ip multicast'
(000) ldh      [12]
(001) jeq      #0x800           jt 2	jf 6
(002) ldb      [30]
(003) jge      #0xe0            jt 4	jf 6
(004) jgt      #0xef            jt 6	jf 5
(005) ret      #262144
(006) ret      #0
$ ./testprogs/filtertest -O EN10MB 'ip multicast'
(000) ldh      [12]
(001) jeq      #0x800           jt 2	jf 7
(002) ldb      [30]
(003) jge      #0xe0            jt 4	jf 7
(004) ldb      [30]
(005) jgt      #0xef            jt 7	jf 6
(006) ret      #262144
(007) ret      #0
```
Thus for the bug fix please use a masked comparison to produce a minimum length program in both cases, except the initial one or two no-op instructions, which are a separate matter (#1022):
```diff
--- a/gencode.c
+++ b/gencode.c
@@ -8095,7 +8095,7 @@ gen_multicast(compiler_state_t *cstate, int proto)
 
 	case Q_IP:
 		b0 = gen_linktype(cstate, ETHERTYPE_IP);
-		b1 = gen_cmp_ge(cstate, OR_LINKPL, 16, BPF_B, 224);
+		b1 = gen_mcmp(cstate, OR_LINKPL, 16, BPF_B, 224, 0xf0); // 224.0.0.0/4
 		gen_and(b0, b1);
 		return b1;
 ```
With this change it may be useful to add an "unopt" version to the test as well.

Regardless of the method, this bug fix will need to be mentioned in the "packet filtering" section of the change log because it changes the currently implemented behaviour.

codido:
Thank you for the prompt review and comments!
Adjusted the patch accordingly.

infrastation:
Thank you, this looks good and is going to merged soon unless anyone objects.

infrastation:
If you would like to make the commit message match the current change to the code, please do so soon.

codido:
Thanks for pointing that out, how's this?

infrastation:
Merged, thank you. This bug fix should be useful in the stable branch as well.

PR #1526, opened 2025-06-21
Add dlt_choices[] entries for USERx

fxlb:
After the change, tcpdump prints more accurate information, like:
reading from file example.pcap, link-type USER2 (Private use 2), \\
  snapshot length 262144
instead of:
reading from file example.pcap, link-type 149, snapshot length 262144

### DIFF
--- a/pcap.c
+++ b/pcap.c
@@ -3324,6 +3324,22 @@ static struct dlt_choice dlt_choices[] = {
 	DLT_CHOICE(FIRA_UCI, "Ultra-wideband controller interface protocol"),
 	DLT_CHOICE(MDB, "Multi-Drop Bus"),
 	DLT_CHOICE(DECT_NR, "DECT New Radio"),
+	DLT_CHOICE(USER0, "Private use 0"),
+	DLT_CHOICE(USER1, "Private use 1"),
+	DLT_CHOICE(USER2, "Private use 2"),
+	DLT_CHOICE(USER3, "Private use 3"),
+	DLT_CHOICE(USER4, "Private use 4"),
+	DLT_CHOICE(USER5, "Private use 5"),
+	DLT_CHOICE(USER6, "Private use 6"),
+	DLT_CHOICE(USER7, "Private use 7"),
+	DLT_CHOICE(USER8, "Private use 8"),
+	DLT_CHOICE(USER9, "Private use 9"),
+	DLT_CHOICE(USER10, "Private use 10"),
+	DLT_CHOICE(USER11, "Private use 11"),
+	DLT_CHOICE(USER12, "Private use 12"),
+	DLT_CHOICE(USER13, "Private use 13"),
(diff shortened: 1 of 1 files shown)


mcr (approved):
Looks good to me... I have two thoughts going forward. 1) we might want to #ifdef EMBED or SMALL or whatever openwrt/lede uses to remove this.  2) are these really DLT values or LINKTYPE values? 3) we might want to generate this table from an XML table from IANA if so.

fxlb:
> 1) we might want to #ifdef EMBED or SMALL or whatever openwrt/lede uses to remove this.

They use `TCPDUMP_MINI` for tcpdump. I don't see macro for libpcap.

fxlb:
> 2) are these really DLT values or LINKTYPE values?

If I understand correctly the `DLT_CHOICE` macro definition, the entries are currently based on DLT_xxx values.
`#define DLT_CHOICE(code, description) { #code, description, DLT_ ## code }`

guyharris:
> are these really DLT values or LINKTYPE values?

They really are DLT_ values.

LINKTYPE_ values are not exposed by the current API; they are translated to DLT_ values before being stored in the `pcap_t`, which means that `pcap_datalink()` returns DLT_ values even for a capture file.

When I introduced the LINKTYPE_ values, the goal was to 1) arrange that what got written to pcap files (I think it predated the existence of pcapng) would be independent of the platform on which the file was being written, so that files written on one platform could be read on a platform where the DLT_ #define had a different numerical value for that link-layer type and 2) *not* break the ABI by changing DLT_ values.

guyharris:
> we might want to #ifdef EMBED or SMALL or whatever openwrt/lede uses to remove this.

I'm not seeing anything in [OpenWRT's libpcap patches](https://github.com/openwrt-mirror/openwrt/tree/master/package/libs/libpcap/patches) to do that.

If lede refers to [this project](https://github.com/lede-project/source), I'm not seeing anything in [its libpcap patches](https://github.com/lede-project/source/tree/master/package/libs/libpcap/patches) to do that, either.

guyharris:
> They use `TCPDUMP_MINI` for tcpdump. I don't see macro for libpcap.

If there's a general consensus on how to minimize tcpdump, we might want to offer that ourselves; if different projects have different ideas of what they want included or excluded, we'd probably end up having to have a large number of different #ifdefs for different projects to use.

fxlb:
> I'm not seeing anything in [OpenWRT's libpcap patches](https://github.com/openwrt-mirror/openwrt/tree/master/package/libs/libpcap/patches) to do that.

The current patches are here:
https://github.com/openwrt/openwrt/tree/master/package/libs/libpcap/patches

fxlb:
> If there's a general consensus on how to minimize tcpdump, we might want to offer that ourselves

Yes, it could be a step after 5.0 release.

fxlb:
There was https://github.com/the-tcpdump-group/tcpdump/pull/518.

PR #1530, opened 2025-06-25
socketcan: add CAN XL Remote Request Substitution CANXL_RRS bit

hartkopp:
The Remote Request Substitution bit is a dominant bit ("0") in the CAN XL frame. As some CAN XL controllers support to access this bit a new CANXL_RRS value has been defined for the canxl_frame.flags element.

Defined in Linux mainline since Linux 6.15

Additionally fixed a typo 'Context' -> 'Content'


hartkopp:
Changed the commit message to "socketcan: add CAN XL Remote Request Substitution CANXL_RRS bit"

PR #1531, opened 2025-07-01
TESTrun: increase the default filter test timeout to 5 seconds.

guyharris:
Sometimes we seem to get glitches in which filter tests time out on some machines but not on others; increase the timeout from 1 second to 5 seconds to handle, for example, heavy loads on slower machines.

### DIFF
--- a/testprogs/TESTrun
+++ b/testprogs/TESTrun
@@ -16473,12 +16473,22 @@ if (system ("$filtertest -h >/dev/null 2>&1") >> 8) {
 	exit 2;
 }
 
-# Every test in this file uses an expression that under normal conditions takes
-# well under one second to process, so if a filtertest invocation is taking
-# longer, it is likely a regression.  Or an invocation via Valgrind, which
-# demands a sensible host-specific override of the timeout value.
+# Every test in this file uses an expression that under normal
+# conditions should take under one second to process; however,
+# "normal conditions" may not always apply, due, for example,
+# to a significant load on the system running the tests, so
+# we allow five seconds for the test, to avoid failures due
+# to glitches of that sort.
+#
+# If a filtertest invocation is taking longer, it is likely a
+# regression.
+#
+# However, an invocation via Valgrind demands a sensible host-specific
+# override of the timeout value, which is done by setting the
+# FILTERTEST_TIMEOUT environment variable. The build_matrix.sh
+# script uses that variable for this purpose.
 $test_timeout = defined $ENV{FILTERTEST_TIMEOUT} ?
-	$ENV{FILTERTEST_TIMEOUT} : 1;
+	$ENV{FILTERTEST_TIMEOUT} : 5;
 
 if ($test_timeout eq '0') {
 	print "INFO: Not using a test timeout (FILTERTEST_TIMEOUT=0).\n";


infrastation:
Buildbot configuration already sets `FILTERTEST_TIMEOUT` to 2 for most Buildbot workers, which overshoots badly enough already (let's say if a test normally takes 0.01 seconds, allowing it to run for 1.99 seconds can mean a failure to detect a regression). Also after the number of tests reached a certain number, I removed Valgrind from the worker hosts because it became not practicable to run the tests for so long. However, if Valgrind is present and `FILTERTEST_TIMEOUT` isn't set, the timeout should increase proportionally, which would not happen with this change.

So perhaps this problem requires a different approach. In the narrow sense, if a particular worker host tends to time out, it would be better to increase its timeout in the configuration. In a more generic sense, it usually is one or just a few tests that time out, so it would be better to retry a test that has timed out another one or two times, and fail it iff it never manages to pass. Then sporadic non-reproducible timeouts will stay out of scope, but apparently reproducible regressions will count as failures.

In any case, it is a matter of getting the balance right rather than raising the timeout every time. CI builds do time out for other reasons as well (usually network and unrelated segfaults) and require occasional restarts anyway, so let's say one restart a month would not be unexpected.

guyharris:
> (let's say if a test normally takes 0.01 seconds, allowing it to run for 1.99 seconds can mean a failure to detect a regression

What sort of regression are you talking about here?

(Ideally, the timeout would be a *CPU* time timeout, as that would be independent of system load, but, whilst `RLIMIT_CPU`/`SIGXCPU` are in POSIX, 1) I don't know whether Windows has an equivalent and 2) I think the last time we discussed this there was an issue with using it on UN\*Xes.)

guyharris:
So by what mechanism can I force a rebuild of https://ci.tcpdump.org/#/builders/60/builds/1525? The "Rebuild" button says I don't have the role 'admins'.

infrastation:
The timeout is meant to catch performance regressions: nesting loops the wrong way around, or reading the same external file 1000 times instead of once, or going into an O(n²) branch for a large n, etc.  The kind of breakage that formally does the expected thing eventually, but late enough to cause breakage elsewhere.

In this sense, enforcing a hard limit on the CPU time per `filtertest` invocation would indeed be better than managing the retries. If at least some OSes implement that reliably and with sub-second granularity, that would do the job well enough.

The instructions should be in your inbox.

infrastation:
Since `setrlimit()` operates in units of integer seconds and the lowest specifiable CPU time limit is 1, the closest practicable thing to the perfect solution I can see would be:
* Setting `RLIMIT_CPU` to 1 (when supported) and setting `FILTERTEST_TIMEOUT` to something larger, to accommodate the host being busy with other work, when running without Valgrind.
* Increasing both `RLIMIT_CPU` (when supported) and `FILTERTEST_TIMEOUT` proportionally, when running with Valgrind.

Would that make sense?

guyharris:
> Since `setrlimit()` operates in units of integer seconds and the lowest specifiable CPU time limit is 1, the closest practicable thing to the perfect solution I can see would be:
> 
> * Setting `RLIMIT_CPU` to 1 (when supported) and setting `FILTERTEST_TIMEOUT` to something larger, to accommodate the host being busy with other work, when running without Valgrind.

If the OS has `setrlimit()`, shouldn't setting the CPU time resource limit to 1 second, and not setting any real-time limit, suffice?

(The Single UNIX Specification didn't, until the 2024 version, require that `ulimit -t` *secs* set `RLIMIT_CPU` to *secs*, but I suspect most if not all of the UN\*Xes on which we run support that - I think it dates back to System V Release 4.)

> * Increasing both `RLIMIT_CPU` (when supported) and `FILTERTEST_TIMEOUT` proportionally, when running with Valgrind.
> 
> Would that make sense?

infrastation:
Indeed it would be easy to show in `config.h` whether `setrlimit(RLIMIT_CPU, ...)` is available and to specify a "CPU limit" argument to `filtertest` in all cases. Then iff it is available, `TESTrun` could omit the `timeout` wrapper and `filtertest` could pass the supplied argument to `setrlimit()`.

Arguably, one reason for setting a wall clock deadline in all cases is to catch code paths that time out without consuming CPU time (network, including DNS resolution, or erroneously using `sleep()` instead of `usleep()` to hide a race condition).

infrastation:
These constraints could be related together as follows:
* If `sysconf()` and `times()` are available, in `filtertest` after the test exit with an error if `tms.tms_utime` exceeds a certain threshold, let's say, 0.1s by default.
* If `setrlimit()` is available, in `filtertest` before the test set `RLIMIT_CPU` to the lowest integer value that is greater than the above threshold (in this case 1s by default).
* In `TESTrun` always set wall clock time limit to a value that is greater than the above threshold (let's say 5s by default).
* If Valgrind is in use, in `TESTrun` multiply the `tms.tms_utime` and the wall clock time threshold by, let's say, 10.

guyharris:
OK, dropping for now, so that I can get my libpcap-master-branch repository back to that branch.

PR #1537, opened 2025-07-24
Fix resource leak on malloc failure in pcap-linux.c

moon044:
### What this fixes

Hi,  
This PR fixes a resource leak in `pcap-linux.c`.  
If `malloc()` fails after opening a socket, the file descriptor is not closed, which can cause a resource leak.

### Fix

Added `close(sock_fd);` before returning on allocation failure.

Thank you for the review!

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -2654,6 +2654,7 @@ setup_socket(pcap_t *handle, int is_any_device)
 		if (handle->dlt_list == NULL) {
 			pcapint_fmt_errmsg_for_errno(handle->errbuf,
 			    PCAP_ERRBUF_SIZE, errno, "malloc");
+			close(sock_fd);
 			return (PCAP_ERROR);
 		}
 		handle->dlt_list[0] = DLT_LINUX_SLL;


pcap-linux.c
```diff
@@ -2652,6 +2652,7 @@ setup_socket(pcap_t *handle, int is_any_device)
 		handle->linktype = DLT_LINUX_SLL;
 		handle->dlt_list = (u_int *) malloc(sizeof(u_int) * 2);
 		if (handle->dlt_list == NULL) {
+			close(sock_fd);
```
guyharris: This should be done *after* the `pcapint_fmt_errmsg_for_errno()` call, so that it doesn't overwrite the `errno` value from the `malloc()` failure.

moon044: Oh, thanks for the feedback!
I’ve updated the code so that `close(sock_fd);` is now called after `pcapint_fmt_errmsg_for_errno()`, as you suggested.
Please let me know if there’s anything else I can improve.
Really appreciate your review!

Issue #1544, opened 2025-08-10
Need to generate arbitrary-length error strings and provide a new API needed to fetch those arbitrary-length error strings


guyharris:
libpcap has traditionally generated error message strings by formatting into a fixed-length buffer of size `PCAP_ERRBUF_SIZE` (or `PCAP_ERRBUF_SIZE + 1`).

This causes `-Wformat-truncation` warnings to be generated by GCC, Clang, and maybe other compilers.

We should, instead, generate dynamically-allocated error messages using `pcapint_asprintf()`, have a new API routine to fetch the arbitrary-length message (callers of `pcap_geterr()` may expect the string it returns to fit in `PCAP_ERRBUF_SIZE` bytes, even though it's not documented as doing that), and have `pcap_geterr()` and `pcap_perror()` truncate that to fit in `PCAP_ERRBUF_SIZE` *and* not to truncate in the middle of a UTF-8 sequence.

That routine will be defined to return UTF-8 error messages; callers who want some other character encoding, for any reason, must do the translation themselves.

That's straightforward for messages fetched from the `pcap_t` - have a `char *` variable that points to the last error message generated, initialize it to `NULL` when a `pcap_t` is allocated, and have routines that are passed a pointer to a `char *` (similar to `pcapint_asprintf()`) and that free whatever that `char *` points to before overwriting the `char *` with a pointer to the new message, and change `pcapint_fmt_errmsg_for_errno()` and `pcapint_fmt_errmsg_for_win32_err()` to work similarly.

It's a bit more work for routines such as `pcap_create()` that are passed pointers to fixed-length buffers; that will require replacements for those routines.

See #1543, #1029, and places in the code for which there *aren't* issues but there *are* `DIAG_OFF_FORMAT_TRUNCATION` calls to suppress the warnings.

infrastation:
A potential means to address the problem could be introducing an option to control the error buffer size and defaulting it (and setting the minimum valid value) to `PCAP_ERRBUF_SIZE`:
```c
char myerrbuf[10240] = "";
pcap_setopt(p, PCAP_OPT_ERRBUFSIZE, sizeof(myerrbuf));
```
Then every function that takes a handle and an error string buffer (e.g. `pcap_setnonblock()`) or just a handle (e.g. `pcap_activate()`) would use the buffer size configured for the handle, and every function that takes an error string buffer only (e.g. `pcap_create()`) would use `PCAP_ERRBUF_SIZE`. Would that work well?

Issue #1543, opened 2025-08-10
dpdk-pcap.c compilation warnings


idogoshen:
[fxlb](https://github.com/fxlb) commented [3 days ago](https://github.com/the-tcpdump-group/libpcap/pull/1541#issuecomment-3163395202) • 
Many warnings with clang-16:
```
pcap-dpdk.c:335:15: warning: variable 'gather_len' set but not used [-Wunused-but-set-variable]
        unsigned int gather_len =0;
                     ^
pcap-dpdk.c:474:32: warning: implicit conversion loses integer precision: 'uint64_t' (aka 'unsigned long') to 'u_int' (aka 'unsigned int') [-Wshorten-64-to-32]
                ps->ps_recv = pd->curr_stats.ipackets;
                            ~ ~~~~~~~~~~~~~~~^~~~~~~~
pcap-dpdk.c:475:32: warning: implicit conversion loses integer precision: 'uint64_t' (aka 'unsigned long') to 'u_int' (aka 'unsigned int') [-Wshorten-64-to-32]
                ps->ps_drop = pd->curr_stats.ierrors;
                            ~ ~~~~~~~~~~~~~~~^~~~~~~
pcap-dpdk.c:477:34: warning: implicit conversion loses integer precision: 'uint64_t' (aka 'unsigned long') to 'u_int' (aka 'unsigned int') [-Wshorten-64-to-32]
                ps->ps_ifdrop = pd->curr_stats.imissed;
                              ~ ~~~~~~~~~~~~~~~^~~~~~~
pcap-dpdk.c:550:13: warning: implicit conversion loses integer precision: 'size_t' (aka 'unsigned long') to 'int' [-Wshorten-64-to-32]
        for (int i=prefix_len; device[i]; i++){
                 ~ ^~~~~~~~~~
```

@[fxlb](https://github.com/fxlb)
Member
fxlb commented [3 days ago](https://github.com/the-tcpdump-group/libpcap/pull/1541#issuecomment-3163437388)
Other warnings with GCC 12.2:
```
pcap-dpdk.c: In function 'pcap_dpdk_activate':
pcap-dpdk.c:785:52: warning: '%s' directive output may be truncated writing up to 255 bytes into a region of size 236 [-Wformat-truncation=]
  785 |                             "Can't open device %s: %s",
      |                                                    ^~
  786 |                             p->opt.device, dpdk_pre_init_errbuf);
[...]

PR #1545, opened 2025-08-10
dpdk: suppress -Wformat-truncation warnings.

guyharris:
This doesn't fix them - nothing short of changing the way we generate error message strings will do that - but at least it means we don't get annoyed by them, and it also marks places that would need to be changed if we change the way we generate those error message strings. See #1544 for an issue for making those changes.

This addresses part of #1543.

### DIFF
--- a/pcap-dpdk.c
+++ b/pcap-dpdk.c
@@ -112,6 +112,8 @@ env DPDK_CFG="--log-level=debug -l0 -dlibrte_pmd_e1000.so -dlibrte_pmd_ixgbe.so 
 #include <rte_bus.h>
 #include <rte_version.h>
 
+#include "diag-control.h"
+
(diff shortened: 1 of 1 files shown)


fxlb:
Test on Debian bookworm with package dpdk-dev:
Missing `#include "diag-control.h"`
Line 820: `s/DIAG_NF_FORMAT_TRUNCATION/DIAG_ON_FORMAT_TRUNCATION/`

guyharris:
> Test on Debian bookworm with package dpdk-dev:
> Missing `#include "diag-control.h"`
> Line 820: `s/DIAG_NF_FORMAT_TRUNCATION/DIAG_ON_FORMAT_TRUNCATION/`

Thanks. Done.

PR #1546, opened 2025-08-12
dpdk: suppress warning unused variable 'gather_len'

idogoshen:
https://github.com/the-tcpdump-group/libpcap/issues/1543

Issue #1159, opened 2023-02-06
Compiling errors after updating dpdk to version higher then v21.11-rc1


Emanuesson:
Due to changes in the shipped headers by DPDK, especially within the following
commit, libpcap became incompatible to compile against dpdk:
<https://github.com/DPDK/dpdk/commit/295968d1740760337e16b0d7914875c5cac52850>

How to reproduce:
```
mkdir build && cd build
cmake .. && make
```

Reproduced with the following versions:
```
$ git describe --tags --always
 libpcap-1.10-bp-695-g8cb3bac2
$ uname -a
 Linux pc1 6.1.8-arch1-1 #1 SMP PREEMPT_DYNAMIC Tue, 24 Jan 2023 21:07:04 +0000 x86_64 GNU/Linux
$ gcc -v
 gcc version 12.2.1 20230111 (GCC) 
$ cmake --version
 cmake version 3.25.2
$ pkg-config --modversion libdpdk
 22.11.1
```

The following compiler-errors are visible during the build:
```
/libpcap/pcap-dpdk.c:202:18: error: ‘struct rte_eth_rxmode’ has no member named ‘split_hdr_size’
  202 |                 .split_hdr_size = 0,
      |                  ^~~~~~~~~~~~~~
/libpcap/pcap-dpdk.c:205:28: error: ‘ETH_MQ_TX_NONE’ undeclared here (not in a function); did you mean ‘RTE_ETH_MQ_TX_NONE’?
  205 |                 .mq_mode = ETH_MQ_TX_NONE,
      |                            ^~~~~~~~~~~~~~
      |                            RTE_ETH_MQ_TX_NONE
/libpcap/pcap-dpdk.c: In function ‘check_link_status’:
/libpcap/pcap-dpdk.c:502:38: error: ‘ETH_LINK_UP’ undeclared (first use in this function); did you mean ‘RTE_ETH_LINK_UP’?
  502 |         return plink->link_status == ETH_LINK_UP;
      |                                      ^~~~~~~~~~~
      |                                      RTE_ETH_LINK_UP
/libpcap/pcap-dpdk.c:502:38: note: each undeclared identifier is reported only once for each function it appears in
/libpcap/pcap-dpdk.c: In function ‘pcap_dpdk_activate’:
[...]

PR #1547, opened 2025-08-14
dpdk: refactor use rte_pktmbuf_read instead of dpdk_gather_data

idogoshen:
Use dpdk standard function [rte_pktmbuf_read](https://doc.dpdk.org/api/rte__mbuf_8h.html#a94675f32ea4655e1f10e209af8d7a229) instead of custom made code

pcap-dpdk will now support packet size up to 16K as defined in [pcap-dpdk.c#L168: #define RTE_ETH_PCAP_SNAPLEN RTE_ETHER_MAX_JUMBO_FRAME_LEN](https://github.com/the-tcpdump-group/libpcap/blob/fab168eca9c9b2e240052b53cd7065322c956b99/pcap-dpdk.c#L168). 
In case anyone thinks more is needed then [pcap-dpdk.c#L198 pcap_tmp_buf](https://github.com/the-tcpdump-group/libpcap/blob/fab168eca9c9b2e240052b53cd7065322c956b99/pcap-dpdk.c#L198) can be increased to 64K.

relates to https://github.com/the-tcpdump-group/libpcap/issues/1543

### DIFF
--- a/pcap-dpdk.c
+++ b/pcap-dpdk.c
@@ -280,17 +280,6 @@ static inline void calculate_timestamp(struct dpdk_ts_helper *helper,struct time
 	timeradd(&(helper->start_time), &cur_time, ts);
 }
 
-static void dpdk_gather_data(unsigned char *data, uint32_t len, struct rte_mbuf *mbuf)
-{
-	uint32_t total_len = 0;
-	while (mbuf && (total_len+mbuf->data_len) < len ){
-		rte_memcpy(data+total_len, rte_pktmbuf_mtod(mbuf,void *),mbuf->data_len);
-		total_len+=mbuf->data_len;
-		mbuf=mbuf->next;
-	}
-}
-
-
 static int dpdk_read_with_timeout(pcap_t *p, struct rte_mbuf **pkts_burst, const uint16_t burst_cnt){
 	struct pcap_dpdk *pd = (struct pcap_dpdk*)(p->priv);
 	int nb_rx = 0;
@@ -331,10 +320,9 @@ static int pcap_dpdk_dispatch(pcap_t *p, int max_cnt, pcap_handler cb, u_char *c
 	// In DPDK, pkt_len is sum of lengths for all segments. And data_len is for one segment
 	uint32_t pkt_len = 0;
 	uint32_t caplen = 0;
-	u_char *bp = NULL;
+	const u_char *bp = NULL;
 	int i=0;
 	int pkt_cnt = 0;
-	u_char *large_buffer=NULL;
 	int timeout_ms = p->opt.timeout;
 
 	/*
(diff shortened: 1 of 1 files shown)


mcr:
Hi, we have decided that testing against DPDK is more than The Tcpdump Group can reasonably do.
We looking for someone from the DPDK team to review all such submissions.  Are you a core DPDK person? Maybe you can refer one.

guyharris:
> Hi, we have decided that testing against DPDK is more than The Tcpdump Group can reasonably do.

In particular:

We can do "smoke tests" to make sure that changes will *compile*, by building on a (virtual or real) machine with the DPDK SDK installed. However, people submitting changes to pcap-dpdk.c have presumably already done that.

We don't, as an organization, have any machines with DPDK-supporting hardware to allow us to test the changes; perhaps some individuals have that, but they may not have the time to do so. People submitting changes to pcap-dpdk.c probably have that.

I don't know whether any libpcap core developers are sufficiently familiar with the DPDK libraries to be able to review those changes. That's why:

> We looking for someone from the DPDK team to review all such submissions. Are you a core DPDK person? Maybe you can refer one.

Part of the problem is that the DPDK APIs don't appear to be "stable" in the sense that code written to older versions will continue to compile; #1159 is an example of the sort of problems that result from that. We don't have the resources to track the DPDK API's changes and update the code to make it compile, so a DPDK team member needs to take that role if libpcap is to continue to support DPDK.

Similarly, we don't have the resources to look into whether the return value of `dpdk_gather_data()` needs to be used or can be discarded; that you for looking into it and submitting #1546 - and for knowing enough about the DPDK API to suggest this pull request.

DPDK seems to be more complicated than the APIs offered by various operating systems for doing network packet capture, and thus supporting it in libpcap needs the help of DPDK specialists who have the knowledge, time, and energy to provide that support.

So if you, or somebody you know, can "own" pcap-dpdk.c, that would be very helpful.

fxlb:
FYI:
https://mails.dpdk.org/archives/dev/2025-March/315377.html

guyharris:
*Somebody* suggested, at one point, a complete redo of the DPDK code to use a different mechanism, in terms similar to what Stephen Hemming said (for example, I think they spoke of allowing regular and capture DPDK use at the same time), but I couldn't seem to find it anywhere in my mailbox the last time I looked (either as a direct email or an email for a comment in an issue/pull request). I think they talked about working on such a redo, but I never heard any more about it.

idogoshen:
Hi

Unfortunately I cannot commit to take ownership as I’m full time busy at a commercial company but I’d try to contribute on best effort basis. I’m not a core DPDK member, but do have decent knowledge being a heavy DPDK user and constantly working in DPDK environments.
In our product we use dpdk to dispatch traffic to third-party/legacy processes that use libpcap for packet acquisition. I can share more info if that's interesting.

Thanx
Ido

idogoshen:
In related to @fxlb  https://github.com/the-tcpdump-group/libpcap/pull/1547#issuecomment-3189838593

[feature/dpdk-ring branch](https://github.com/idogoshen/libpcap/tree/feature/dpdk-ring) adds option to use libpcap as secondary dpdk process. It can be used not only for capture per se but enables packet acquisition from dpdk for legacy libpcap using apps.

For more details see: [README.dpdk-ring.md](https://github.com/idogoshen/libpcap/blob/feature/dpdk-ring/doc/README.dpdk-ring.md)

Should I open pull request for it?

infrastation:
This pull request will not be merged because the old DPDK 18.11 module has been removed via #1697. If the dpdk-ring code you have prepared still has merit with the current DPDK version, let's have a look at it later.

rjarry:
> If the dpdk-ring code you have prepared still has merit with the current DPDK version, let's have a look at it later.

Relying here for context.

Honest warning: DPDK multi-process is a nightmare. In my opinion, libpcap should not deal with *anything* DPDK related.

If performance isn't a concern, a simple rpcapd exporter would do it. If performance *is* a concern -- which should be, if you're trying to capture packets from a DPDK application -- then the following approach is much better (in my opinion).

* Libpcap should provide support for pluggable capture backends.
* Each custom backend should deal with its own internal procedures to gather packets and expose them to the library users.

General reasoning:

* When libpcap requests for a capture start, it provides the BPF filter byte code to the application.
* If supported, the filter is compiled to native code.
* It gets a file descriptor in return and `mmap`s it to read captured packets from a lock-less ring (DPDK-less API, no multi-process constraints).
* Zero system call in the capture path. Dataplane workers simply apply the BPF filter and enqueue packet buffers to the shared ring.
* If the ring gets full because the pcap application cannot keep up, dataplane workers simply ignores the extraneous packets.

Proof of concept: https://github.com/DPDK/grout/pull/565

Let's continue discussion over there or at #1659

PR #1555, opened 2025-09-15
usb-linux: Support larger usbmon ring buffer limit

cyrozap:
I've recently submitted [a patch to the Linux kernel](https://lore.kernel.org/all/CAO3ALPxU5RzcoueC454L=WZ1qGMfAcnxm+T+p+9D8O9mcrUbCQ@mail.gmail.com/) to increase the maximum usbmon ring buffer size from 1200 KiB to 64 MiB. libpcap needs to be made aware of this larger limit in order for applications that use it to be able to capture larger, multi-MB URBs.

This PR adds that support.

### DIFF
--- a/pcap-usb-linux.c
+++ b/pcap-usb-linux.c
@@ -211,10 +211,15 @@ usb_findalldevs(pcap_if_list_t *devlistp, char *err_str)
 }
 
 /*
- * Matches what's in mon_bin.c in the Linux kernel.
+ * Matches what's in mon_bin.c in older Linux kernels.
  */
-#define MIN_RING_SIZE	(8*1024)
-#define MAX_RING_SIZE	(1200*1024)
+#define MIN_RING_SIZE		(8*1024)
+#define MAX_RING_SIZE_1200K	(1200*1024)
+
+/*
+ * The maximum ring size in newer kernels.
+ */
+#define MAX_RING_SIZE_64M	(64*1024*1024)
 
 static int
 usb_set_ring_size(pcap_t* handle, int header_size)
@@ -226,10 +231,10 @@ usb_set_ring_size(pcap_t* handle, int header_size)
 	 *  2) descriptors, for isochronous transfers;
 	 *  3) the payload.
 	 *
-	 * The kernel buffer has a size, defaulting to 300KB, with a
-	 * minimum of 8KB and a maximum of 1200KB.  The size is set with
-	 * the MON_IOCT_RING_SIZE ioctl; the size passed in is rounded up
-	 * to a page size.
+	 * The kernel buffer has a size, defaulting to 300 KiB, with a
+	 * minimum of 8 KiB and a maximum of 64 MiB (1200 KiB on older
+	 * kernels).  The size is set with the MON_IOCT_RING_SIZE ioctl;
+	 * the size passed in is rounded up to a page size.
 	 *
 	 * No more than {buffer size}/5 bytes worth of payload is saved.
 	 * Therefore, if we subtract the fixed-length size from the
@@ -248,18 +253,19 @@ usb_set_ring_size(pcap_t* handle, int header_size)
 
 	/*
 	 * Will this get an error?
-	 * (There's no way to query the minimum or maximum, so we just
-	 * copy the value from the kernel source.  We don't round it
-	 * up to a multiple of the page size.)
+	 * (There's no way to query the minimum or maximum, so we start
+	 * with the new maximum, and, if that fails with EINVAL, try
+	 * the old maximum.  We don't round it up to a multiple of the
+	 * page size.)
 	 */
-	if (ring_size > MAX_RING_SIZE) {
+	if (ring_size > MAX_RING_SIZE_64M) {
 		/*
 		 * Yes.  Lower the ring size to the maximum, and set the
 		 * snapshot length to the value that would give us a
 		 * maximum-size ring.
 		 */
-		ring_size = MAX_RING_SIZE;
-		handle->snapshot = header_size + (MAX_RING_SIZE/5);
+		ring_size = MAX_RING_SIZE_64M;
+		handle->snapshot = header_size + (MAX_RING_SIZE_64M/5);
 	} else if (ring_size < MIN_RING_SIZE) {
 		/*
 		 * Yes.  Raise the ring size to the minimum, but leave
@@ -270,12 +276,36 @@ usb_set_ring_size(pcap_t* handle, int header_size)
 		ring_size = MIN_RING_SIZE;
 	}
 
-	if (ioctl(handle->fd, MON_IOCT_RING_SIZE, ring_size) == -1) {
-		pcapint_fmt_errmsg_for_errno(handle->errbuf, PCAP_ERRBUF_SIZE,
-		    errno, "Can't set ring size from fd %d", handle->fd);
-		return -1;
+	if (ioctl(handle->fd, MON_IOCT_RING_SIZE, ring_size) == 0) {
+		/* It worked. */
+		return ring_size;
 	}
-	return ring_size;
+
+	/*
+	 * It failed.  If the error was EINVAL, and the ring size
+	 * was bigger than the old maximum, we may be on a kernel
+	 * that doesn't support ring sizes that large.
+	 */
+	if (errno == EINVAL && ring_size > MAX_RING_SIZE_1200K) {
+		/*
+		 * Try again with the old maximum.
+		 */
+		ring_size = MAX_RING_SIZE_1200K;
+		handle->snapshot = header_size + (MAX_RING_SIZE_1200K/5);
+		if (ioctl(handle->fd, MON_IOCT_RING_SIZE, ring_size) == 0) {
+			/* That worked. */
+			return ring_size;
+		}
+	}
+
+	/*
+	 * Either the first ioctl failed for some reason other than
+	 * EINVAL with a ring size bigger than the old maximum, or
+	 * the second one failed.  Report the error.
+	 */
+	pcapint_fmt_errmsg_for_errno(handle->errbuf, PCAP_ERRBUF_SIZE,
+	    errno, "Can't set ring size from fd %d", handle->fd);
+	return -1;
 }
 
 static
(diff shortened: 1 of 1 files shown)


infrastation:
Thank you for preparing this change. It would be very helpful to say in the comments which Linux kernel version is supposed to implement the increased buffer limit.

guyharris:
> Thank you for preparing this change. It would be very helpful to say in the comments which Linux kernel version is supposed to implement the increased buffer limit.

It was submitted today (2025-09-15), so it's more like "which Linux kernel version you expect to implement the increased buffer limit", without any guarantee that it'll *be* in that version (or if it'll be in *any* kernel version; nobody's commented on it yet, much less approved it yet).

The code forces the size to be >= the minimum size, so a `MON_IOCT_RING_SIZE` ioctl should fail with `EINVAL` only if the size is too big; the code first clamps the size at the new maximum and, if that gets `EINVAL` and the size is greater than the old maximum, clamps it at the old maximum and tries again, so it should work with either the old or new maximum.

guyharris:
This makes the maximum snapshot length for USB captures much larger.

Libpcap currently imposes the limit on the snapshot length it considers valid in capture file, in an attempt to avoid overly-large snapshot length causing code to eat a huge chunk of address space for the packet buffer; this might not be *too* much of an issue on 64-bit systems, other than maybe eating large chunks of *swap* space, but on a 32-bit system it could eat a large chunk of the *address* space, such that memory allocations after allocating the buffer fail.

Wireshark has a similar scheme for that.

My Brilliant Idea(TM) of making the maximum snapshot length be link-layer-type-dependent may have been well-intentioned, but I'm no longer sure it made any sense to just give a few link-layer types special dispensation, given that you're not guaranteed not to be reading a capture with one of those link-layer types.

The current maximum for D-Bus captures is the largest maximum we have now, and it's 128 MB, which is twice `MAX_RING_SIZE_64M`, and `MAX_RING_SIZE_64M` is 5 times the maximum snapshot length for USB captures in the modified code, so maybe libpcap and Wireshark should just set the maximum snapshot length accepted in capture files to 128 MB for now and stop bothering with checking the link-layer type.

cyrozap:
> > Thank you for preparing this change. It would be very helpful to say in the comments which Linux kernel version is supposed to implement the increased buffer limit.
> 
> It was submitted today (2025-09-15), so it's more like "which Linux kernel version you expect to implement the increased buffer limit", without any guarantee that it'll _be_ in that version (or if it'll be in _any_ kernel version; nobody's commented on it yet, much less approved it yet).

Yeah, I submitted this MR before getting anything merged in mainline for two reasons. First reason is, I had an earlier conversation with the USB subsystem maintainer where I proposed that this change be made and his response suggested he would likely accept a patch provided I can confirm that it actually works for my use case (which I did). The second is, once merged into mainline, it's going to take a _very_ long time before most users see the change in their kernel, so I wanted to make sure my patching efforts throughout the dependency tree (kernel, libpcap, and Wireshark) were all happening in parallel.

Incidentally, my patch was merged into the `usb-testing` branch [today](https://git.kernel.org/pub/scm/linux/kernel/git/gregkh/usb.git/commit/?h=usb-testing&id=368ed48a5ef52e384f54d5809f0a0b79ac567479) (2025-09-17).

> This makes the maximum snapshot length for USB captures much larger.
>
> ...

Correct, and I'm not actually 100% certain if the approach I've taken here is the correct one, so if you have a better idea, please let me know! It certainly feels very _weird_ to be copy-pasting and hard-coding kernel constants in userspace code. In theory, any future kernel could change those constants without warning (like the one that just got accepted into `usb-testing`), since they're unspecified in the usbmon userspace API, so ideally there would be some way to either query this limit or find it iteratively.

It also feels a bit weird to silently upgrade/downgrade the max buffer size depending on what libpcap detects the kernel supports, without telling the library user (since they now _also_ have to infer both what the kernel supports _and_ what libpcap supports). Granted, libpcap currently silently downgrades the buffer size anyways, so the way my patch works isn't a change from current behavior, but it's still a bit weird to me. But I don't know of any "smarter" way to do this without also changing the libpcap API, which I understand is supposed to be stable.

Also, to be perfectly honest, I have no idea what the point of forcing a maximum snaplen is, other than to preemptively enforce constraints of the kernel. And because I didn't understand the purpose of this limit, I retained this functionality in my patch. But I felt like it might be better to remove that extra limit it since the code that does the kernel buffer size detection _already_ enforces a maximum snaplen, and adding another limitation in the code feels arbitrary (and tripped me up when I was trying to figure out why my changes to `usb_set_ring_size` weren't working). That's not to say that the snaplen limit is bad--I literally just don't know the purpose or the design rationale, so if someone who knows the deep lore on this could let me know, I'd greatly appreciate it.

Finally, for context on my perspectives, I never use libpcap directly--I exclusively perform USB captures using Wireshark, primarily on 64-bit systems, but probably might also do so on some 32-bit ARM systems (mini PCs and Chromebooks, each with 1-4 GB of memory). And if memory usage is a concern, why not let the API user decide? Since I assume they're going to have a better idea of how much memory usage their system has available for capture buffers than libpcap can infer. Please note though that I have no idea how the API works since I've never used it, so maybe this is already the case?

Anyways, I'm mostly just thinking out loud here. And once again, if you have any suggestions on better ways to do what I'm trying to accomplish, I'll be happy to hear them!

guyharris:
> > This makes the maximum snapshot length for USB captures much larger.
> > ...
> 
> Correct, and I'm not actually 100% certain if the approach I've taken here is the correct one, so if you have a better idea, please let me know!

I don't have any better idea, "This makes the maximum snapshot length for USB captures much larger." means "libpcap and Wireshark will need to be changed as a result", not "it's a problem".



> It certainly feels very _weird_ to be copy-pasting and hard-coding kernel constants in userspace code. In theory, any future kernel could change those constants without warning (like the one that just got accepted into `usb-testing`), since they're unspecified in the usbmon userspace API, so ideally there would be some way to either query this limit

Yes - you might want to send a patch that adds new ioctls to query the minimum and maximum buffer sizes.

Oh, and *another* patch to move the ioctl codes and appropriate data structures from drivers/usb/mon/mon_bin.c and put them into a header file in include/uapi/linux/usb, so that all of this is *exported to userland* rather than having to be copied and pasted into code that uses the USB monitor stuff.

> It also feels a bit weird to silently upgrade/downgrade the max buffer size depending on what libpcap detects the kernel supports,

Libpcap does the same thing for the BPF capture mechanism on \*BSD/macOS/AIX/Solaris 11, so it's not as if this is something unique to the USB monitor code.

> without telling the library user (since they now _also_ have to infer both what the kernel supports _and_ what libpcap supports).

Why does the library user have to infer anything? The whole point of libpcap is to *hide* details such as that from the library user; most programs shouldn't need to set the buffer size, they should just let the library pick the default (if the default isn't appropriate for most captures, it should be fixed to be appropriate), with an API to request a different size if the default isn't big enough.

> Granted, libpcap currently silently downgrades the buffer size anyways,

As is the case when capturing on a \*BSD/macOS/AIX/Solaris 11 network interface.

> But I don't know of any "smarter" way to do this without also changing the libpcap API, which I understand is supposed to be stable.

`pcap_get_buffer_size(pcap_t *p)`

> Also, to be perfectly honest, I have no idea what the point of forcing a maximum snaplen is

To avoid capture files with a snapshot length that causes code that reads capture files to allocate a Really Huge Buffer, which on a 32-bit machine could succeed but not leave enough address space for the next allocation, and even on a 64-bit machine might either require allocation of a large chunk of swap space as backing store or cause some sort of overcommit error at run time.

> But I felt like it might be better to remove that extra limit it since the code that does the kernel buffer size detection _already_ enforces a maximum snaplen,

There are multiple parts of the code that handle kernel buffers - it's in most if no all files with pcap-XXX.c names. It's not unique to the USB monitor.

Note also that it's directly limiting the *snapshot length*, i.e. the maximum size of a single packet. How that affects the *buffer size* is platform-dependent and capture-device-dependent.

> And if memory usage is a concern,

That's the concern in programs reading the capture files; the kernel *itself* will presumably limit the kernel buffer size.

cyrozap:
I've just learned that my kernel changes have been [merged into mainline](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=368ed48a5ef52e384f54d5809f0a0b79ac567479), and we'll see the increased max buffer size in the 6.18 release! 🎉

@guyharris Thank you for answering my questions. Is there anything I need to change to get this PR merged? If big changes need to be made I'll be happy to make them, I just need to know the "what" and "where".

cyrozap:
I wasn't expecting this, but it seems the usbmon buffer limit increase [is getting backported to the stable kernels down to Linux 5.4](https://lore.kernel.org/all/20251025160905.3857885-304-sashal@kernel.org/). This means that change will appear in distro kernels much sooner than I expected, which is great news for ordinary users who want to capture USB traffic!

cyrozap:
As of November 13th, the increased buffer limit is available in stable kernel versions [6.12.58](https://lore.kernel.org/all/2025111334-liability-brush-8c0e@gregkh/) and [6.17.8](https://lore.kernel.org/all/2025111349-poker-undiluted-89b3@gregkh/).

cyrozap:
As of November 24th, the increased buffer limit is available in stable kernel version [6.6.117](https://lore.kernel.org/all/2025112429-cranberry-uncurious-d21f@gregkh/).

cyrozap:
As of December 7th, the increased buffer limit is available in stable kernel versions [5.4.302](https://lore.kernel.org/all/2025120319-blip-grime-93e8@gregkh/), [5.10.247](https://lore.kernel.org/all/2025120714-freeness-armless-d986@gregkh/), [5.15.197](https://lore.kernel.org/all/2025120721-remover-frosty-82ae@gregkh/), and [6.1.159](https://lore.kernel.org/all/2025120729-throwaway-rewash-7ebc@gregkh/). With these releases, the buffer limit increase is now available in all actively-supported stable kernels.

@guyharris and @infrastation, is there any chance this PR could be merged now? If any changes still need to be made to this PR to make it suitable to be merged, please let me know.

PR #1558, opened 2025-09-18
pcap-filter: Indicate that lldp and slow are not keywords

fxlb:
[skip ci]


infrastation:
The man page would also need changes to "backward compatibility" and the date stamp.

Issue #1538, opened 2025-07-29
Attempts to open a non-existent interface should always return "no such interface" even if you don't have capture permissions


guyharris:
See the-tcpdump-group/tcpdump#1334.

PR #1560, opened 2025-09-23
Make opening a non-existent interface on Linux return 'no such interface'

afshin-paydar:
- https://github.com/the-tcpdump-group/libpcap/issues/1538
- https://github.com/the-tcpdump-group/tcpdump/issues/1334

Fix interface existence check on Linux before privileged operations

  When attempting to open a non-existent network interface on Linux,
  libpcap would return PCAP_ERROR_PERM_DENIED instead of the correct
  PCAP_ERROR_NO_SUCH_DEVICE. This occurred because the privileged
  socket(PF_PACKET, SOCK_RAW, 0) call in setup_socket() would fail
  with permission errors before interface validation could occur.

  This fix adds an interface existence check using an unprivileged
  AF_INET socket and SIOCGIFINDEX ioctl before attempting to create
  the privileged packet socket. Non-existent interfaces now correctly
  return PCAP_ERROR_NO_SUCH_DEVICE.

  This resolves the issue where applications like tcpdump would report
  "Permission denied" instead of "No such device exists" for non-existent
  interfaces, breaking their fallback logic from interface names to indices.

  Note: BSD/macOS platforms have the same underlying issue where BPF device
  access fails with permission errors before interface validation. This
  should be addressed in a separate commit.

  Fixes: libpcap issue #1538
  Related: tcpdump issue #1334

```
./tcpdump -i nonexistent
----------------------------------------
tcpdump: nonexistent: No such device exists
(No such device exists)

./tcpdump -i eth999
----------------------------------------
tcpdump: eth999: No such device exists
(No such device exists)

./tcpdump -i 000000000000008
----------------------------------------
tcpdump: Invalid adapter index 8: only 7 interfaces found

./tcpdump -i 00000000000000073
----------------------------------------
tcpdump: Invalid adapter index 73: only 7 interfaces found
```

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -342,6 +342,7 @@ static const struct timeval netdown_timeout = {
 /*
  * Wrap some ioctl calls
  */
+static int	check_device_exists(const char *device, char *ebuf);
 static int	iface_get_id(int fd, const char *device, char *ebuf);
 static int	iface_get_mtu(int fd, const char *device, char *ebuf);
 static int	iface_get_arptype(int fd, const char *device, char *ebuf);
@@ -1199,6 +1200,28 @@ pcap_activate_linux(pcap_t *handle)
 	}
 
 	/*
+	 * The "any" device is a special device which causes us not
+	 * to bind to a particular device and thus to look at all
+	 * devices.
+	 */
+	is_any_device = (strcmp(device, "any") == 0);
+
+	if (!is_any_device) {
+		/*
+		 * Check whether the device exists, so that we report
+		 * PCAP_ERROR_NO_SUCH_DEVICE for a non-existent device
+		 * rather than an error from some later operation, such
+		 * as PCAP_ERROR_PERM_DENIED if we don't have permission
+		 * to create a PF_PACKET socket.
+		 */
+		ret = check_device_exists(device, handle->errbuf);
+		if (ret < 0) {
+			status = ret;
+			goto fail;
+		}
+	}
+
+	/*
 	 * Turn a negative snapshot value (invalid), a snapshot value of
 	 * 0 (unspecified), or a value bigger than the normal maximum
 	 * value, into the maximum allowed value.
@@ -1217,12 +1240,6 @@ pcap_activate_linux(pcap_t *handle)
 		goto fail;
 	}
 
-	/*
-	 * The "any" device is a special device which causes us not
-	 * to bind to a particular device and thus to look at all
-	 * devices.
-	 */
-	is_any_device = (strcmp(device, "any") == 0);
 	if (is_any_device) {
 		if (handle->opt.promisc) {
 			handle->opt.promisc = 0;
@@ -1765,6 +1782,40 @@ get_if_ioctl_socket(void)
 	 * types popular in non-Internet embedded systems?
 	 */
 	return (socket(AF_INET, SOCK_DGRAM, 0));
+}
+
+/*
+ * Check whether a device exists, so that we can report
+ * PCAP_ERROR_NO_SUCH_DEVICE for a non-existent device before
+ * attempting any operation that requires privileges the caller
+ * might not have, so that the caller gets "no such device" rather
+ * than, for example, "you don't have permission to capture".
+ *
+ * Returns 0 if the device exists and a PCAP_ERROR_ code, filling
+ * in ebuf, if it doesn't exist or if the attempt to check fails.
+ */
+static int
+check_device_exists(const char *device, char *ebuf)
+{
+	int	fd;
+	int	status;
+
+	fd = get_if_ioctl_socket();
+	if (fd < 0) {
+		pcapint_fmt_errmsg_for_errno(ebuf, PCAP_ERRBUF_SIZE,
+		    errno, "socket");
+		return PCAP_ERROR;
+	}
+
+	/*
+	 * This fails with ENODEV, and thus returns
+	 * PCAP_ERROR_NO_SUCH_DEVICE, if the device doesn't exist.
+	 */
+	status = iface_get_arptype(fd, device, ebuf);
+	close(fd);
+	if (status < 0)
+		return status;
+	return 0;
 }
 
 /*


infrastation:
Thank you for preparing this change. The comments that clarify the problem and the solution should be in the commit message.

pcap-bpf.c
```diff
@@ -626,6 +637,66 @@ bpf_open(char *errbuf)
...
+			    save_errno, "SIOCGIFFLAGS on %s", name);
+			return PCAP_ERROR;
+		}
+	}
+#else
+	/*
+	 * SIOCGIFFLAGS not available on this platform.
+	 * We can't reliably check interface existence without privileges,
+	 * so we skip the check and let the BPF bind operation handle it.
+	 */
```
infrastation: Which platform(s) do not have `SIOCGIFFLAGS`? (It would be useful to give an exact list in the comment.) If there is such a platform, it would be right to put the `ifr` variable into the `#ifdef` block because outside of it it is effectively unused.

pcap-linux.c
```diff
@@ -5164,13 +5164,13 @@ iface_get_ts_types(const char *device, pcap_t *handle, char *ebuf)
...
-			 * OK, no such device.
-			 * The user will find that out when they try to
-			 * activate the device; just return an empty
-			 * list of time stamp types.
+			 * No such device.
+			 *
+			 * There's nothing more to say, so clear the
+			 * error message.
 			 */
-			handle->tstamp_type_list = NULL;
```
infrastation, line 5185: Why `tstamp_type_list` no longer needs to be set here?

pcap-bpf.c
```diff
@@ -41,6 +41,17 @@
...
 #include <sys/ioccom.h>
 #endif
+
+/*
+ * On most supported platforms <sys/ioctl.h> also defines the SIOCGIF* macros.
+ * However, on Haiku, illumos and Solaris the macros need <sys/sockio.h>,
+ * which does not exist in AIX 7, HP-UX 11, GNU/Hurd and Linux (both GNU and
+ * musl libc).
+ */
+#if defined(HAVE_SOLARIS) || defined(__HAIKU__) || defined(__sun) || defined(__SVR4)
```
infrastation: Why does this condition need `__sun` or `__SVR4`?

pcap-bpf.c
```diff
@@ -626,6 +637,66 @@ bpf_open(char *errbuf)
...
+	/*
+	 * SIOCGIFFLAGS not available on this platform.
+	 * We can't reliably check interface existence without privileges,
+	 * so we skip the check and let the BPF bind operation handle it.
+	 */
+#endif
+
+	close(fd);
+	return 0;
+#else
```
infrastation: The short branch of the `#if` should be the first to make it easier to follow the logic. Also the nested `#if` block should be indented and both `#endif` lines should have a comment to show which block terminates where. Please see `nametoaddr.c` for an example of better style.

infrastation:
On Linux the only proposed change is to `iface_get_ts_types()`, thus the "check for existence before trying a privileged operation" logic does not apply to Linux:
```
$ ./testprogs/activatetest 
Trying to use capture device "nosuchdevice"...
FAIL: Permission denied from pcap_activate(), retry with higher privileges.
```
It looks like the two proposed changes are not related and should be two different commits, each stating and solving a separate problem. Also, since `pcap-linux.c` already implements helper functions such as `iface_get_id()`, it would be easy to implement the same logic for Linux. That said, the logic increases the race condition window, which would require some attention.

**This requires a bit more work before it is ready.**

pcap-linux.c
afipay: Thank you for reviewing these changes
because current code ignoring the ENODEV error and treating it as a success case, assumes the user will discover the real problem later during activation

```
  case ENODEV:
      ..
      handle->tstamp_type_list = NULL;
      return 0;  // SUCCESS, continue execution
```

This is problematic because it allows execution to continue with a non existent interface, leading to confusing error messages later

The new code immediately fails when it detects ENODEV:
```
  case ENODEV:
      ...
      ebuf[0] = '\0';
      return PCAP_ERROR_NO_SUCH_DEVICE;  // MMEDIATE FAILURE
```

after returning PCAP_ERROR_NO_SUCH_DEVICE, the calling function (`pcapint_create_interface`) do:
```
  if (iface_get_ts_types(device, handle, ebuf) == -1) {
      pcap_close(handle);  // Cleans up the handle
      return NULL;         // Function exits immediately
  }
```
Since the function exits immediately and calls `pcap_close(handle)`, there's no point in setting `tstamp_type_list`

pcap-bpf.c
afipay: removed extra `__sun` and `__SVR4` conditions were unnecessary

pcap-bpf.c
afipay: Removed unnecessary `#ifdef SIOCGIFFLAGS`

afipay: reverted this part, as it's not in the issue

pcap-linux.c
```diff
@@ -4812,6 +4824,50 @@ pcap_setfilter_linux(pcap_t *handle, struct bpf_program *filter)
...
+static int
+iface_exists(const char *device, char *ebuf)
+{
+	int fd;
+	struct ifreq ifr;
+
+	if (strlen(device) >= sizeof(ifr.ifr_name)) {
+		ebuf[0] = '\0';
+		return PCAP_ERROR_NO_SUCH_DEVICE;
+	}
```
infrastation: What is the purpose of this check here, given `pcap_activate_linux()` makes this check before invoking `setup_socket()`, which is the only code path into `iface_exists()`?

pcap-linux.c
```diff
@@ -4812,6 +4824,50 @@ pcap_setfilter_linux(pcap_t *handle, struct bpf_program *filter)
...
+
+		if (save_errno == ENODEV) {
+			ebuf[0] = '\0';
+			return PCAP_ERROR_NO_SUCH_DEVICE;
+		} else {
+			pcapint_fmt_errmsg_for_errno(ebuf, PCAP_ERRBUF_SIZE,
+			    save_errno, "SIOCGIFINDEX on %s", device);
+			return PCAP_ERROR;
+		}
+	}
```
infrastation: This would be the same as invoking the existing `iface_get_id()` and checking `errno` afterwards, correct?

infrastation: Furthermore, the existing `iface_get_arptype()`, would check for `ENODEV`, as it seems from the code.

pcap-linux.c
infrastation: Please note that this explanation does not match the code: if the modified `iface_get_ts_types()` returns `PCAP_ERROR_NO_SUCH_DEVICE` (which is defined to -5), the test for -1 is false because -1 != -5 and there is no immediate return.

To add to that, modifying the immediate return condition in `pcapint_create_interface()` to include `PCAP_ERROR_NO_SUCH_DEVICE` would break things. The only way for `pcapint_create_interface()` to indicate a failure is to return `NULL`. To indicate a specific reason for a failure in a reliable way, it takes to get a `PCAP_ERROR_NO_SUCH_DEVICE` from `pcap_activate_linux()`, which requires that `pcapint_create_interface()` does not fail, which requires that `iface_get_ts_types()` does not fail, hence the current arrangement to defer the failure on `ENODEV`.

Therefore please leave `iface_get_ts_types()` intact.

infrastation (changes requested):
As noted.

pcap-linux.c
afipay: Thanks for explanation, it's reverted

pcap-linux.c
afipay: Correct, removed redundant check

pcap-linux.c
afipay: Yes, reused the existing code

fxlb:
Rebased.

pcap-linux.c
```diff
@@ -2439,6 +2439,30 @@ setup_socket(pcap_t *handle, int is_any_device)
 	int			err = 0;
 	struct packet_mreq	mr;
 
+	/*
```
guyharris: I'd put the routine that does this check into a separate routine, and perhaps call it earlier in `pcap_activate_linux()`.

afipay: Thanks for review, moved it into a separate routine and call it early in `pcap_activate_linux()`

guyharris:
Should the check whether a device exists be done before trying to bind it, or after the attempt to bind the device fails with a permission error?

There's a time-of-check/time-of-use issue for both of those, but, if interfaces are appearing or disappearing, there's a risk of an error being reported for a condition that's not true by the time the user sees the error no matter *what* the code does, so I'm not bothered by that.

The one advantage of doing it afterwards is that fewer system calls are made in the case where everything succeeds, i.e. if the program *does* have permission to capture. I'm not sure that's a big enough difference to worry about.

infrastation:
Make it work, make it right, make it fast?

pcap-linux.c
guyharris: Looks good.

PR #1562, opened 2025-10-09
Add the size of time_t to the version string.

guyharris:
This makes it easier to detect time_t size mismatches between libpcap and code that uses libpcap, at least with code that reports the version.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -153,6 +153,7 @@ DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
       gencode: Fix an undefined behavior in gen_mcode().
       gencode: Add a missing free() in gen_scode().
       Remove "DLT_" from the descriptions of two dlt_choices[] entries.
+      Report the size of time_t in the version string.
     Source code:
       Remove some unneeded includes.
       pcapint_find_function() changed to return "void *" to avoid
--- a/Makefile.in
+++ b/Makefile.in
@@ -368,6 +368,7 @@ EXTRA_DIST = \
 	testprogs/threadsignaltest.c \
 	testprogs/unix.h \
 	testprogs/valgrindtest.c \
+	testprogs/versiontest.c \
 	testprogs/visopts.py \
 	testprogs/writecaptest.c
 
--- a/build.sh
+++ b/build.sh
@@ -147,12 +147,14 @@ run_after_echo "$PREFIX/bin/pcap-config" --additional-libs --static-pcap-only
 # VALGRIND_CMD is meant either to collapse or to expand.
 # shellcheck disable=SC2086
 if [ "$CMAKE" = no ]; then
+    run_after_echo $VALGRIND_CMD testprogs/versiontest
     FILTERTEST_BIN="$VALGRIND_CMD testprogs/filtertest"
     export FILTERTEST_BIN
     run_after_echo "$MAKE_BIN" -s check
     run_after_echo $VALGRIND_CMD testprogs/findalldevstest
     [ "$TEST_RELEASETAR" = yes ] && run_after_echo "$MAKE_BIN" releasetar
 else
+    run_after_echo $VALGRIND_CMD run/versiontest
     FILTERTEST_BIN="$VALGRIND_CMD run/filtertest"
     export FILTERTEST_BIN
     run_after_echo "$MAKE_BIN" -s check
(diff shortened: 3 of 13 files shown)


infrastation:
```
Compiled with libpcap version 1.11.0-PRE-GIT, 64-bit time_t (with zerocopy and netmap support)
```
It may be better to place the additional text inside the parentheses because that's where the unstructured non-version-number text appears, this way I estimate a typical version-matching regexp in a 3rd-party program would be less likely to break. This would be a bit convoluted because the text in parentheses originates from several different places.

In any case, this change has to be mentioned in the change log because it changes an observed behaviour.

fxlb:
> It may be better to place the additional text inside the parentheses

Not sure because inside the parentheses is the capture type specific part. time_t size is not capture type specific.

guyharris:
> Not sure because inside the parentheses is the capture type specific part. time_t size is not capture type specific.

The text inside the parentheses *currently* happens only to contain capture-type-specific information.  There's no *inherent* reason why that should be the case.

Perhaps we should also introduce an API to get some information about libpcap, such as the version number - in *numeric* form, giving the major version, minor version, and dot-dot-release version - and perhaps also including the size of `time_t`. It could perhaps also include a `const char *` pointing to a string giving the capture-type-specific information. This would be used if code wants to get that information for *its* use, rather than for display to the user.

infrastation:
As can be inferred from appneta/tcpreplay#750, the comma after the number in a release version could break things unexpectedly.

guyharris:
> In any case, this change has to be mentioned in the change log because it changes an observed behaviour.

I've added it, for now, to the 1.10.6 section, as the  tcpdump 4.99 branch adds the pointer size and `time_t` size to the `tcpdump --version` output.

infrastation:
Thank you. There is a space missing:
```
Compiled with libpcap version 1.11.0-PRE-GIT(64-bit time_t)
```

guyharris:
> There is a space missing:

Fixed, but there should really be a macro that generates a proper version string when passed a string of capture-type-specific information, so the implementations of `pcap_lib_version()` are less fragile.

guyharris:
> there should really be a macro that generates a proper version string when passed a string of capture-type-specific information, so the implementations of `pcap_lib_version()` are less fragile.

Done.

infrastation:
Regarding the means of getting libpcap version in an integer (and easily comparable) rather than string form, the last discussion about it was [this](https://lists.tcpdump.org/mailman3/hyperkitty/list/tcpdump-workers@lists.tcpdump.org/thread/BDSTWYHCCQZSUWTWVY4EZV2RYFI3LF45/) about 1.5 years ago. Would anybody be interested to prototype some code along these lines?

guyharris:
> Regarding the means of getting libpcap version in an integer (and easily comparable) rather than string form, the last discussion about it was [this](https://lists.tcpdump.org/mailman3/hyperkitty/list/tcpdump-workers@lists.tcpdump.org/thread/BDSTWYHCCQZSUWTWVY4EZV2RYFI3LF45/) about 1.5 years ago. Would anybody be interested to prototype some code along these lines?

That's mostly discussing getting configuration information at configure time;; the API I was mentioning is for programs to get version info from libpcap at run time, in some form other than a string that needs parsing.

infrastation:
There is more complexity to the version handling than I remember. `make releasecheck` does not pass.

guyharris:
> There is more complexity to the version handling than I remember. `make releasecheck` does not pass.

That's more like "complexity to adding a new test program" - it has to be added to the release tarball list as well as the testprogs Makefile.in.

infrastation:
I suppose `versiontest.c` is where the future compile-time and run-time libpcap version checks are going to be exercised, correct?

infrastation:
Anyway, this looks good enough to merge, then other improvements can be made later as and if required.

guyharris:
> I suppose `versiontest.c` is where the future compile-time and run-time libpcap version checks are going to be exercised, correct?

The immediate goal was to print out the version bring in builds so that I could check them by hand; a future goal is, indeed, to do automated checks on its output and thus on the version-information-fetching APIs, e.g. making sure, at minimum, that it gets the version number right and has a non-empty parenthesized list of characteristics after it.

But those aren't checks on the libpcap version itself; those checks would be done by programs that use libpcap, either in their configuration process for compile-time checks or in their code for run-time checks.

guyharris:
> Anyway, this looks good enough to merge, then other improvements can be made later as and if required.

Yup.

fxlb:
Useful, thank you!

PR #1564, opened 2025-10-15
DECnet (again) and test gaps

infrastation:
These changes concern a number of similar code paths, which is why it took longer than I expected to enumerate the bugs and to make sure the tests cover the problem space. But here it is.

### DIFF
--- a/testprogs/TESTrun
+++ b/testprogs/TESTrun
@@ -14201,6 +14201,17 @@ sub errstr_nohost_af {
 	return sprintf ('unknown host \'%s\' for specified address family', shift);
 }
 
+# gen_scode()
+# gen_ncode()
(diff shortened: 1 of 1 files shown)


infrastation:
Apparently, the CI failures on illumos-amd64 are due to DNS delay:
```
Failed tests:
    reject_arp_host_nonexistent             : filtertest timeout
    reject_ip6_host_nonexistent             : filtertest timeout
    reject_ip_host_nonexistent              : filtertest timeout
    reject_rarp_host_nonexistent            : filtertest timeout
```
The host occasionally takes more than one second to fail resolving `host.invalid`. This reproduces using the current master branch and is not specific to the proposed changes. This pull request is going to be merged tomorrow.

Issue #1571, opened 2025-10-24
the "gateway" primitive correctly implements an incorrect definition


infrastation:
The master branch now has a few "apply" filter tests for `gateway`, which demonstrate that the current implementation is way less useful than the documentation leads to think.  In particular, the man page says:
```
       gateway host
              True  if the packet used host as a gateway.  I.e., the source or
              the destination Ethernet-like address is host  but  neither  the
              source  nor the destination ARP/IPv4/RARP address is host.  This
              primitive is valid only for the same  link-layer  types  as  the
              ether  host  primitive  above.  May be qualified with a specific
              protocol (arp, ip, rarp).  For example,
                   ip gateway host
              is equivalent to
                   ether host ethernameaddr and not ip host hostnameaddr
              which can be used with either names or numbers for  hostnameaddr
              and ethernameaddr.
```

In detail:
* `ip gateway HOST` means: (the source or the destination MAC address matches) and ((the link-layer protocol is IPv4 and neither the source nor the destination IPv4 address matches) or (**the link-layer protocol is not IPv4**))
* `arp gateway HOST` means: (the source or the destination MAC address matches) and ((the link-layer protocol is ARP and neither the source nor the destination ARP address matches) or (**the link-layer protocol is not ARP**))
* `rarp gateway HOST` means: (the source or the destination MAC address matches) and ((the link-layer protocol is RARP and neither the source nor the destination RARP address matches) or (**the link-layer protocol is not RARP**))
* `gateway HOST` means: (the source or the destination MAC address matches) and ((the link-layer protocol is IPv4 and neither the source nor the destination IPv4 address matches) or (the link-layer protocol is ARP and neither the source nor the destination ARP address matches) or (the link-layer protocol is RARP and neither the source nor the destination RARP address matches) or (**the link-layer protocol is neither IPv4 nor ARP nor RARP**))

The root cause of this confusion seems to originate in the grammar: for example, `not ip src host 1.2.3.4` means exactly the same as `ip src host not 1.2.3.4`, but a human tends to understand the former as "all packets except those that are IPv4 and have the specified source address" (correct) and the latter as "all IPv4 packets except those that have the specified source address" (incorrect). With this in mind, the informal part of the definition (_True  if the packet used host as a gateway._) makes perfect sense, but the formal part needs to be different. If the correct solution is to match the protocol separately from the address (not tested!), then:
[...]

PR #1573, opened 2025-10-27
Produce fewer instructions in gen_and() and gen_or().

infrastation:
Recently I attempted adding IPv6 support to the `gateway` primitive, which did not work in the expected trivial manner, but instead produced a number of useful by-products. One of those is bug report #1571, another is this series of commits. It includes a more reliable revision of the commit I earlier removed from pull request #1498 (_Have gen_and(), gen_or() and gen_not() return the result._), which in turn enables an early optimization of logical AND and OR as discussed in bug report #1022 — the optimization reduces that problem space a little bit. Another optimization (_Do not special-case BPF_JEQ in gen_relation_internal()._) addresses a different root cause, but produces comparable results.

One of the main complexities here was to compose the tests right to make sure they cover the affected code paths, but at the same time do not inflate the combinatoric space too much. Another was to organize the changes into reasonably focused commits, which after many rounds of clean-ups looks about right. One exception is the 40 new test blocks, which look very messy in the diff, even though they mostly add new Perl code.

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -4592,16 +4592,13 @@ gen_hostop6(compiler_state_t *cstate, struct in6_addr *addr,
 	/* this order is important */
 	memcpy(a, addr, sizeof(a));
 	memcpy(m, mask, sizeof(m));
-	b1 = NULL;
+	b1 = gen_true(cstate);
 	for (int i = 3; i >= 0; i--) {
-		// Same as the Q_IP case in gen_host().
-		if (m[i] == 0 && a[i] == 0)
-			continue;
 		b0 = gen_mcmp(cstate, OR_LINKPL, offset + 4 * i, BPF_W,
 		    ntohl(a[i]), ntohl(m[i]));
-		b1 = b1 ? gen_and(b0, b1) : b0;
+		b1 = gen_and(b0, b1);
 	}
-	return b1 ? b1 : gen_true(cstate);
+	return b1;
 }
 
 /*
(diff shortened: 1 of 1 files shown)


infrastation:
One cause of the CI failures is a switch on an enum, another is a missing skip condition in a number of test blocks, both are known and trivial to address. Yet another cause is now bug report #1574, which blocks this pull request.

infrastation:
This revision contains a workaround to process IPv4 before IPv6, let's see if this is the only thing required to pass all tests.

infrastation:
This rebased revision passes all tests, one commit has been merged into the master branch. This is going to be merged in a few days.

infrastation:
Rebased on the current master branch. I would like to try making the messy diff commit look better before merging.

infrastation:
This revision has the previously messy diff readable and incorporates a few cosmetic improvements. It looks ready to me and is going to be merged tomorrow.

infrastation:
Had to fix a commit message, this is going to be merged soon after the CI has passed.

infrastation:
As it turns out, `geneve and tcp` has broken in tcpdump tests, which shows it ought to be a libpcap apply filter test instead. The same stands for `geneve 0xb`, which still works.

infrastation:
The regression concerns GENEVE and VXLAN only. My working copy has a bug fix, it needs better comments and better tests before committing.

infrastation:
Also there is another bug, for which there is a separate work-in-progress bug fix.

PR #1577, opened 2025-10-31
fix remote capture starting bug

zhao85:
fix pcap_startcapture_remote() command length incorrect when fp->fcode.bf_len==0


infrastation:
Please describe the problem and the steps to reproduce it. The solution probably requires a better comment in the commit message and/or the source code.

fxlb:
Please remove your merge commit, use rebase.

PR #1584, opened 2025-11-15
Standardize some error messages (bogus->invalid)

fxlb:


scanner.l
```diff
@@ -396,14 +396,14 @@ ${B}			{ yylval->s = sdup(yyextra, yytext); return AID; }
...
-				bpf_set_error(yyextra, "bogus IPv6 address %s", yytext);
+				bpf_set_error(yyextra, "invalid IPv6 address %s", yytext);
 				yylval->s = NULL;
 			  } else {
 				yylval->s = sdup(yyextra, (char *)yytext);
 			  }
 			  return HID6;
 			}
-{B}:+({B}:+)+		{ bpf_set_error(yyextra, "bogus ethernet address %s", yytext); yylval->s = NULL; return EID; }
+{B}:+({B}:+)+		{ bpf_set_error(yyextra, "invalid ethernet address %s", yytext); yylval->s = NULL; return EID; }
```
guyharris, line 406: Should this capitalize "Ethernet"?

testprogs/TESTrun
```diff
@@ -18633,7 +18633,7 @@ my @reject_tests = (
 		name => 'ether_host',
 		DLT => 'EN10MB',
 		expr => 'ether ab:cd:ef:0g:00:00',
-		errstr => 'bogus ethernet address',
+		errstr => 'invalid ethernet address',
```
guyharris, line 18636: Ditto.

scanner.l
guyharris: Perhaps, but that should be done as a separate change.

testprogs/TESTrun
guyharris: Perhaps, but that should be done as a separate change.

guyharris:
For the AppVeyor build failures, see #1583.

scanner.l
fxlb: > Should this capitalize "Ethernet"?

Yes, See #1585.

PR #1589, opened 2025-11-18
cmake: attempt to work around the problem with VS 2019 ARM64 builds.

guyharris:


CMakeLists.txt
```diff
@@ -370,6 +370,14 @@ if(MSVC)
     if(NOT ${CMAKE_C_COMPILER} MATCHES "clang*")
         set(C_ADDITIONAL_FLAGS "${C_ADDITIONAL_FLAGS} /utf-8")
     endif(NOT ${CMAKE_C_COMPILER} MATCHES "clang*")
+
+    #
+    # Show the Windows SDK version, if defined.
+    #
+    if(DEFINED ${CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION})
+        message(STATUS "Using Windows SDK ${CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION}")
+    fi
```
infrastation: `endif()` ?

guyharris: Yeah, my Bourne shell neurons kicked in at that point. Fixed.

guyharris:
[The Npcap Changelog](https://npcap.com/changelog) says, in the Npcap 1.80 section:

> * Upgraded build system to Visual Studio 2022. This version prevents building drivers for earlier Windows versions, so the npcap.sys driver for Windows 7 through 8.1 is still built on Visual Studio 2019. See [#629](http://issues.npcap.org/629).

Neither Windows 7 nor Windows 8.1 support ARM64, as far as I know, so I'm not sure building the ARM64 code needs to be tested on earlier versions of Visual Studio.

PR #1593, opened 2025-12-05
Add gsw1xx mdio link support

peteraxis:
This is adding the MaxLinear dsa switch protocol.

v2: Add offset for finding ethernet header. Renamed to "Special Tag".

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -4458,7 +4458,7 @@ gen_llc_linktype(compiler_state_t *cstate, bpf_u_int32 ll_proto)
 	default:
 		/*
 		 * XXX - we don't have to check for IPX 802.3
-		 * here, but should we check for the IPX Ethertype?
+		 * here, but should we check for the IPX EtherType?
 		 */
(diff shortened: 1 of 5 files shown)


gencode.c
```diff
@@ -1747,6 +1747,13 @@ init_linktype(compiler_state_t *cstate, pcap_t *p)
 		cstate->off_nl_nosnap = 3;	/* 802.3+802.2 */
 		break;
 
+	case DLT_DSA_TAG_GSW1XX:
+            cstate->off_linktype.constant_part = 20;
+		cstate->off_linkpl.constant_part = 22;	/* Ethernet header length + Special Tag Length */
+		cstate->off_nl = 0;		/* Ethernet II */
+		cstate->off_nl_nosnap = 3;	/* 802.3+802.2 */
+		break;
+
```
infrastation, line 1756: Packet filtering is a bit more complicated than that (see #1195), but it does not have to be implemented right at the start. I suggest to attempt filtering as one of the next steps.

Issue #1511, opened 2025-04-30
Problem building rpcapd on Solaris 11 SPARC, Studio compiler, in some case (64-bit, CMAKE=yes, REMOTE=yes)


fxlb:
Fails with 'ldd rpcapd' or 'rpcapd -h':
ldd: /tmp/.../sbin/rpcapd: execution failed due to signal 11 (core dumped)
If print_so_deps() call commented (ldd run):
$ /tmp/.../sbin/rpcapd -h
./build.sh: line 294: 18232: Memory fault(coredump)

Found via:
https://buildfarm.opencsw.org/buildbot_admin/builders/libpcap-solaris11-sparcv9/builds/3/steps/shell/logs/stdio
https://buildfarm.opencsw.org/buildbot_admin/builders/libpcap-solaris11-sparcv9/builds/4/steps/shell/logs/stdio

Reproduce when buiding 64-bit, CMAKE=yes, REMOTE=yes.
(Autoconf: OK)

e.g.
CC=/opt/developerstudio12.5/bin/cc CFLAGS=-m64 LDFLAGS=-m64 LIBPCAP_TAINTED=yes CMAKE=yes REMOTE=yes ./build.sh

Failed:
/opt/SUNWspro/prod/bin/cc
/opt/developerstudio12.5/bin/cc
/opt/developerstudio12.6/bin/cc
/opt/solarisstudio12.3/prod/bin/cc
/opt/solarisstudio12.4/bin/cc
/opt/solstudio12.2/prod/bin/cc

OK:
gcc

infrastation:
This could be a problem of Solaris and/or the Studio rather than the binary.

fxlb:
@dago Hello, if you have an idea...

fxlb:
Recent log (old are gone):
https://buildfarm.opencsw.org/buildbot/builders/libpcap-solaris11-sparcv9/builds/78/steps/shell/logs/stdio

guyharris:
To quote the comment at the beginning of Illumos's ldd.c, which probably reflects the Solaris code (which probably dates back to SunOS 4.0 or 4.1 - I forget whether it shipped in 4.0 or was added in 4.1):

```
/*
* Print the list of shared objects required by a dynamic executable or shared
* object.
*
* usage is: ldd [-d | -r] [-c] [-D] [-e envar] [-i] [-f] [-L] [-l] [-p] [-s]
*		[-U | -u] [-v] [-w] file(s)
*
[...]

PR #1594, opened 2025-12-11
Draft: fix Solaris 11 64-bit SPARC CMake build.

guyharris:
At this point, this is just doing verbose builds, so we can see whether the autotools and CMake builds pass different compiler arguments.

See #1511.


guyharris:
This is not provoking OpenCSW builds, so I'm going to close it and try again. DO NOT REOPEH.

PR #1595, opened 2025-12-11
Fix Solaris 11 64-bit SPARC CMake build.

guyharris:
At this point, this is just doing verbose builds, so we can see whether the autotools and CMake builds pass different compiler arguments.

See #1511.


guyharris:
This also didn't provoke a build on OpenCSW.  Closing.

guyharris:
DO NOT REOPEN.

PR #1596, opened 2025-12-16
Stub out all the rpcap-URL-related APIs if no remote capture.

guyharris:
The code needs a lot of cleanup, due o a number of bugs found in it.

For now, at least, let's just stub it out, with routines that always fail and provide a "this isn't supported" error string, if ENABLE_REMOTE isn't defined.

Move routines back into pcap-new.c; they are routines provided for 1) the ability to open and enumerate remote devices and 2) compatibility with WinPcap for access to local devices and capture files, but those aren't APIs we want to recommend.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1884,6 +1884,7 @@ set(PROJECT_SOURCE_LIST_C
     pcap-common.c
     pcap-util.c
     pcap.c
+    pcap-new.c
     savefile.c
     sf-pcapng.c
     sf-pcap.c
--- a/Makefile.in
+++ b/Makefile.in
@@ -87,7 +87,7 @@ PLATFORM_C_SRC =	@PLATFORM_C_SRC@
 MODULE_C_SRC =		@MODULE_C_SRC@
 REMOTE_C_SRC =		@REMOTE_C_SRC@
 COMMON_C_SRC =	pcap.c gencode.c optimize.c nametoaddr.c etherent.c \
-		fmtutils.c pcap-util.c \
+		fmtutils.c pcap-util.c pcap-new.c \
 		savefile.c sf-pcap.c sf-pcapng.c pcap-common.c \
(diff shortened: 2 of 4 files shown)


fxlb:
The file named “pcap-new.c” appeared in 2016 and was deleted in 2023.
Should we choose a more meaningful name for its return?

guyharris:
> Should we choose a more meaningful name for its return?

Yes.

In addition, the APIs for parsing and constructing URLs should be moved there, as:

1. they're only useful for `pcap_findalldevs_ex()` and `pcap_open()` - in the longer term, I'd like to have better APIs for enumerating interfaces and allow remote capture without `pcap_open()`;
2. the parser fills in *fixed-length buffers* with URL components, which is Not A Good Idea;
3. I'm unconvinced that using file: URLs for capture files was the right thing to do;

so I don't want to promote them as the right way to go.

My inclination is to merge this now, and then do the renaming and moving after Denis has checked in fixes for the problems he's found with various of those routines, as those will probably be committed both to the main branch *and* the 1.10 branch (given tha the 1.10 branch versions are being used in Npcap).

infrastation:
No objections, this does not block me from adding the bug fixes into libpcap-1.10.

guyharris:
> No objections, this does not block me from adding the bug fixes into libpcap-1.10.

I was thinking this change might make it easier to cherry-pick them from the main branch to the 1.10 branch, by reducing the differences between them.

fxlb:
> My inclination is to merge this now, and then do the renaming and moving [...]

OK.

guyharris:
Any suggestions for the new name? pcap-wpcap-compat.c, as, in the long term, I hope to have better APIs that we document and recommend, with these kept around for backwards compatibility with code written for the WinPcap APIs?

infrastation:
If there is a better naming convention, it is not obvious to me.

fxlb:
No idea of a better name.

PR #1597, opened 2025-12-16
FreeBSD USB: remove interface creation for FreeBSD 16.0

glebius:
In FreeBSD 16.0 the USB bpf(4) taps are always available and the creation of pseudo-interfaces, that used to exist merely for a bpf(4) tap, is neither required nor is possible.  The list of all tap points can be obtained with BIOCGETIFLIST.  Note that we still use pcapint_findalldevs_interfaces() as the main method of collecting bpf taps, and we use BIOCGETIFLIST only to fill the gaps there - get those bpf taps, that are not a network interface.

### DIFF
--- a/pcap-bpf.c
+++ b/pcap-bpf.c
@@ -43,7 +43,7 @@
 #endif
 #include <sys/utsname.h>
 
-#if defined(__FreeBSD__) && defined(SIOCIFCREATE2)
+#if defined(__FreeBSD__) && __FreeBSD_version < 1600006
 /*
  * Add support for capturing on FreeBSD usbusN interfaces.
  */
@@ -1736,7 +1736,7 @@ pcap_cleanup_bpf(pcap_t *p)
 		}
 #endif /* HAVE_BSD_IEEE80211 */
 
-#if defined(__FreeBSD__) && defined(SIOCIFCREATE2)
+#if defined(__FreeBSD__) && __FreeBSD_version < 1600006
 		/*
 		 * Attempt to destroy the usbusN interface that we created.
 		 */
@@ -1753,7 +1753,7 @@ pcap_cleanup_bpf(pcap_t *p)
 				}
 			}
 		}
-#endif /* defined(__FreeBSD__) && defined(SIOCIFCREATE2) */
+#endif /* defined(__FreeBSD__) && __FreeBSD_version < 1600006 */
 		/*
 		 * Take this pcap out of the list of pcaps for which we
 		 * have to take the interface out of some mode.
@@ -2049,11 +2049,11 @@ pcap_activate_bpf(pcap_t *p)
 	}
 #endif /* __APPLE__ */
 
+#if defined(__FreeBSD__) && __FreeBSD_version < 1600006
 	/*
-	 * If this is FreeBSD, and the device name begins with "usbus",
+	 * If this is legacy FreeBSD, and the device name begins with "usbus",
 	 * try to create the interface if it's not available.
 	 */
-#if defined(__FreeBSD__) && defined(SIOCIFCREATE2)
 	if (strncmp(p->opt.device, usbus_prefix, USBUS_PREFIX_LEN) == 0) {
 		/*
 		 * Do we already have an interface with that name?
@@ -2126,7 +2126,7 @@ pcap_activate_bpf(pcap_t *p)
 			pcapint_add_to_pcaps_to_close(p);
 		}
 	}
-#endif /* defined(__FreeBSD__) && defined(SIOCIFCREATE2) */
+#endif /* defined(__FreeBSD__) && __FreeBSD_version < 1600006 */
 
 #ifdef HAVE_ZEROCOPY_BPF
 	/*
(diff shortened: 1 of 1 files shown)


glebius:
P.S. This is exactly the code that was already committed to FreeBSD's contrib/libpcap directory.

glebius:
The OpenBSD test failure doesn't seem to be related to my change.

pcap-bpf.c
```diff
@@ -2049,11 +2049,11 @@ pcap_activate_bpf(pcap_t *p)
 	}
 #endif /* __APPLE__ */
 
+#if defined(__FreeBSD__) && __FreeBSD_version < 1600006
 	/*
-	 * If this is FreeBSD, and the device name begins with "usbus",
+	 * If this is legacy FreeBSD, and the device name begins with "usbus",
```
guyharris, line 2054: "FreeBSD prior to 16.0" might be a bit more neutral.

glebius: Sure. Do you want me to push update or you can do it?

pcap-bpf.c
```diff
@@ -43,7 +43,7 @@
 #endif
 #include <sys/utsname.h>
 
-#if defined(__FreeBSD__) && defined(SIOCIFCREATE2)
+#if defined(__FreeBSD__) && __FreeBSD_version < 1600006
```
guyharris, line 46: Why this change?

guyharris: For example, what if it's compiled on a platform that's newer but run on a platform that's older?

pcap-bpf.c
```diff
@@ -2857,18 +2857,50 @@ check_bpf_bindable(const char *name)
...
 {
-	/*
-	 * XXX - if there's a way to determine whether there's something
-	 * plugged into a given USB bus, use that to determine whether
-	 * this device is "connected" or not.
-	 */
 	return (0);
 }
 
+#if __FreeBSD_version >= 1600006
```
guyharris, line 2867: Why not test for `#if defined(BIOCGETIFLIST)`, in case *other* operating systems implement this?

pcap-bpf.c
glebius: > For example, what if it's compiled on a platform that's newer but run on a platform that's older?

The `defined(SIOCIFCREATE2)` compile time check has the same problem. Many ifdefed code in libpcap doesn't cover that case. Well, many ifdefed code in many projects has this problem :) I don't know of any generic solution.

pcap-bpf.c
glebius: Sure, that would be better. I submitted in the form that I submitted to make the patch isolated to FreeBSD only, to increase chances of being accepted. But indeed using `#if defined(BIOCGETIFLIST)` is better.

glebius:
> [guyharris](https://github.com/guyharris) merged commit [ea5317a](https://github.com/the-tcpdump-group/libpcap/commit/ea5317adaf0e4e308543fbd66977f89344f9ef7f)

Thanks!

infrastation:
On a stylistic note, the logic would be easier to follow if it had one key block at the top, something along these lines:
```c
// Some useful comment here...
#if defined(__FreeBSD__) && __FreeBSD_version >= 1600006
  #define USE_BPF_FOR_USB 1
#else
  #define USE_BPF_FOR_USB 0
#endif
```

And then the code further down could be conditional on `#if USE_BPF_FOR_USB`.

guyharris:
@glebius - please check the changes in #1599 to make sure they build on FreeBSD-CURRENT; they fix a file descriptor leak, as well as memory leaks on some errors, and also fill in the error message buffer for `ioctl()` and `malloc()` errors.

glebius:
> On a stylistic note, the logic would be easier to follow if it had one key block at the top, something along these lines:
...
> And then the code further down could be conditional on `#if USE_BPF_FOR_USB`.

I don't agree. The BIOCGETIFLIST returns all possible bpf taps, it is not _for USB_. It already returns ipfw(4) taps on FreeBSD CURRENT and will return a few other kinds in near future. On FreeBSD CURRENT we can use finddevs_bpf() even without a call to  pcapint_findalldevs_interfaces(). But that would be quite an intrusive patch, so I decided to avoid that.

So mentioning USB would be a retrospective thinking as in the past USB needed a special treatment. Now it doesn't. Now we first collect all interface names and assume each of them has a bpf tap (a legacy approach). Then we do BIOCGETIFLIST to collect those that the legacy mechanism missed.

guyharris:
> Now we first collect all interface names and assume each of them has a bpf tap (a legacy approach).

What we should do is mark all the ones in the network stack's interface list that *aren't* in the `BIOCGETIFLIST` as "these don't support packet capture or injection", so that, instead of a user asking "why doesn't my fubar0 interface show up?", they can ask "why does it say capturing isn't supported on my fubar0 interface?" - the former requires some research, the latter can hopefully just tell the user to ask the OS supplier to make it support capturing.

(Yes, this really is a thing; Wireshark users would ask about it on the Q&A site and mailing lists.)

(And, yes, that's why I asked about @infrastation's "capture not supported"/"injection not supported" flags - I wanted to use them for that purpose, turning them *both* on for interfaces not supported by the capture mechanism.)

> On FreeBSD CURRENT we can use finddevs_bpf() even without a call to pcapint_findalldevs_interfaces(). But that would be quite an intrusive patch, so I decided to avoid that.

And it also might provoke users to ask "why doesn't my fubar0 interface show up?", which is another reason to avoid it.

PR #1599, opened 2025-12-21
bpf: fix resource leaks and set error messages.

guyharris:
In the FreeBSD-CURRENT code to use BIOCGETIFLIST, close the BPF device descriptor and free the device list buffer before returning.

For ioctl() and malloc() errors, use pcapint_fmt_errmsg_for_errno() to generate an error message.

Fix some comments.

### DIFF
--- a/pcap-bpf.c
+++ b/pcap-bpf.c
@@ -45,7 +45,7 @@
 
 #if defined(__FreeBSD__) && __FreeBSD_version < 1600006
 /*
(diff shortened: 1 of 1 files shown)


guyharris:
@glebius - FreeBSD should pick up these changes; they fix a file descriptor leak, as well as memory leaks on some errors, and also fill in the error message buffer for ioctl() and malloc() errors.

glebius:
Sure! Thanks a lot for the review.

PR #1600, opened 2025-12-22
Check for nanosecond precision at create time

dmiller-nmap:
Npcap's nanosecond precision depends on npcap driver support, so on the off chance that the driver version is wrong, we will check whether we can actually set the related mode on this adapter.

### DIFF
--- a/pcap-npf.c
+++ b/pcap-npf.c
@@ -1635,6 +1635,26 @@ get_ts_types(const char *device, pcap_t *p, char *ebuf)
 			}
 		}
 		p->tstamp_type_count = num_ts_types;
+
+#ifdef PACKET_MODE_NANO
+		/*
+		 * Check if we support nanosecond time stamps.
+		 */
+		if (p->tstamp_precision_list == NULL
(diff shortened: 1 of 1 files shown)


guyharris:
> we will check whether we can actually set the related mode on this adapter.

And, if it succeeds, that *sets* the mode - even if the program doesn't end up setting it after the create and before the activate, so it might result in defaulting to nanosecond precision.

You might have to change the code in `pcap_activate_npf()` to set the time stamp to microsecond resolution if `p->opt.tstamp_precision` is `PCAP_TSTAMP_PRECISION_MICRO`, or change the code in `get_ts_types()` to set the precision back to microsecond precision if the call to change it to nanosecond precision succeeds.

(Also, you might want to have a separate `get_ts_precisions()` routine, as the "types" in `get_ts_types()` refers to types such as "host", "adapter", etc..)

dmiller-nmap:
The adapter handle opened in this function is closed immediately after
this, so nothing persists.

PR #1601, opened 2025-12-22
Introduce and use translatetest.

infrastation:
These changes cover the latest bug fixes with some tests. This is is going to be merged soon.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -150,6 +150,8 @@ if [ "$CMAKE" = no ]; then
     run_after_echo $VALGRIND_CMD testprogs/versiontest
     FILTERTEST_BIN="$VALGRIND_CMD testprogs/filtertest"
     export FILTERTEST_BIN
+    TRANSLATETEST_BIN="$VALGRIND_CMD testprogs/translatetest"
+    export TRANSLATETEST_BIN
     run_after_echo "$MAKE_BIN" -s check
     run_after_echo $VALGRIND_CMD testprogs/findalldevstest
     [ "$TEST_RELEASETAR" = yes ] && run_after_echo "$MAKE_BIN" releasetar
@@ -157,6 +159,8 @@ else
     run_after_echo $VALGRIND_CMD run/versiontest
     FILTERTEST_BIN="$VALGRIND_CMD run/filtertest"
     export FILTERTEST_BIN
+    TRANSLATETEST_BIN="$VALGRIND_CMD run/translatetest"
+    export TRANSLATETEST_BIN
     run_after_echo "$MAKE_BIN" -s check
     run_after_echo $VALGRIND_CMD run/findalldevstest
 fi
--- a/testprogs/TESTrun
+++ b/testprogs/TESTrun
@@ -74,7 +74,7 @@ use constant {
 	EX_USAGE => 64,
 	# In this libpcap version filtertest exits with EX_DATAERR if the user input
 	# is invalid and with other status codes for other (memory/file/etc) error
-	# conditions.
+	# conditions.  translatetest implements the same convention.
 	EX_DATAERR => 65,
 	# Likewise, EX_NOINPUT if pcap_open_offline() has failed.
 	EX_NOINPUT => 66,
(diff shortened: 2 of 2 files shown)


infrastation:
> stringtest.c(8): fatal error C1083: Cannot open include file: 'arpa/inet.h': No such file or directory

So `htons()` will have to come from somewhere else on Windows.

infrastation:
As far as I understand it, `pcap/pcap.h` and other headers include `<winsock2.h>` on Windows, which declares the function, and outwith Windows the appropriate solution would be to do the same as `extract.h` does:
```c
#ifndef _WIN32
#include <arpa/inet.h>
#endif
```
Let's see if anything else does not pass the CI.

testprogs/stringtest.c
```diff
@@ -0,0 +1,214 @@
+// for "pcap-int.h"
+#include <config.h>
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdarg.h>
+#include <string.h>
+#include <arpa/inet.h>
```
guyharris: Windows doesn't have arpa/inet.h; use winsock2.h instead.

infrastation:
This time it does not like feeding a more-than-16-bit integer to `htons()`...

infrastation:
This revision does not use `htons()`, loses some other unnecessary code and includes testing of `pcapint_xdtoi()`. The latter made it clear that the new test program is not limited to parsing of strings, so it is called `translatetest` now.

infrastation:
This revision exempts Windows from building and using `translatetest`.

guyharris:
> The latter made it clear that the new test program is not limited to parsing of strings, so it is called `translatetest` now.

libpcap does not currently use any APIs such as the `gettext` APIs, to allow localization of messages, so the name `translatetest`, for now, doesn't have alternative interpretations.

testprogs/CMakeLists.txt
```diff
@@ -39,6 +39,18 @@ add_test_executable(writecaptest)
 
 if(NOT WIN32)
   add_test_executable(selpolltest)
+  # Fails to link on Windows:
+  # translatetest.obj : error LNK2019: unresolved external symbol
+  #   _pcapint_atodn referenced in function _test_pcapint_atodn
```
guyharris, line 44: Is `translatetest` being linked with the static version of libpcap (.lib, on Windows) or the dynamic (.dll, on Windows) version?

Linking with the dynamic version doesn't allow access to internal routines, it only allows access to explicitly-exported API routines.

On  UN\*X (where limiting access to explicitly-exported API routines isn't the default, but we enable it), the static version of libpcap is just libpcap.a.

On Windows, it's complicated by the dynamically library providing a separate "import library", with which code using the library is linked at build time; for libpcap, that library has the same name as the dynamic library, with .dll replaced with .lib - so you have to give the static version of a library a different name.

I think the different name we use is pcap_static.lib, with pcap.lib being the import library for pcap.dll.

rpcapd is also linked with the static libpcap library, in order to get at the socket helper and rpcap protocol routines. (That's a bit of an ugly hack.) That's handled in rpcapd/CMakeLists.txt (or, on Windows, rpcapd\\CMakeLists.txt :-)); see

```
  if(WIN32)
    target_link_libraries(rpcapd ${LIBRARY_NAME} 
      ${RPCAPD_LINK_LIBRARIES} ${PCAP_LINK_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT})
  else(WIN32)
    target_link_libraries(rpcapd ${LIBRARY_NAME}_static
      ${RPCAPD_LINK_LIBRARIES} ${PCAP_LINK_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT})
  endif(WIN32)
```

infrastation: Thank you for explaining that. There is a similar provision in `testprogs/CMakeLists.txt:add_test_executable`, but `translatetest` seems to be the only test program that requires the internal (yet not static) functions. Maybe all test programs should always link with libpcap statically, it serves the purpose well enough when Windows is not involved. Right now I do not have a good idea about the linking method, so I would rather merge this working revision and switch to other pieces in need of attention.

infrastation:
Testing of [hypothetical] code paths that involve libintl could be done using an executable named `gettexttest`, `intltest` or something nearly as descriptive.

PR #1603, opened 2025-12-29
Introduce a routine to parse decimal numbers in a string.

guyharris:
This wraps strtoul() to do more error checking.

It's currently used only in one place, but there are other places where it could be useful.

Make translatetest's getopt() code do the standard "add optind to argv and subtract it frm argc" after the loop, so the first argument after the option arguments is argv[0] and there are a total of argc post-options arguments, so that it can handle a "--" argument to allow "-1" to be passed as a non-option argument, even with GNU getopt().

Add some tests, courtesy of Denis Ovsienko.

### DIFF
--- a/pcap.c
+++ b/pcap.c
@@ -64,6 +64,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <ctype.h>
 #if !defined(_MSC_VER) && !defined(__BORLANDC__) && !defined(__MINGW32__)
 #include <unistd.h>
 #endif
@@ -3866,6 +3867,78 @@ pcap_strerror(int errnum)
 	 */
 	return (strerror(errnum));
 #endif /* _WIN32 */
+}
+
+/*
+ * Routine to parse a string containing an unsigned decimal integer,
+ * which catches some errors that strtoul() doesn't (strtoul() appears
+ * to parse numbers according to the way a C compiler does, so it skips
+ * leading white space and will happily allow an unsigned number with
+ * a negative sign), and also can treat the string not being completely
+ * numeric as an error and will treat values that don't fit into
+ * an unsigned int as an error.
+ *
+ * On success, returns 0 and sets the value pointed to by the last
+ * argument to the integer value; on error, returns EINVAL for an
+ * invalid number or ERANGE for a value that's too large to fit in
+ * an unsigned int.
+ */
+int
+pcapint_get_decuint(const char *cp, char **endptr, unsigned *nump)
+{
+	unsigned long val;
+	char *end;
+
+	if (cp == NULL) {
+		/* Don't do this. */
+		return (EINVAL);
+	}
+
+	if ((cp[0] & 0x80) != 0 || isspace((unsigned char)cp[0]) ||
+	    cp[0] == '-' || cp[0] == '+') {
+		/*
+		 * Numbers don't begin with non-ASCII characters or white
+		 * space and unsigned numbers don't have a sign.
+		 */
+		*nump = 0;
+		if (endptr != NULL)
+			*endptr = (char *)cp;
+		return (EINVAL);
+	}
+
+	/*
+	 * Clear errno, so we can check whether it was set by strtoul().
+	 */
+	errno = 0;
+	val = strtoul(cp, &end, 10);
+	if ((val == 0 && end == cp) || (endptr == NULL && *end != '\0')) {
+		/*
+		 * Parsing error, including, if endptr is NULL, the
+		 * string ending with a non-digit character.
+		 */
+		*nump = 0;
+		if (endptr != NULL)
+			*endptr = end;
+		return (EINVAL);
+	}
+	if (val == ULONG_MAX && errno == ERANGE) {
+		/* Number is bigger than ULONG_MAX */
+		*nump = 0;
+		if (endptr != NULL)
+			*endptr = end;
+		return (ERANGE);
+	}
+	if (val > UINT_MAX) {
+		/* Number won't fit in an unsigned int */
+		*nump = 0;
+		if (endptr != NULL)
+			*endptr = end;
+		return (ERANGE);
+	}
+	if (endptr != NULL)
+		*endptr = end;
+	*nump = (unsigned) val;
+	return (0);
 }
 
 int
--- a/testprogs/TESTrun
+++ b/testprogs/TESTrun
@@ -24813,6 +24813,32 @@ my @translate_accept_blocks = (
 		aliases => ['rpcap://eth0', 'eth0'],
 		expect => 'rpcap://eth0',
 	},
+
+	# pcapint_get_decuint()
+	{
+		name => 'decuint_noendp_1',
+		cfunc => 'pcapint_get_decuint/noendp',
+		aliases => ['1', '01'],
+		expect => '1',
+	},
+	{
+		name => 'decuint_noendp_4294967295',
+		cfunc => 'pcapint_get_decuint/noendp',
+		aliases => ['4294967295', '04294967295'],
+		expect => '4294967295',
+	},
+	{
+		name => 'decuint_endp_4294967295',
+		cfunc => 'pcapint_get_decuint/endp',
+		aliases => ['4294967295!', '04294967295!'],
+		expect => '4294967295 "!"',
+	},
+	{
+		name => 'decuint_endp_1',
+		cfunc => 'pcapint_get_decuint/endp',
+		aliases => ['1a', '01a'],
+		expect => '1 "a"',
+	},
 );
 
 # This works similar to @filter_reject_tests.  In each array element the hash
@@ -25103,6 +25129,62 @@ my @translate_reject_tests = (
 		arg => 'rpcap://eth0',
 		errstr => 'pcapint_parsesrcstr_ex() is not supported',
 	},
+
+	# pcapint_get_decuint()
+	{
+		name => 'decuint_noendp_nullargument',
+		cfunc => 'pcapint_get_decuint/noendp',
+		arg => undef,
+		errstr => 'EINVAL',
+	},
+	{
+		name => 'decuint_noendp_emptystring',
+		cfunc => 'pcapint_get_decuint/noendp',
+		arg => '',
+		errstr => 'EINVAL',
+	},
+	{
+		name => 'decuint_noendp_negative1',
+		cfunc => 'pcapint_get_decuint/noendp',
+		arg => '-1',
+		errstr => 'EINVAL',
+	},
+	{
+		name => 'decuint_noendp_positive1',
+		cfunc => 'pcapint_get_decuint/noendp',
+		arg => '+1',
+		errstr => 'EINVAL',
+	},
+	{
+		name => 'decuint_noendp_space1',
+		cfunc => 'pcapint_get_decuint/noendp',
+		arg => ' 1',
+		errstr => 'EINVAL',
+	},
+	{
+		name => 'decuint_noendp_tab1',
+		cfunc => 'pcapint_get_decuint/noendp',
+		arg => '\t1',
+		errstr => 'EINVAL',
+	},
+	{
+		name => 'decuint_noendp_nonascii1',
+		cfunc => 'pcapint_get_decuint/noendp',
+		arg => '\xEF1',
+		errstr => 'EINVAL',
+	},
+	{
+		name => 'decuint_noendp_4294967296',
+		cfunc => 'pcapint_get_decuint/noendp',
+		arg => '4294967296',
+		errstr => 'ERANGE',
+	},
+	{
+		name => 'decuint_noendp_1111111111111111111',
+		cfunc => 'pcapint_get_decuint/noendp',
+		arg => '1111111111111111111',
+		errstr => 'ERANGE',
+	},
 );
 
 ##### more pcapint_parsesrcstr_ex() accept blocks
@@ -25326,7 +25408,43 @@ sub common_filtertest_args {
 
 sub common_translatetest_args {
 	my $test = shift;
-	my @args = ($translatetest, $test->{cfuncname});
+	#
+	# The GNU libc getopt() doesn't behave as one might expect,
+	# and, at least as I read the POSIX/SUS page on getopt(),
+	# it doesn't behave as it says it should.
+	#
+	# The POSIX/SUS page says that
+	#
+	#   If, when getopt() is called:
+	#
+	#   argv[optind]  is a null pointer
+	#  *argv[optind]  is not the character -
+	#   argv[optind]  points to the string "-"
+	#
+	#   getopt() shall return -1 without changing optind.
+	#
+	# which *I* read as indicating that an argument that
+	# doesn't begin with '-' stops option argument processing.
+	#
+	# That is *not* what GNU getopt() does by default.
+	#
+	# To get what I read as the POSIX/SUS behavior, either:
+	#
+	#   the environment variable POSIXLY_CORRECT must be set;
+	#   the option string argument must begin with '+'.
+	#
+	# Otherwise, a "--" argument must be placed before any
+	# of the non-option arguments.
+	#
+	# We choose the latter option - the POSIX/SUS getopt()
+	# page doesn't obviously seem to require that a "+" not
+	# have other undesired effects.
+	#
+	# Had the GNU folks made the standard behavior the default,
+	# and required a "+" to get the behavior they wanted, That
+	# Would Have Been Wonderful. But they didn't, so here we are.
+	#
+	my @args = ($translatetest, "--", $test->{cfuncname});
 	# The hash key exists.  For the value Perl undef means C NULL and no
 	# command-line argument in the shell command, an empty Perl string
 	# means an empty C string and a quoted empty string argument in the
--- a/pcap-bt-linux.c
+++ b/pcap-bt-linux.c
@@ -150,7 +150,8 @@ bt_create(const char *device, char *ebuf, int *is_ours)
 {
 	const char *cp;
 	char *cpend;
-	long devnum;
+	int ret;
+	unsigned devnum;
 	pcap_t *p;
 
 	/* Does this look like a Bluetooth device? */
@@ -163,14 +164,9 @@ bt_create(const char *device, char *ebuf, int *is_ours)
 	}
 	/* Yes - is BT_IFACE followed by a number? */
 	cp += sizeof BT_IFACE - 1;
-	devnum = strtol(cp, &cpend, 10);
-	if (cpend == cp || *cpend != '\0') {
-		/* Not followed by a number. */
-		*is_ours = 0;
-		return NULL;
-	}
-	if (devnum < 0) {
-		/* Followed by a non-valid number. */
+	ret = pcapint_get_decuint(cp, &cpend, &devnum);
+	if (ret != 0) {
+		/* Not followed by a valid number */
 		*is_ours = 0;
 		return NULL;
 	}
--- a/pcap-int.h
+++ b/pcap-int.h
@@ -617,6 +617,22 @@ int	pcapint_createsrcstr_ex(char *, int, const char *, const char *,
 int	pcapint_parsesrcstr_ex(const char *, int *, char *, char *,
     char *, char *, unsigned char *, char *);
 
+/*
+ * Routine to parse a string containing an unsigned decimal integer,
+ * which catches some errors that strtoul() doesn't (strtoul() appears
+ * to parse numbers according to the way a C compiler does, so it skips
+ * leading white space and will happily allow an unsigned number with
+ * a negative sign), and also can treat the string not being completely
+ * numeric as an error and will treat values that don't fit into
+ * an unsigned int as an error.
+ *
+ * On success, returns 0 and sets the value pointed to by the last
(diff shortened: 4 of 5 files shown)


pcap.c
```diff
@@ -3868,6 +3868,69 @@ pcap_strerror(int errnum)
...
+ * numeric as an error and will treat values that don't fit into
+ * an unsigned int as an error.
+ */
+int
+pcapint_strtou(const char *cp, char **endptr, int base, unsigned *nump)
+{
+	unsigned long val;
+	char *end;
+
+	if (cp[0] == '-' || cp[0] == '+') {
```
infrastation: What if the first character is a space, and the second is a sign?

infrastation:
I had to produce similar code about a year ago (`strtouint31()`) and another recently (`test_pcapint_xdtoi()`). There is still a call to `atoi()` and a few calls to `strtoul()` in various files, so perhaps the new function could be reused in more places.

One easy way to verify a parsing function is to print the parsed value to a temporary string and to compare the temporary string with the input string. For decimal base that would trivially reject whitespace, sign and anything else `strtoul()` did not reject. For hexadecimal, if required, that would require additional moves.

pcap.c
guyharris: > What if the first character is a space

In several use cases, that's an error - "bluetooth 0" isn't a valid Bluetooth device.

My inclination is to have it fail if the string has any leading white space, and, if somebody's parsing a string with multiple tokens separated by white space, have the scanner handle the white space.

infrastation: If the new function came with tests, it would be easier to see what is and what is not valid input.

guyharris:
> One easy way to verify a parsing function is to print the parsed value to a temporary string and to compare the temporary string with the input string. For decimal base that would trivially reject whitespace, sign and anything else strtoul() did not reject.

In error cases, the parsed value is 0, so the temporary string would be "0"; that obviously won't match a string with a non-zero value, and an invalid string with a zero value would be " 0" or "-0" or "+0" or "0hello".

But if the idea is to check whether it fails on those cases, why not just check for a return value of 0 ("false")?

> For hexadecimal, if required, that would require additional moves.

For what failure cases wouldn't it work?

infrastation:
For many parsing problems printing the result and comparing it with the input string is a reliable way to catch a corner case that more sophisticated checks have missed. To validate hexadecimal numbers, this would have to disregard letter case and to decide whether the "0x" prefix is a part of the input string.

guyharris:
> To validate hexadecimal numbers, this would have to disregard letter case

`strcasecmp()` or, if not present on the platform, our own implementation of it.

> and to decide whether the "0x" prefix is a part of the input string.

There's also the issue of leading zeroes - if we print decimal values with `"%u"`, octal values with `"%#o"`, and hex values with `"%#x"`, we should skip a leading `"0x"` for hex and then skip all 0s after that before comparing.

For a base of 0, we'd have to check the input string for a leading `"0x"` and print in hex, otherwise check for a leading `"0"` and print in octal, otherwise print in decimal.

(We won't bother with base 17, for example. :-))

pcap.c
guyharris: > My inclination is to have it fail if the string has any leading white space,

Done.

infrastation:
It may be a good idea to nail it down to base-10 until there is a good use case for any other base.

infrastation:
I wonder how much of this could be solved using `translatetest` and `TESTrun`.

guyharris:
> I wonder how much of this could be solved using `translatetest` and `TESTrun`.

I'm not sure how much of it could be done without transplanting most of the code into translatetest.c.

infrastation:
Very little, it seems. Most of the new text would be the new test cases in Perl parlance, and the logistics is already in place.
```patch
diff --git a/testprogs/TESTrun b/testprogs/TESTrun
index 7fd296d6..6b827f5d 100755
--- a/testprogs/TESTrun
+++ b/testprogs/TESTrun
@@ -24813,6 +24813,32 @@ my @translate_accept_blocks = (
 		aliases => ['rpcap://eth0', 'eth0'],
 		expect => 'rpcap://eth0',
 	},
+
+	# pcapint_get_decuint()
+	{
+		name => 'decuint_noendp_1',
+		cfunc => 'pcapint_get_decuint/noendp',
+		aliases => ['1', '01'],
+		expect => '1',
+	},
+	{
+		name => 'decuint_noendp_4294967295',
+		cfunc => 'pcapint_get_decuint/noendp',
+		aliases => ['4294967295', '04294967295'],
+		expect => '4294967295',
+	},
+	{
+		name => 'decuint_endp_4294967295',
+		cfunc => 'pcapint_get_decuint/endp',
+		aliases => ['04294967295!'],
+		expect => '4294967295 "!"',
+	},
+	{
+		name => 'decuint_endp_1',
+		cfunc => 'pcapint_get_decuint/endp',
+		aliases => ['1a', '01a'],
+		expect => '1 "a"',
+	},
 );
 
 # This works similar to @filter_reject_tests.  In each array element the hash
@@ -25103,6 +25129,26 @@ my @translate_reject_tests = (
 		arg => 'rpcap://eth0',
 		errstr => 'pcapint_parsesrcstr_ex() is not supported',
 	},
+
+	# pcapint_get_decuint()
+	{
+		name => 'decuint_noendp_space1',
+		cfunc => 'pcapint_get_decuint/noendp',
+		arg => ' 1',
+		errstr => 'EINVAL',
+	},
+	{
+		name => 'decuint_noendp_tab1',
+		cfunc => 'pcapint_get_decuint/noendp',
+		arg => "\t1",
+		errstr => 'EINVAL',
+	},
+	{
+		name => 'decuint_noendp_4294967296',
+		cfunc => 'pcapint_get_decuint/noendp',
+		arg => "4294967296",
+		errstr => 'ERANGE',
+	},
 );
 
 ##### more pcapint_parsesrcstr_ex() accept blocks
diff --git a/testprogs/translatetest.c b/testprogs/translatetest.c
index 43744c74..b9241391 100644
--- a/testprogs/translatetest.c
+++ b/testprogs/translatetest.c
@@ -158,6 +158,42 @@ test_pcapint_parsesrcstr_ex(const char *arg)
 	return EX_OK;
 }
 
+static const char *
+errcode(const int e)
+{
+	return e == EINVAL ? "EINVAL" :
+		e == ERANGE ? "ERANGE" :
+		"unknown";
+}
+
+static int
+test_pcapint_get_decint_endp(const char *arg)
+{
+	char *endp;
+	unsigned output;
+	int res = pcapint_get_decuint(arg, &endp, &output);
+	if (res != 1) {
+		fprintf(stderr, "ERROR: %s (%d)\n", errcode(errno), errno);
+		return EX_DATAERR;
+	}
+	printf("OK: %u \"%s\"\n", output, endp);
+	return EX_OK;
+
+}
+
+static int
+test_pcapint_get_decint_noendp(const char *arg)
+{
+	unsigned output;
+	int res = pcapint_get_decuint(arg, NULL, &output);
+	if (res != 1) {
+		fprintf(stderr, "ERROR: %s (%d)\n", errcode(errno), errno);
+		return EX_DATAERR;
+	}
+	printf("OK: %u\n", output);
+	return EX_OK;
+
+}
 static const struct {
 	const char *name;
 	u_char null_ok;
@@ -171,6 +207,8 @@ static const struct {
 	{"pcapint_atoan", 0, test_pcapint_atoan, "ARCnet address"},
 	{"pcap_ether_aton", 0, test_pcap_ether_aton, "MAC-48 address"},
 	{"pcapint_parsesrcstr_ex", 1, test_pcapint_parsesrcstr_ex, "source string"},
+	{"pcapint_get_decuint/endp", 1, test_pcapint_get_decint_endp, "unsigned integer"},
+	{"pcapint_get_decuint/noendp", 1, test_pcapint_get_decint_noendp, "unsigned integer"},
 };
 #define NUM_FUNCS (sizeof(testfunc) / sizeof(testfunc[0]))
 ```
For a shorter feedback loop you can set `TESTRUN_SHORT=1` and `TESTRUN_JOBS=4`.

**Comment by guyharris** [comment]
> Very little, it seems. Most of the new text would be the new test cases in Perl parlance, and the logistics is already in place.

Plus tests for numbers with a leading sign, which should fail with EINVAL, as -1 is not a valid unsigned number and +1 is redundant.

**Comment by infrastation** [comment]
The patch includes only as many tests as is necessary to convey the point, and could use a couple other cleanups. Feel free to adopt as you see fit.

**Comment by infrastation** [comment]
For negative values the invocation probably requires `translatetest -- <args>` or extra quotes in order not to confuse `getopt()` (or perhaps it could do without `getopt()` since the only option is `-h`), I can have a look later if you do not fix this first.

The tests for the new function should include test cases where the input is an empty string, which the new function properly rejects:
```
$ ./testprogs/translatetest pcapint_get_decuint/endp ''
ERROR: EINVAL (22)
$ ./testprogs/translatetest pcapint_get_decuint/noendp ''
ERROR: EINVAL (22)
```

Also the tests should include passing a C `NULL` (Perl `undef`), which the new function currently does not reject safely:
```
$ ./testprogs/translatetest pcapint_get_decuint/endp
Segmentation fault
$ ./testprogs/translatetest pcapint_get_decuint/noendp
Segmentation fault
```

For reference, your C tests from the previous revision are below:
```c
struct get_int_tests {
	const char *input;
	const char *output;
	const char *remaining;
	int error;
};

static const struct get_int_tests get_decuint_tests[] = {
	/* basic test */
	{ "1", "1", NULL, 0 },
	{ "4294967295", "4294967295", NULL, 0 },

	/* leading zeroes */
	{ "01", "1", NULL, 0 },
	{ "04294967295", "4294967295", NULL, 0 },

	/* value possibly followed by more */
	{ "1", "1", "", 0 },
	{ "01", "1", "", 0 },
	{ "04294967295", "4294967295", "", 0 },
	{ " 1", NULL, NULL, EINVAL },
	{ "\t1", NULL, NULL, EINVAL },
	{ " 4294967295", NULL, NULL, EINVAL },
	{ "\t4294967295", NULL, NULL, EINVAL },
	{ "1a", "1", "a", 0 },
	{ " 1a", NULL, NULL, EINVAL },
	{ "\t1a", NULL, NULL, EINVAL },
	{ "4294967295a", "4294967295", "a", 0 },
	{ " 4294967295a", NULL, NULL, EINVAL },
	{ "\t4294967295a", NULL, NULL, EINVAL },
	{ "01a", "1", "a", 0 },
	{ " 01a", NULL, NULL, EINVAL },
	{ "\t01a", NULL, NULL, EINVAL },
	{ "04294967295a", "4294967295", "a", 0 },
	{ " 04294967295a", NULL, NULL, EINVAL },
	{ "\t04294967295a", NULL, NULL, EINVAL },
	{ "1!", "1", "!", 0 },
	{ " 1!", NULL, NULL, EINVAL },
	{ "\t1!", NULL, NULL, EINVAL },
	{ "4294967295!", "4294967295", "!", 0 },
	{ " 4294967295!", NULL, NULL, EINVAL },
	{ "\t4294967295!", NULL, NULL, EINVAL },
	{ "01!", "1", "!", 0 },
	{ " 01!", NULL, NULL, EINVAL },
	{ "\t01!", NULL, NULL, EINVAL },
	{ "04294967295!", "4294967295", "!", 0 },
	{ " 04294967295!", NULL, NULL, EINVAL },
	{ "\t04294967295!", NULL, NULL, EINVAL },

	/* leading spaces aren't valid */
	{ " 1", NULL, NULL, EINVAL },
	{ "\t1", NULL, NULL, EINVAL },

	/* negative numbers aren't valid */
	{ "-1", NULL, NULL, EINVAL },
	{ " -1", NULL, NULL, EINVAL },
	{ "\t-1", NULL, NULL, EINVAL },

	/* leading posiive sign isn't valid */
	{ "+1", NULL, NULL, EINVAL },
	{ " +1", NULL, NULL, EINVAL },
	{ "\t+1", NULL, NULL, EINVAL },

	/* non-ASCII characters aren't valid */
	{ "\xEF" "1", NULL, NULL, EINVAL },
	{ " \xEF" "1", NULL, NULL, EINVAL },
	{ "\t\xEF" "1", NULL, NULL, EINVAL },

	/* test whether anything non-numeric is invalid */
	{ "1a", NULL, NULL, EINVAL },
	{ " 1a", NULL, NULL, EINVAL },
	{ "\t1a", NULL, NULL, EINVAL },
	{ "1!", NULL, NULL, EINVAL },
	{ " 1!", NULL, NULL, EINVAL },
	{ "\t1!", NULL, NULL, EINVAL },

	/* test whether anything non-numeric is handled correctly */
	{ "1a", "1", "a", EINVAL },
	{ " 1a", NULL, NULL, EINVAL },
	{ "\t1a", NULL, NULL, EINVAL },
	{ "1!", "1", "!", EINVAL },
	{ " 1!", NULL, NULL, EINVAL },
	{ "\t1!", NULL, NULL, EINVAL },

	/* value that doesn't fit into an unsigned int (assumed 32-bit) */
	{ "4294967296", NULL, NULL, ERANGE },
	{ "1111111111111111111", NULL, NULL, ERANGE }
};
```

guyharris:
> For negative values the invocation probably requires `translatetest -- <args>` or extra quotes in order not to confuse `getopt()`

`getopt()` should stop as soon as it sees an argument that *doesn't* begin with`-`, and `pcapint_get_decuint/endp` and `pcapint_get_decuint/noendp` are arguments that doesn't begin with `-`, so `translatetest pcapint_get_decuint/endp -1` and  `translatetest pcapint_get_decuint/noendp -1` should Just Work.

guyharris:
> > For negative values the invocation probably requires `translatetest -- <args>` or extra quotes in order not to confuse `getopt()`
> 
> `getopt()` should stop as soon as it sees an argument that _doesn't_ begin with`-`, and `pcapint_get_decuint/endp` and `pcapint_get_decuint/noendp` are arguments that doesn't begin with `-`, so `translatetest pcapint_get_decuint/endp -1` and `translatetest pcapint_get_decuint/noendp -1` should Just Work.

Silly me, I thought GNU getopt worked  like a normal getopt by default. It doesn't.

testprogs/TESTrun
```diff
@@ -25326,7 +25384,43 @@ sub common_filtertest_args {
...
+	#   If, when getopt() is called:
+	#
+	#   argv[optind]  is a null pointer
+	#  *argv[optind]  is not the character -
+	#   argv[optind]  points to the string "-"
+	#
+	#   getopt() shall return -1 without changing optind.
+	#
+	# which *I* read as indicating that an argument that
+	# doesn't begin with '-' stops option argument rpocessing.
```
infrastation: "processing"

testprogs/TESTrun
```diff
@@ -25326,7 +25384,43 @@ sub common_filtertest_args {
...
+	#   the option string argument must begin with '+'.
+	#
+	# Otherwise, a "--" argument must be placed before any
+	# of the non-option arguments.
+	#
+	# We choose the latter option - the POSIX/SUS getopt()
+	# page doesn't obviously seem to require that a "+" not
+	# have other undesired effects.
+	#
+	# Had the GNU folks made the standard behavir the default,
```
infrastation: "behavio[u]r"

testprogs/TESTrun
```diff
@@ -25326,7 +25384,43 @@ sub common_filtertest_args {
...
+	# and required a "+" to ge the behavior they wanted, That
```
infrastation: "to get"

testprogs/TESTrun
```diff
@@ -25326,7 +25384,43 @@ sub common_filtertest_args {
...
+	# Would Have Been Wonderful. But we didn't, so here we are.
```
infrastation: they didn't?

testprogs/TESTrun
```diff
@@ -25326,7 +25384,43 @@ sub common_filtertest_args {
 
 sub common_translatetest_args {
 	my $test = shift;
-	my @args = ($translatetest, $test->{cfuncname});
+	#
+	# The GNU libc getopt() doesn't behave as one might expect,
+	# and, at least as I read the POSIX/SUS page on getopt(),
+	# it dosn't behave as it says it should.
```
fxlb: dosn't ==> doesn't ?

infrastation:
One other thing to consider for the new function is to return 0 on success and a non-zero on an error, then the function could return an `errno`-style reason for failure without having a side effect of modifying the global `errno`.

testprogs/TESTrun
guyharris: Fixed.

testprogs/TESTrun
guyharris: Fixed (US-style).

testprogs/TESTrun
guyharris: Fixed.

testprogs/TESTrun
guyharris: Yes. Fixed.

testprogs/TESTrun
guyharris: Yes. Fixed.

guyharris:
> One other thing to consider for the new function is to return 0 on success and a non-zero on an error, then the function could return an `errno`-style reason for failure without having a side effect of modifying the global `errno`.

Done.

guyharris:
> For negative values the invocation probably requires `translatetest -- <args>` or extra quotes in order not to confuse `getopt()` (or perhaps it could do without `getopt()` since the only option is `-h`), I can have a look later if you do not fix this first.

Done.

Maybe we should copy tcpdump's `getopt_long()` implementation over, replacing our missing/getopt.c, and use that. Unlike `getopt()`, where *most* implementations are POSIX-compatible by default but the GNU libc implementation isn't, and some implementations might not understand "+" as the first character of the option string, all the implementations of `getopt_long()` for which I've read the documentation - \*BSD, GNU, musl, Solaris - support "+" as the first character of the option string meaning "POSIX behavior". Then we could just use "+" to force POSIX behavior.

(BTW, we should do this for tcpdump as well! Current tcpdump *fails* with

    tcpdump -d ip[4:4] == -1

as it thinks "-1" is a flag argument. And, as it's using `getopt_long()`, the failure is portable.)

(Oh, and I should probably whip up a proposal for `getopt_long()` for the next POSIX. Maybe that'll make IBM get with the program and add it to AIX; of the currently-active UN\*Xes, AIX appears to be the only holdout - Solaris has it, Linux distributions have it if they use either GNU, musl, or Bionic libc, and the \*BSDs, including CupertinoBSD, have it.)

> The tests for the new function should include test cases where the input is an empty string, which the new function properly rejects:

Done.

> Also the tests should include passing a C `NULL` (Perl `undef`)

Done.

> which the new function currently does not reject safely:

It now returns EINVAL.

> For reference, your C tests from the previous revision are below:

I've added some of them. I don't know if we need to be *that* orthogonal, so I didn't add all of them.

guyharris:
> AIX appears to be the only holdout

Good grief, IBM's other main in-house OS, *z/OS* (a descendant of [an OS that came out *more than fifty years ago*](https://en.wikipedia.org/wiki/MVS), and that was compatible with, and probably somewhat based on, [an OS that will have come out *sixty years ago* at the end of March 2026!](https://en.wikipedia.org/wiki/OS/360_and_successors)) [has `getopt_long()`](https://www.ibm.com/docs/en/zos/3.2.0?topic=functions-getopt-long-command-long-option-parsing)!

infrastation:
From what you are writing it seems you see the best solution already.

guyharris:
> From what you are writing it seems you see the best solution already.

Unless you see something I've missed....

infrastation:
In case it helps, [OpenVMS documentation](https://docs.vmssoftware.com/vsi-c-run-time-library-reference-manual-for-openvms-systems/) does not mention `getopt_long()`, but lists `getopt()` as available since 6.2. According to the document, the latter implements `--`, but not `POSIXLY_CORRECT` or `+`.

infrastation:
Also QNX 8.0 documents `getopt_long()` with support for `--`, `+` and `POSIXLY_CORRECT`, from Haiku source code I gather it includes GNU `getopt_long()` an an extension, hence the same behaviour, at least on paper.

guyharris:
> In case it helps, [OpenVMS documentation](https://docs.vmssoftware.com/vsi-c-run-time-library-reference-manual-for-openvms-systems/) does not mention `getopt_long()`, but lists `getopt()` as available since 6.2. According to the document, the latter implements `--`, but not `POSIXLY_CORRECT` or `+`.

The documentation for [a later OS by Dave Cutler](https://en.wikipedia.org/wiki/Windows_NT) doesn't mention it either; it doesn't even mention `getopt()`.

I.e., we already have a precedent for a non-UN\*X OS that we support and that doesn't have `getopt_long()` - we fall back on our (BSD-based) implementation in missing/getopt_long.c. DEC, Compaq, HP, or VMS Software (I think it was back when DEC still existed) probably added `getopt()` as part of the "provide a POSIX environment" project. As such, it probably implements POSIX behavior rather than GNU behavior.

guyharris:
> Also QNX 8.0 documents `getopt_long()` with support for `--`, `+` and `POSIXLY_CORRECT`

They probably got it from either GNU or BSD, both of which support all of those.

> from Haiku source code I gather it includes GNU `getopt_long()` an an extension,

They appear to be using the GNU libc version or something based on that.

guyharris:
> Maybe we should copy tcpdump's `getopt_long()` implementation over,

Well, maybe not. It involves a lot of configure script/CMake/source changes, so I'm inclined to leave it as it is for now.

PR #1608, opened 2026-01-09
optimize: avoid __builtin_ctz(0) in lowest_set_bit()

rootvector2:
__builtin_ctz(0) is undefined behavior under GCC/Clang and may lead to
miscompilation if the compiler assumes the argument is non-zero.

Replace the macro implementation with a static inline helper that
explicitly handles the zero case before calling the builtin. Apply the
same defined behavior to the MSVC _BitScanForward() path for consistency.

### DIFF
--- a/optimize.c
+++ b/optimize.c
@@ -104,8 +104,9 @@ pcap_set_print_dot_graph(int value)
  * If handed a non-zero value, returns the index of the lowest set bit,
  * counting upwards from zero.
  *
- * If handed zero, it returns 0 (historically this was unspecified/UB on
- * some compilers; callers must not rely on any particular result for 0).
+ * If handed zero, the results are platform- and compiler-dependent.
+ * Keep it out of the light, don't give it any water, don't feed it
+ * after midnight, and don't pass zero to it.
  *
  * This is the same as the count of trailing zeroes in the word.
  */
@@ -136,10 +137,10 @@ static __forceinline u_int
 lowest_set_bit(int mask)
 {
 	unsigned long index;
-
-	if (mask == 0)
-		return 0;
-
+	/*
+	 * Don't sign-extend mask if long is longer than int.
+	 * (It's currently not, in MSVC, even on 64-bit platforms, but....)
+	 */
 	(void)_BitScanForward(&index, (unsigned long)mask);
 	return (u_int)index;
 }
@@ -2120,7 +2121,7 @@ find_inedges(opt_state_t *opt_state, struct block *root)
 	int level;
 	struct block *b;
 
-	for ( i = 0; i < opt_state->n_blocks; ++i)
+	for (i = 0; i < opt_state->n_blocks; ++i)
 		opt_state->blocks[i]->in_edges = 0;
 
 	/*
--- a/optimize.c
+++ b/optimize.c
@@ -120,7 +120,7 @@ pcap_set_print_dot_graph(int value)
   {
   	if (mask == 0)
   		return 0;
-  	return (u_int)__builtin_ctz((unsigned int)mask);
+  	return (u_int)__builtin_ctz(mask);
   }
 #elif defined(_MSC_VER)
   /*
--- a/optimize.c
+++ b/optimize.c
@@ -104,17 +104,23 @@ pcap_set_print_dot_graph(int value)
  * If handed a non-zero value, returns the index of the lowest set bit,
  * counting upwards from zero.
  *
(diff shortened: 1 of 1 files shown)


optimize.c
```diff
@@ -129,15 +135,13 @@ pcap_set_print_dot_graph(int value)
 static __forceinline u_int
 lowest_set_bit(int mask)
 {
-	unsigned long bit;
+	unsigned long index;
 
-	/*
```
guyharris, line 134: You can remove the test for the result being zero, but please don't remove the comment, as it still applies, as far as I know.

optimize.c
```diff
@@ -2116,7 +2120,7 @@ find_inedges(opt_state_t *opt_state, struct block *root)
 	int level;
 	struct block *b;
 
-	for (i = 0; i < opt_state->n_blocks; ++i)
+	for ( i = 0; i < opt_state->n_blocks; ++i)
```
guyharris: Please don't add an extra space there.

optimize.c
```diff
@@ -104,17 +104,23 @@ pcap_set_print_dot_graph(int value)
  * If handed a non-zero value, returns the index of the lowest set bit,
  * counting upwards from zero.
  *
- * If handed zero, the results are platform- and compiler-dependent.
- * Keep it out of the light, don't give it any water, don't feed it
- * after midnight, and don't pass zero to it.
```
guyharris, line 109: [Joe Dante will cry if you remove the second sentence of that comment.](https://en.wikipedia.org/wiki/Gremlins)

guyharris: [Chris Columbus](https://en.wikipedia.org/wiki/Chris_Columbus_(filmmaker)), too.

guyharris:
As noted in the comment before all the declarations of `lowest_set_bit()`, it ***MUST*** not be called with a zero value and as seen in the one and only place `lowest_set_bit()` is called, it *isn't* called with a zero value (the loop it's called in loops as long as `x` is non-zero, and `x` is the argument to `lowest_set_bit()`), so it's not as if there's currently a bug caused by this.

However, any compiler worth its salt will figure out that it's being called with a guaranteed-not-to-be-zero value and optimize the test away, so this probably will cause the same code to be generated on any such compiler, so we might as well change it in case any *other* code ends up using it.

rootvector2:
I’ve pushed updates addressing all review feedback: the original comment is restored, the redundant zero check is removed, and the formatting issue is fixed.

Please let me know if there’s anything further to adjust.

optimize.c
```diff
@@ -113,8 +113,15 @@ pcap_set_print_dot_graph(int value)
...
    * GCC 3.4 and later; we have __builtin_ctz().
+   * Note: __builtin_ctz(0) is undefined behavior, so guard it.
    */
-  #define lowest_set_bit(mask) ((u_int)__builtin_ctz(mask))
+  static inline u_int
+  lowest_set_bit(u_int mask)
+  {
+  	if (mask == 0)
+  		return 0;
+  	return (u_int)__builtin_ctz((unsigned int)mask);
```
fxlb: mask is an `u_int `. Why do you cast it to `unsigned int`?

rootvector2:
Good catch the cast is unnecessary here. mask is already an unsigned integer type and the zero case is handled before the builtin is called I’ll drop the cast to keep the code clearer and consistent

optimize.c
```diff
@@ -129,15 +136,13 @@ pcap_set_print_dot_graph(int value)
 static __forceinline u_int
 lowest_set_bit(int mask)
```
guyharris: For consistency, make the argument a `u_int`, just as it is for the other version. The argument passed to it in the only call we currently make to it is a `bpf_u_int32`, which is the exact same type on all the platforms on which we run.

That would also mean that we don't need to cast `mask` to `unsigned long` - it will get promoted to `unsigned long` and no sign-extension will be done.

optimize.c
guyharris: And if the parameter is changed to `u_int`, and the case is removed, as per https://github.com/the-tcpdump-group/libpcap/pull/1608/files#r2678473653, the comment no longer applies.

optimize.c
rootvector2: I’ve changed lowest_set_bit() to take a u_int argument for consistency with the other implementation and the current call site, and removed the unnecessary cast.

infrastation:
Banned due to artificial imitation spam, closing

PR #1609, opened 2026-01-14
Fix the "gateway" primitive.

infrastation:
This is the first part of a bigger change, let's see how many OSes besides Linux pass the CI at this revision before I add the next part, which changes the tested behaviour.


infrastation:
The tests were passing, as far as I could see. This revision is supposedly free of compiler warnings.  It includes the second part, the third part with a few more commits is going to follow later.

infrastation:
This revision gets to the point and if everything goes well, is going to be merged in several days.

infrastation:
Fixed a typo.

PR #1611, opened 2026-01-23
pcap-dag: Add NULL check for inf before dereferencing.

suunj:
In dag_stream_long_description(), inf is checked for NULL in the previous line using a ternary operator, but immediately dereferenced without NULL check in the if condition.

Add NULL check to prevent NULL pointer dereference.

### DIFF
--- a/pcap-dag.c
+++ b/pcap-dag.c
@@ -1392,7 +1392,7 @@ dag_stream_long_description(const unsigned stream, const dag_size_t bufsize,
 	    stream,
 	    bufsize / 1024 / 1024,
 	    inf ? dag_device_name(inf->device_code, 1) : "N/A");
-	if (inf->device_code != PCI_DEVICE_ID_VDAG)
+	if (inf && inf->device_code != PCI_DEVICE_ID_VDAG)
 		snprintf(buf + done, sizeof(buf) - done,
 		    " rev %c at %s",
 		    (inf && inf->brd_rev < 26) ? ('A' + inf->brd_rev) : '?',


infrastation:
That's well spotted, thank you. Please note there is a similar dereference of `inf` in `dag_findalldevs()` before the call to `dag_stream_long_description()`, yet in practice this code works fine for both root and non-root user (or at least it did the last time I tested it about a year ago). Perhaps it did not fail because if both `dag_open()` and `dag_get_stream_buffer_size64()` have succeeded, `dag_pciinfo()` has succeeded too. I agree it would be wrong to rely on that and the code needs a fix, but the fix needs to be a bit different.

suunj:
Thank you for the feedback!

I see that `dag_findalldevs()` already checks `inf` before the if-condition, 
but passes it to `dag_stream_long_description()` where it could still be NULL.

Would you prefer:
- Option A: Keep the fix in `dag_stream_long_description()` as a defensive check
- Option B: Ensure `inf != NULL` before calling `dag_stream_long_description()` 
  in `dag_findalldevs()`, and remove the NULL checks inside the function

Please let me know which approach is preferred.

Thanks!

infrastation:
Let me have some time to think about it. This would need to be resolved before 1.11.0.

infrastation:
Pull request #1619 eliminates this place for failure differently: it errors out if `dag_pciinfo()` returns `NULL`.

infrastation:
This has been fixed as noted, the message of commit f7d89cb credits you for the discovery of this problem. Thank you!

PR #1613, opened 2026-01-28
introduce a set of SECURITY reporting details

mcr:
this is for libpcap, and represents some feedback from the GVIP Summit <span>#</span>01.


mcr:
point at web site

infrastation:
The draft branch has been moved [here](https://github.com/the-tcpdump-group/libpcap-incomplete-works/tree/more-security-reporting).

PR #1618, opened 2026-02-02
Free p->opt.device on close, not on cleanup.

### LINKED ISSUE #1615: NULL dereference in pcap_can_set_rfmon_linux via handle->opt.device == NULL after activation

Hi, there is a potential bug with invoking certain functions after a failed activation on `pcap_t`.

This bug was reproduced on https://github.com/the-tcpdump-group/libpcap/commit/a22678a.

### Description

This case is mostly about API hardening, but the general trend is:

- it's possible to create a new `pcap_t` with an invalid interface name (e.g. "A"), `pcap_create` returns a non-null `pcap_t *`
- this object properly has `p->opt.device` set to a string so functions like `pcap_can_set_rfmon` work
- attempting to activate it `pcap_activate` fails with an error code as expected, but as a byproduct it nulls out `p->opt.device`
- as a result, the pcap is in a new state where subsequent calls to e.g. `pcap_can_set_rfmon` crash via a null dereference

It's a fairly benign case, but since certain functions like `pcap_activate` can leave the pcap with certain fields in weird states (i.e. `opt.device` may be null), it could be safe to harden functions like `pcap_can_set_rfmon` with a null check, returning an error like `PCAP_ERROR` as already allowed by the documentation.

### POC

The following testcase demonstrates the bug:

**testcase.cpp**
```cpp
#include <cstdio>
#include <cstdlib>
extern "C" {
#include "/fuzz/install/include/pcap/pcap.h"
}
int main(){
  char errbuf[PCAP_ERRBUF_SIZE] = {0};
  // Benign, invalid interface name still triggers the crash after activation
  const char* src = "A";
  pcap_t* p = pcap_create(src, errbuf);
  if (!p) return 0;
  // Activate (may fail with negative code, but internal state gets set)
  (void)pcap_activate(p);
  // This calls pcap_can_set_rfmon_linux, which does strcmp(handle->opt.device, "any")
  // with handle->opt.device == NULL, causing a NULL dereference in strcmp.
  (void)pcap_can_set_rfmon(p);
  return 0;
}

```

**stdout**
```text

```

**stderr**
```text
AddressSanitizer:DEADLYSIGNAL
=================================================================
==1==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x55c06a87de70 bp 0x7ffd9f7de9d0 sp 0x7ffd9f7de170 T0)
==1==The signal is caused by a READ memory access.
==1==Hint: address points to the zero page.
    #0 0x55c06a87de70 in strcmp (/fuzz/test+0x42e70) (BuildId: e0b1f1698f5f9a3ce2e9affb54290ec0ec8ee452)
    #1 0x7f5256fceccf in pcap_can_set_rfmon_linux pcap-linux.c
    #2 0x55c06a9425a4 in main /fuzz/testcase.cpp:16:9
    #3 0x7f52564dbd8f in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #4 0x7f52564dbe3f in __libc_start_main csu/../csu/libc-start.c:392:3
    #5 0x55c06a867314 in _start (/fuzz/test+0x2c314) (BuildId: e0b1f1698f5f9a3ce2e9affb54290ec0ec8ee452)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV (/fuzz/test+0x42e70) (BuildId: e0b1f1698f5f9a3ce2e9affb54290ec0ec8ee452) in strcmp
==1==ABORTING

```

<details><summary>Steps to Reproduce</summary>
<p>

The crash was triaged with the following Dockerfile:

**Dockerfile**
```dockerfile
# Ubuntu 22.04 with some packages pre-installed
FROM hgarrereyn/stitch_repro_base@sha256:3ae94cdb7bf2660f4941dc523fe48cd2555049f6fb7d17577f5efd32a40fdd2c

RUN git clone https://github.com/the-tcpdump-group/libpcap.git /fuzz/src && \
    cd /fuzz/src && \
    git checkout a22678a && \
    git submodule update --init --remote --recursive

ENV LD_LIBRARY_PATH=/fuzz/install/lib
ENV ASAN_OPTIONS=hard_rss_limit_mb=1024:detect_leaks=0

RUN echo '#!/bin/bash\nexec clang-17 -fsanitize=address -O0 "$@"' > /usr/local/bin/clang_wrapper && \
    chmod +x /usr/local/bin/clang_wrapper && \
    echo '#!/bin/bash\nexec clang++-17 -fsanitize=address -O0 "$@"' > /usr/local/bin/clang_wrapper++ && \
    chmod +x /usr/local/bin/clang_wrapper++

# Install extra build dependencies
RUN apt-get update && apt-get install -y \
    cmake \
    build-essential \
    flex \
    bison \
    pkg-config \
    libssl-dev \
    libnl-3-dev \
    libnl-genl-3-dev \
    libbluetooth-dev \
    libibverbs-dev \
    libdbus-1-dev \
    libpthread-stubs0-dev \
    libnet1-dev \
    libcap-ng-dev \
    libsystemd-dev \
    && rm -rf /var/lib/apt/lists/*

RUN cmake /fuzz/src \
    -DCMAKE_BUILD_TYPE=Release \
    -DCMAKE_C_COMPILER=clang_wrapper \
    -DCMAKE_CXX_COMPILER=clang_wrapper++ \
    -DCMAKE_INSTALL_PREFIX=/fuzz/install \
    -DBUILD_SHARED_LIBS=ON \
    && make -j$(nproc) \
    && make install
```

**Build Command**
```bash
clang++-17 -fsanitize=address -g -O0 -o /fuzz/test /fuzz/testcase.cpp -I/fuzz/install/include -L/fuzz/install/lib -lpcap -lssl -lcrypto && /fuzz/test
```

**Reproduce**
1. Copy `Dockerfile` and `testcase.cpp` into a local folder.
2. Build the repro image:
```bash
docker build . -t repro --platform=linux/amd64
```
3. Compile and run the testcase in the image:
```bash
docker run \
    -it --rm \
    --platform linux/amd64 \
    --mount type=bind,source="$(pwd)/testcase.cpp",target=/fuzz/testcase.cpp \
    repro \
    bash -c "clang++-17 -fsanitize=address -g -O0 -o /fuzz/test /fuzz/testcase.cpp -I/fuzz/install/include -L/fuzz/install/lib -lpcap -lssl -lcrypto && /fuzz/test"
```

</p>
</details> 

---

### Additional Info
This testcase was discovered by `STITCH`, an autonomous fuzzing system. All reports are reviewed manually (by a human) before submission.

guyharris:
That way, if a failed pcap_activate() cleans up anything done by the activate by calling the module's cleanup routine, which should always call pcapint_cleanup_live_common(), p->opt.device, which is set by pcap_create() and shouldn't be freed or modified until the pcap_t is closed, will be left alone.

Commit e333a6044f7d2d3225a6a22205b6b7c1e389945f moved the freeing of p->opt.device to pcap_cleanup_live_common(), but the reason for doing that is that, before that commit, we were freeing p->opt.device *before* calling the module's cleanup routine, and pcap-npf.c's cleanup routine used p->opt.device to turn monitor mode off.

This change moves freeing of p->opt.device back, but does it *after* the module's cleanup routine is called, so it's safe for that rouine to use p->opt.device - as long as it doesn't save the value of p->opt.device and attempt to use it later. Copy the string if you need to do that.

Fix #1615.

Add a test for that crash in testprogs/activatetest.c. That won't necessarily catch *all* cases where this is a problem on *all* platforms, but it should at least catch it on Linux.

### DIFF
--- a/testprogs/activatetest.c
+++ b/testprogs/activatetest.c
@@ -44,6 +44,13 @@ int main(void)
 		        "FAIL: Unexpected error %d from pcap_activate().\n",
 		        err);
 	}
+
+	/*
+	 * And attempts to determine whether attributes can be set
+	 * on the pcap_t shouldn't crash after a failed activate.
+	 * See GitHub issue #1615.
+	 */
+	(void) pcap_can_set_rfmon(p);
 	pcap_close(p);
 	return ret;
 }
--- a/pcap.c
+++ b/pcap.c
@@ -4159,10 +4159,15 @@ pcapint_breakloop_common(pcap_t *p)
 void
 pcapint_cleanup_live_common(pcap_t *p)
 {
-	if (p->opt.device != NULL) {
-		free(p->opt.device);
-		p->opt.device = NULL;
-	}
+	/*
+	 * This must not free p->opt.device; that must only be done
+	 * in pcap_close(), as this might be cleaning up after a
+	 * failed pcap_activate(), in which case p->opt.device must
+	 * still be valid, in case further actions are done on the
+	 * not-yet-activated pcap_t.
+	 *
+	 * See GitHub issue #1615.
+	 */
 	if (p->buffer != NULL) {
 		free(p->buffer);
 		p->buffer = NULL;
@@ -4242,6 +4247,17 @@ void
 pcap_close(pcap_t *p)
 {
 	p->cleanup_op(p);
+
+	/*
+	 * Free information set by pcap_create() *after* calling
+	 * the module's cleanup routine; that routine might have
+	 * to use p->opt.device (see commit
+	 * e333a6044f7d2d3225a6a22205b6b7c1e389945f).
+	 */
+	if (p->opt.device != NULL) {
+		free(p->opt.device);
+		p->opt.device = NULL;
+	}
 	free(p);
 }
 


guyharris:
I haven't yet tested to see whether this fixes #1615; I'll do this once I have a version that works on all platforms. Once I have a version that I know fixes #1615, I'll mark this as non-draft.

testprogs/activatetest.c
```diff
@@ -44,6 +44,13 @@ int main(void)
 		        "FAIL: Unexpected error %d from pcap_activate().\n",
 		        err);
 	}
+
+	/*
+	 * And attempts to determine whether attributes can be set
+	 * on the pcap_t shouldn't crashafter a failed activate.
```
infrastation: "crash after" (also "creash" in the commit message)

guyharris: Fixed.

guyharris:
> I haven't yet tested to see whether this fixes #1615 

I have now, and: 

- without this fix, testprogs/activatetest crashes in `pcap_can_set_rfmon_linux()`;
- with this fix, it doesn't crash;

so, yes, it fixes it.

PR #1619, opened 2026-02-04
DAG NULL dereference etc.

infrastation:
@guyharris, this moves the ctype-style macros to `pcap-int.h` and includes the latter from rpcapd source, if you think this should be done differently, please say.

### DIFF
--- a/pcap-dag.c
+++ b/pcap-dag.c
@@ -1389,12 +1389,12 @@ dag_stream_long_description(char *strbuf, const size_t strbufsize,
 	    RXTX_STR(stream),
 	    stream,
 	    dagbufsize / 1024 / 1024,
-	    inf ? dag_device_name(inf->device_code, 1) : "N/A");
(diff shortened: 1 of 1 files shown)


guyharris:
> @guyharris, this moves the ctype-style macros to `pcap-int.h` and includes the latter from rpcapd source, if you think this should be done differently, please say.

Perhaps there should be a top-level header "pcap-ctype.h" containing those macros, to separate "this is internal to the libpcap library" from "this is internal to the libpcap project" - the latter includes rpcapd. That's similar to, for example, "portability.h".

(Or maybe it should be "li-ctype.h" - "li" standing for "locale-independent" - and the macros renamed to `LI_XXX()`.)

But this is fine for now; we can worry about cleanups such as that later.

infrastation:
I considered condensing the macros in `portability.h`, but it did not quite look right. Alright, this is going to be merged later today.

PR #1620, opened 2026-02-05
CI cross-compiling

infrastation:
Let's see if this works.


infrastation:
The cross-compiling part works as expected, this is going to be merged after all other checks have passed.

PR #1621, opened 2026-02-05
DO NOT MERGE: Implement new "ip flag" and "tcp flag" primitives.

infrastation:
This is one of the potential solutions discussed on tcpdump-workers in November 2025 instead of the changes in pull request #1210. Needs some feedback.


infrastation:
One problem with this revision is that the `ip flag` implementation does not check whether the packet is an IPv4 packet before testing the part of the packet where IPv4 header flags are in an IPv4 packet. This is trivial to fix, but this also indicates a gap in the new tests because the new tests pass. So this will require at least one more revision to address both problems.

infrastation:
This revision gets `ip flag` right. Apparently, one of the new apply filter tests would normally have failed due to this bug, but I had set the expected results incorrectly, which resulted in a false positive. Which yet another time proves that even seemingly simple changes have potential to be incorrect. To that end, this revision exercises the flags more intensively and the capture files have comments with packet decoding to make it easier to interpret the expected results.

infrastation:
Rebased on the current master branch. This revision uses C99 types in the new code, keeps all packet data loads 8-bit, replaces more code with declarations and matches TCP flags using the same style as in `gen_load_internal()`.

This still requires minor corrections in the man page before merging, but otherwise looks nearly ready.

fxlb:
I dont like very much `-cleared`.
Why not
`-set/-unset`
`-on/-off`
`-yes/-no`
`-y/-n`
?
The last three are shorter.

infrastation:
My message to tcpdump-workers on February 6th among other things explains the choice of string IDs. Because this will have to be fixed and maintained long-term, the design choices should be recorded long-term, so the mailing list is the best place to discuss the syntax. Please make a better suggestion there if you can.

infrastation:
Current status: this depends on a choice of labels for the flag states, as discussed in a bit more detail on tcpdump-workers in March 2026.

PR #1627, opened 2026-02-20
various improvements to handling of protocols and packet data loads

infrastation:
These are `tcphf` research by-products that look ready for merging, at least to me.

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -257,6 +257,8 @@ struct addrinfo {
 
 /*
  * Network layer protocol identifiers
+ * ITU-T Rec. X.263 (1998 E)
+ * ISO/IEC TR 9577:1999(E)
  */
 #ifndef ISO8473_CLNP
 #define ISO8473_CLNP		0x81
@@ -264,11 +266,14 @@ struct addrinfo {
 #ifndef	ISO9542_ESIS
 #define	ISO9542_ESIS		0x82
 #endif
-#ifndef ISO9542X25_ESIS
-#define ISO9542X25_ESIS		0x8a
-#endif
 #ifndef	ISO10589_ISIS
 #define	ISO10589_ISIS		0x83
+#endif
+#ifndef ISO9577_IPV6
+#define ISO9577_IPV6		0x8e
+#endif
+#ifndef ISO9577_IPV4
+#define ISO9577_IPV4		0xcc
 #endif
 
 #define ISIS_L1_LAN_IIH      15
@@ -286,13 +291,6 @@ struct addrinfo {
  * octet, see sections 9.5~9.13 of ISO/IEC 10589:2002(E).
  */
 #define ISIS_PDU_TYPE_MAX 0x1FU
-
-#ifndef ISO8878A_CONS
-#define	ISO8878A_CONS		0x84
-#endif
-#ifndef	ISO10747_IDRP
-#define	ISO10747_IDRP		0x85
-#endif
 
 // Same as in tcpdump/print-sl.c.
 #define SLIPDIR_IN 0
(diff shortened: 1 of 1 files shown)


infrastation:
Looks like the OS-specific `AF_INET6` in `DLT_PFLOG` will require a few more conditional tests.

gencode.c
```diff
@@ -1211,6 +1213,63 @@ port_pq_to_ipproto(compiler_state_t *cstate, const int proto, const char *kw)
 	bpf_error(cstate, ERRSTR_INVALID_QUAL, pqkw(proto), kw);
 }
 
+static uint8_t
+pq_to_ipproto(compiler_state_t *cstate, const u_char pqual)
```
guyharris: Given that we now require C99 or later, should that be `u_int8_t` rather than `u_char`?

Or is that a large enough code change that it should be done separately?

(I don't think it'd require a type change in the API; a type change in the API would run the risk of requiring the new C99 types to be defined by pcap.h including <stdint.h>, which might add stuff to the namespace for files that use pcap.h, which might cause some code to get compile errors.)

guyharris:
> Looks like the OS-specific `AF_INET6` in `DLT_PFLOG` will require a few more conditional tests.

Presumably similar t what we currently do for `DLT_NULL` and `DLT_LOOP` (which we don't need to do for live captures but *do* need t do when reading a savefile).

gencode.c
infrastation: This uses the type in `struct qual`. The main point here is to keep excess bit width and negative values out of scope. Could be switched to a C99 type before or after, I do not mind.

This brushes on a bigger problem: various pieces of code get an `int` where they use an unsigned 8-bit or 16-bit value only, or use an 8-bit value (ARCnet SC or LLC SAP), but get a hopefully-translatable 16-bit EtherType value and so on. I have been reducing this problem space for a while by switching to `u_char` for the values that come from a `struct qual` only, but some work still remains to be done in this direction before it becomes clear where unrelated code point spaces can and cannot clash.

infrastation:
Yes, and `DLT_ENC` is another example, and byte order is also a factor, but this problem already has a solution:
```
name => 'link_proto_ip_1LE_NULL',
name => 'link_proto_ip_2LE_NULL',
name => 'link_proto_ip_2BE_NULL',
name => 'link_proto_ip6_5LE_NULL',
name => 'link_proto_ip6_10LE_NULL',
name => 'link_proto_ip6_10BE_NULL',
name => 'link_proto_ip6_22BE_NULL',
name => 'link_proto_ip6_24LE_NULL',
name => 'link_proto_ip6_24BE_NULL',
name => 'link_proto_ip6_26LE_NULL',
name => 'link_proto_ip6_26BE_NULL',
name => 'link_proto_ip6_28LE_NULL',
name => 'link_proto_ip6_28BE_NULL',
name => 'link_proto_ip6_30LE_NULL',
name => 'link_proto_ip6_30BE_NULL',
```
That said, the only reason the new tests include `DLT_PFLOG` is because it has a variable-length prefix.

gencode.c
guyharris: > This uses the type in `struct qual`. The main point here is to keep excess bit width and negative values out of scope.

...and, at least with MSVC, getting warnings if somebody's passing something wider as an argument. (The only GCC/Clang-based compiler I know of that issues narrowing warnings are Apple's compiler's, which got a 64-32 warning when Apple went 64-bit; perhaps newer GCCs or Clangs have gotten a `-W` flag for narrowing implicit conversions).

> Could be switched to a C99 type before or after, I do not mind.

I'd say "after" at this point, so it doesn't pile more stuff into your work on this.

> This brushes on a bigger problem: various pieces of code get an `int` where they use an unsigned 8-bit or 16-bit value only, or use an 8-bit value (ARCnet SC or LLC SAP), but get a hopefully-translatable 16-bit EtherType value and so on. I have been reducing this problem space for a while by switching to `u_char` for the values that come from a `struct qual` only, but some work still remains to be done in this direction before it becomes clear where unrelated code point spaces can and cannot clash.

Unfortunately, `enum`s are just decorated `int`s rather than full-blown types. There may be `-W` flags that provoke warnings for implicit conversions between `enum` types.

infrastation:
This revision uses `uint8_t` for the new functions only. It obviously fails the tests for `icmp6[]` because IPv6 transport with `DLT_PFLOG` depends on the value of `AF_INET6`. It also fails the tests for `tcp[]`, `icmp[]`, `igrp[]` and `vrrp[]` with `DLT_PFLOG` on Haiku. Requires a bit more work.

infrastation:
And until `gen_load_internal()` stops generating the dead IPv6 branch in a commit after the new tests, the OS-specific `AF_INET6` value remains a part of the filter program, so each case of an IPv4 transport with `DLT_PFLOG` should originally stand for 7 test blocks (there are 7 unique permutations of (`AF_INET`, `AF_INET6`) and then reduce to 2 for `AF_INET` only. Which means this combinatorial space needs to be handled a bit differently. Fortunately, byte order is not a factor in this case because in `DLT_PFLOG` the AF is an 8-bit field.

infrastation:
Almost there, didn't get FreeBSD and macOS right in this revision.

infrastation:
Rebased, renamed a function and added another minor clean-up commit. This is going to stay up for review for another couple days because these changes concern `gen_load_a()`, which is a common code path for much of `gencode.c`.

infrastation:
And now it needs a type cast, of course.

PR #1631, opened 2026-02-23
Both AF_INET and AF_INET6 are mandatory.

infrastation:
Let's see whether this theory holds...

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -7987,12 +7987,6 @@ gen_less(compiler_state_t *cstate, int n)
 /*
  * This is for "byte {idx} {op} {val}"; "idx" is treated as relative to
  * the beginning of the link-layer header.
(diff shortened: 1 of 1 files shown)


guyharris:
> Let's see whether this theory holds...

I.e., the theory that all the platforms we build on - and probably all the platform we support - define `AF_INET6`? (I.e., the 1990s/early noughts called, they want their IPv4-only network APIs back?)

infrastation:
It would be surprising to discover this is not the case, but let's test it.

guyharris:
> It would be surprising to discover this is not the case, but let's test it.

Yes, it's just a theory until tested.

infrastation:
For reference, this builds fine for QNX and on Haiku.

PR #1632, opened 2026-02-25
CMake: Disable remote capture support on Windows by default

fxlb:
This feature is experimental and is not ready for production use.

Add a warning when enabling remote capture, as with Autoconf.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -501,14 +501,10 @@ endif()
 set(PCAP_TYPE "" CACHE STRING "Packet capture type")
 
 #
-# Default to having remote capture support on Windows and, for now, to
-# not having it on UN*X.
+# Disable remote capture support by default.
+# This feature is experimental and is not ready for production use.
 #
-if(WIN32)
-    option(ENABLE_REMOTE "Enable remote capture" ON)
-else()
-    option(ENABLE_REMOTE "Enable remote capture" OFF)
-endif(WIN32)
+option(ENABLE_REMOTE "Enable remote capture" OFF)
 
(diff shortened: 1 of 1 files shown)


guyharris:
That'll cause a warning when, for example, it's built with Npcap...

...but it *should* give that warning, at least until we consider the code checked, double-checked, and triple-checked (and given a clean bill of health by static analyzers and anything *else* that can be thrown at it).

fxlb:
```
#
# Default to having remote capture support on Windows and, for now, to
# not having it on UN*X.
#
if(WIN32)
    option(ENABLE_REMOTE "Enable remote capture" ON)
```
Should we default to OFF, even on Windows?

guyharris:
> Should we default to OFF, even on Windows?

@bonsaiviking? Remote capture support was enabled in WinPcap, and I *think* it's enabled in Npcap. If it is enabled in Npcap, then, were we to make it off-by-default even on Windows, your build process could presumably just override that.

infrastation:
Should default to off in CMake too, yes. And the new CMake warning, as well as the existing Autoconf warning, should be printed at the end of the output because currently it just scrolls out of view quickly, and should say the feature is experimental and is not ready for production use.

bonsaiviking:
I don't recall if we include it, but we can work with any changes you make.

fxlb:
> the existing Autoconf warning, should be printed at the end of the output because currently it just scrolls out of view quickly, and should say the feature is experimental and is not ready for production use.

Agreed, this change is PR #1635.

fxlb:
> Should default to off in CMake too, yes.

Update done.

PR #1633, opened 2026-02-25
rdmasniff: clean up some resource freeing issues.

### LINKED ISSUE #1300: ibv_open_device() not called before ibv_free_device_list()?
https://github.com/the-tcpdump-group/libpcap/blob/408e16ce185cb049da5c686d1bce5024cc8f43df/pcap-rdmasniff.c#L425C2-L425C33

According to the verbs.h API comment, Client code must open all devices it intends to use before calling ibv_free_device_list().

### LINKED ISSUE #1532: RDMA Sniffer "rdmasniff_cleanup" function resource cleanup order needs change.
A RES_MEMORY_ERROR is observed in the dmesg, when the tcpdump is capturing rdma packets from RoCE device, and the tcpdump is exited while the roce traffic was still flowing.  The error is as below,
> infiniband bnxt_re0: responder: (2) RTR      : (18) RES_MEMORY_ERROR

The same error is not obseved if the roce traffic has been stopped first and then the tcpdum exited.
Tcpdup and libpcap are built from the source and following are the details,
 ./tcpdump --version
tcpdump version 5.0.0-PRE-GIT
OpenSSL 1.1.1k  FIPS 25 Mar 2021
64-bit build, 64-bit time_t

Further analysis of the libpcap code shows that,
Currently the rdmasniff_cleanup function has the following implementaion,
`static void
rdmasniff_cleanup(pcap_t *handle)
{
        struct pcap_rdmasniff *priv = handle->priv;

        ibv_dereg_mr(priv->mr);
        ibv_destroy_flow(priv->flow);
        ibv_destroy_qp(priv->qp);
        ibv_destroy_cq(priv->cq);
        ibv_dealloc_pd(priv->pd);
        ibv_destroy_comp_channel(priv->channel);
        ibv_close_device(priv->context);
        free(priv->oneshot_buffer);

        pcapint_cleanup_live_common(handle);
}`

In libpcap we do ibv_dereg_mr first  in the "rdmasniff_cleanup" fucntion, while the flow and qps are still active. It appears that this is the reason for the above mentioned error. In general the flow and qps needs to be destroyed first before destroying the mr.  Application like perftest destroy the mr after they are done with destropying qp, cq.

A quick re-ordering of the clean up code as shown in the below patch, fixes the issue.
`diff --git a/pcap-rdmasniff.c b/pcap-rdmasniff.c
index ba4705b7..043e72d4 100644
--- a/pcap-rdmasniff.c
+++ b/pcap-rdmasniff.c
@@ -78,10 +78,10 @@ rdmasniff_cleanup(pcap_t *handle)
 {
        struct pcap_rdmasniff *priv = handle->priv;

-       ibv_dereg_mr(priv->mr);
        ibv_destroy_flow(priv->flow);
        ibv_destroy_qp(priv->qp);
        ibv_destroy_cq(priv->cq);
+       ibv_dereg_mr(priv->mr);
        ibv_dealloc_pd(priv->pd);
        ibv_destroy_comp_channel(priv->channel);
        ibv_close_device(priv->context);
--
` 

We would like to submit the above patch based on the feedback from the community/owners of the libpcap.
Please share your thoughts/opinions on the patch.

Thanks and Regards
Ananth

guyharris:
Add a routine rdmasniff_free_resoources() to free up all resources pointed to by the struct pcap_rdmasniff structure for an open device. Use it both when closing a pcap_t and when cleaning up after a failed attempt to activate a pcap_t.

Free those routines in the order specified in issue #1532.

Don't save the struct ibv_device * found in pcap_create(); we free up the  ibv_device array allocated by ibv_get_device_list() when we return from rdmasniff_create(), and that invalidates the entries in that array. See the ibv_get_device_list() documentation, and issue #1300.

Fixes #1532.

Fixes #1300.

Improve checking for ibv_ error returns and messages for those errors.

Improve handling of the port number - don't fetch or check it in the create routine, so that an invalid port number doesn't prevent the underlying device from being recognized as a valid RDMA device, but do fetch and check it in the activate routine, so we can report the error to the user.

### DIFF
--- a/pcap-rdmasniff.c
+++ b/pcap-rdmasniff.c
@@ -43,11 +43,14 @@
 #define IBV_FLOW_ATTR_SNIFFER	3
 #endif
 
-static const int RDMASNIFF_NUM_RECEIVES = 128;
+static const unsigned RDMASNIFF_NUM_RECEIVES = 128;
+/*
+ * This is used as a snapshot size, and the snapshot size is stored in
+ * an int, so we make this an int.
+ */
 static const int RDMASNIFF_RECEIVE_SIZE = 10000;
 
 struct pcap_rdmasniff {
-	struct ibv_device *		rdma_device;
 	struct ibv_context *		context;
 	struct ibv_comp_channel *	channel;
 	struct ibv_pd *			pd;
@@ -56,7 +59,6 @@ struct pcap_rdmasniff {
 	struct ibv_flow *               flow;
 	struct ibv_mr *			mr;
 	u_char *			oneshot_buffer;
-	unsigned long			port_num;
 	int                             cq_event;
 	u_int                           packets_recv;
 };
@@ -74,19 +76,55 @@ rdmasniff_stats(pcap_t *handle, struct pcap_stat *stat)
 }
 
 static void
+rdmasniff_free_resources(struct pcap_rdmasniff *priv)
+{
+	if (priv->flow) {
+		ibv_destroy_flow(priv->flow);
+		priv->flow = NULL;
+	}
+
+	if (priv->qp) {
+		ibv_destroy_qp(priv->qp);
+		priv->qp = NULL;
+	}
+
+	if (priv->cq) {
+		ibv_destroy_cq(priv->cq);
+		priv->cq = NULL;
+	}
+
+	if (priv->mr) {
+		ibv_dereg_mr(priv->mr);
+		priv->mr = NULL;
+	}
+
+	if (priv->pd) {
+		ibv_dealloc_pd(priv->pd);
+		priv->pd = NULL;
+	}
+
+	if (priv->channel) {
+		ibv_destroy_comp_channel(priv->channel);
+		priv->channel = NULL;
+	}
+
+	if (priv->context) {
+		ibv_close_device(priv->context);
+		priv->context = NULL;
+	}
+
+	if (priv->oneshot_buffer) {
+		free(priv->oneshot_buffer);
+		priv->oneshot_buffer = NULL;
+	}
+}
+
+static void
 rdmasniff_cleanup(pcap_t *handle)
 {
 	struct pcap_rdmasniff *priv = handle->priv;
 
-	ibv_dereg_mr(priv->mr);
-	ibv_destroy_flow(priv->flow);
-	ibv_destroy_qp(priv->qp);
-	ibv_destroy_cq(priv->cq);
-	ibv_dealloc_pd(priv->pd);
-	ibv_destroy_comp_channel(priv->channel);
-	ibv_close_device(priv->context);
-	free(priv->oneshot_buffer);
-
+	rdmasniff_free_resources(priv);
 	pcapint_cleanup_live_common(handle);
 }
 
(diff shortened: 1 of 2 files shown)


pcap-rdmasniff.c
```diff
@@ -73,20 +75,56 @@ rdmasniff_stats(pcap_t *handle, struct pcap_stat *stat)
 	return 0;
 }
 
+static void
+rdmasniff_free_resoources(struct pcap_rdmasniff *priv)
```
infrastation: there is a typo in "resources"

pcap-rdmasniff.c
```diff
@@ -199,34 +237,125 @@ rdmasniff_oneshot(u_char *user, const struct pcap_pkthdr *h, const u_char *bytes
...
+
+	namelen = strlen(handle->opt.device);
+
+	/*
+	 * The syntax of a name on which to capture is
+	 * device[:port].
+	 */
+	port = strchr(handle->opt.device, ':');
+	if (port) {
+		port_num = strtoul(port + 1, NULL, 10);
```
infrastation: Would the new `pcapint_get_decuint()` be a better fit?

pcap-rdmasniff.c
```diff
@@ -199,34 +237,125 @@ rdmasniff_oneshot(u_char *user, const struct pcap_pkthdr *h, const u_char *bytes
...
+		if (port_num > 0) {
+			namelen = port - handle->opt.device;
+		} else {
```
infrastation: How would this treat `name:` (with the colon, without a number)?

pcap-rdmasniff.c
```diff
@@ -392,34 +496,63 @@ rdmasniff_create(const char *device, char *ebuf, int *is_ours)
...
+	 */
 	port = strchr(device, ':');
 	if (port) {
+		/*
+		 * XXX - wil any RDMA device name ever have a colon in it?
+		 *
+		 * If so, if there's more than one port to specify on it,
+		 * we'll need to handle that.
+		 */
 		port_num = strtoul(port + 1, NULL, 10);
```
infrastation: This looks similar to the earlier block in `rdmasniff_activate()`. If the name and the port remain the same, would it be better to parse this exactly once and to keep `struct rdmasniff` as it is?

pcap-rdmasniff.c
guyharris: Fixed.

pcap-rdmasniff.c
```diff
@@ -392,34 +496,63 @@ rdmasniff_create(const char *device, char *ebuf, int *is_ours)
...
 	namelen = strlen(device);
 
+	/*
+	 * The syntax of a name on which to capture is
+	 * device[:port].
+	 */
 	port = strchr(device, ':');
 	if (port) {
+		/*
+		 * XXX - wil any RDMA device name ever have a colon in it?
```
fxlb: will ?

pcap-rdmasniff.c
guyharris: The same way as it always did, so there might be a bug there.

C99 says on page 310 that `strtoul()` and other functions

> If the subject sequence is empty or does not have the expected form, no conversion is performed; the value of `nptr` is stored in the object pointed to by `endptr`, provided that `endptr` is not a null pointer.

Unfortunately, `endptr` is NULL here, so we can't test for that. However, C99 also says, on the same page, under "Returns" that "If no conversion could be performed, zero is returned.", so presumably it returns 0 for `port_num`, and it defaults to port 1.

pcap-rdmasniff.c
guyharris: Yeah, we could save the results of the parsing and reuse them.

pcap-rdmasniff.c
guyharris: Fixed.

fxlb:
In the commit message:
s/rdmasniff_free_resoources9/rdmasniff_free_resources/

pcap-rdmasniff.c
infrastation: It would be wrong to treat an invalid input as a valid input.

guyharris:
> In the commit message: s/rdmasniff_free_resoources9/rdmasniff_free_resources/

Fixed.

pcap-rdmasniff.c
guyharris: Yes - I discovered the use-after-free issue while `pcapint_get_decuint()`ifying a bunch of modules, but decided to make this change separately from the `pcapint_get_decuint()`ifying, which I'll do in a subsequent change.

(Is this a change that should be backported? It fixes some significant bugs, but it's also a somewhat significant change.)

pcap-rdmasniff.c
guyharris: > It would be wrong to treat an invalid input as a valid input.

Yes, but it's worked that way since Day One (f5f1484a931851d2d3b67d6033aedea8da5acb42).

So the question is "what, if anything, would break if `iwarp0:` no longer meant 'port 1 of device `iwarp0`?"

If the answer is "nothing" or "so little that they could just adapt and explicitly say `iwarp0:1` - or just use `iwarp0`, as port 1 is the default port - without any real disruption", we might as well reject that.

guyharris: The original pull request for RDMA sniffer support is #585.

guyharris: And I've asked some questions in that pull request.

guyharris:
@rolandd do you have any comments on this change or on any of the issues it's intended to fix?

guyharris:
@paravmellanox do you have any comments on this change or on any of the issues it's intended to fix?

pcap-rdmasniff.c
```diff
@@ -199,34 +237,125 @@ rdmasniff_oneshot(u_char *user, const struct pcap_pkthdr *h, const u_char *bytes
 	*sp->pkt = priv->oneshot_buffer;
 }
 
+static struct ibv_device *
+rdma_find_device_in_list(struct ibv_device **dev_list, int numdev,
```
paravmellanox, line 281: if dev_list can be made const, it will be good.

pcap-rdmasniff.c
```diff
@@ -199,34 +237,125 @@ rdmasniff_oneshot(u_char *user, const struct pcap_pkthdr *h, const u_char *bytes
...
 	struct ibv_qp_init_attr qp_init_attr;
 	struct ibv_qp_attr qp_attr;
 	struct ibv_flow_attr flow_attr;
 	struct ibv_port_attr port_attr;
-	int i;
 
-	priv->context = ibv_open_device(priv->rdma_device);
+	dev_list = ibv_get_device_list(&numdev);
+	if (!dev_list) {
+		if (errno == EPERM) {
```
paravmellanox: this is unlikely to happen as this mainly reads the /sys/class/infiniband files. but does not hurt either.

pcap-rdmasniff.c
```diff
@@ -199,34 +237,125 @@ rdmasniff_oneshot(u_char *user, const struct pcap_pkthdr *h, const u_char *bytes
...
+		} else {
+			pcapint_fmt_errmsg_for_errno(handle->errbuf,
+				      PCAP_ERRBUF_SIZE, errno,
+				      "ibv_get_device_list() failed for %s: %s",
+				      handle->opt.device, strerror(errno));
+			ret = PCAP_ERROR;
+		}
+		goto error;
+	}
+	if (!numdev) {
```
paravmellanox, line 346: it would be more stronger to see dev_list as null pointer check to call ibv_free_device_list(), so even if for some reason, if the library returned a list with zero devices, it can still free up the memory.

pcap-rdmasniff.c
infrastation: For severe problems only, obviously correct bug fixes only, or we risk delaying 1.11.0 even more.

pcap-rdmasniff.c
infrastation: Isn't a valid empty list of devices a `NULL` pointer?

guyharris: > Isn't a valid empty list of devices a `NULL` pointer?

No. According to [the RDMAmojo documentation for `ibv_get_device_list()`](https://www.rdmamojo.com/2012/05/31/ibv_get_device_list/):

> **ibv_get_device_list()** returns the array of available RDMA devices on success, returns NULL and sets **errno** if the request fails. If no devices are found, then num_devices is set to 0, and non-NULL is returned.

and according to [the Linux man page for `ibv_get_device_list()`](https://man7.org/linux/man-pages/man3/ibv_get_device_list.3.html):

> **ibv_get_device_list()** returns a NULL-terminated array of RDMA devices currently available.

I.e., if there are no devices, a pointer to a one-element array of `struct ibv_device *` is returned, with the one and only element of that array being a null pointer.

guyharris: > it would be more stronger to see dev_list as null pointer check to call ibv_free_device_list(), so even if for some reason, if the library returned a list with zero devices, it can still free up the memory.

What are you suggesting here? If the library returns a list with zero devices, then the return value of `ibv_get_device_list()` is non-null, and `numdev` is set to 0. In that case, we free up the list.

pcap-rdmasniff.c
guyharris: Or not bother parsing the port number in the create routine, so that we accept anything of the form <known_rdma_device>:<anything>, and then parse it in the activate routine and return an error if the port number isn't valid, so if somebody does rdma17:ffffffff they get an error from pcap-rdmasniff.c pointing that out, rather than having pcap-rdmasniffer.c not accept it as an RDMA device - even if rdma17 is a valid RDMA device - so that it probably ends up getting opened as a regular networking device and probably getting an error claiming that rdma17:ffffffff isn't a known device.

That's what I've done.

pcap-rdmasniff.c
guyharris: Unfortunately, if you make it `const struct ibv_device **dev_list`, then you have to make  `dev_list` in a `const struct ibv_device *dev_list` in the routines that call it, and, if you do tha, you'll get warnings from the compiler about `ibv_free_device_list()` being passed a `const struct ibv_device **list`, because you're throwing away the constness.

(Fun with C data types. Some way of saying "you can't modify this, but you *can* throw it away" would be nice.)

paravmellanox: @guyharris  right. current code is ok.

pcap-rdmasniff.c
paravmellanox: @guyharris I am suggesting that 

if (dev_list)
   ibv_free_device_list()

guyharris: In all of the places where `ibv_free_device_list()` is called, it has been established that `dev_list` is non-null.

paravmellanox: @guyharris ok thanks.

pcap-rdmasniff.c
guyharris: Might as well catch unexpected errors and report them, so we can figure out what's causing them and either fixing them or, perhaps, ignoring them if they're unfixable.

guyharris:
Added CHANGES items.

infrastation:
Looks like on linux-mips libibverbs did not flip the error code sign the correct number of times (that's the negation of `ENOSYS`):
```
Error in pcap_findalldevs: ibv_get_device_list() failed: Unknown error -89
```

guyharris:
> Looks like on linux-mips libibverbs did not flip the error code sign the correct number of times (that's the negation of `ENOSYS`):
> 
> ```
> Error in pcap_findalldevs: ibv_get_device_list() failed: Unknown error -89
> ```

What's wrong with this picture?

- https://github.com/linux-rdma/rdma-core/blob/master/libibverbs/ibdev_nl.c#L197
- https://github.com/linux-rdma/rdma-core/blob/master/libibverbs/ibdev_nl.c#L214

Consistency is Job Minus-One.

guyharris:
Although that's `EOPNOTSUPP`, not `ENOSYS`. There may be similar issues elsewhere, including ones with `ENOSYS`.

infrastation:
On that host `ENOSYS` is 89:
```
$ errno 89
ENOSYS 89 Function not implemented
$ grep 89 /usr/include/asm/errno.h 
#define ENOSYS		89	/* Function not implemented */
```

guyharris:
> On that host `ENOSYS` is 89:
> 
> ```
> $ errno 89
> ENOSYS 89 Function not implemented
> $ grep 89 /usr/include/asm/errno.h 
> #define ENOSYS		89	/* Function not implemented */
> ```

"That" in "that's `EOPNOTSUPP`, not `ENOSYS`" refers to "that routine that inconsistently returned a negated `EOPNOTSUPP` for one error and returned a non-negated `EINVAL` for other errors", i.e. that libibverbs isn't exactly consistent about errno values versus the negation of errno values. (I'll file an issue on rdma-core about that - and any other such inconsistencies I find.)

The right thing to do is probably to check for both positive *and* negative values of various "not supported" errno values - `EOPNOTSUPP`, `ENOSYS`, and `ENOTSUPP` (not to be confused with `ENOTSUP`, which, on some platforms, is defined as an alias for `EOPNOTSUPP`; Linux errno weirdness probably dates back to attempts to provide system call name and number compatibility with existing UN\*Xes on various platforms). I'm in the process of doing that.

guyharris:
> (I'll file an issue on rdma-core about that - and any other such inconsistencies I find.)

I didn't find any other obvious signedness inconsistencies *in the current master branch of rdma-core*.

There may have been code paths that would result in `errno` being set to `-ENOSYS` in the past, causing the problem that shows up on the linux-mips machine, so, given the rdma-core folks' careful attention to consistency, we should check for both normal and negated `errno` values in this case, as the current version of this code does - and treat `ENOSYS` (positive or negative) and `EOPNOTSUPP` (positive or negative) as "this kernel doesn't support RDMA devices" errors, as the current version of this code does.

infrastation:
That's libibverbs 22.1 from 2019 that returns that. If you leave a comment next to the workaround, it will be easier to get rid of it in future.

guyharris:
> That's libibverbs 22.1 from 2019 that returns that. If you leave a comment next to the workaround, it will be easier to get rid of it in future.

Done, although releases with the problem appear to have been made as late as 2022-10.

guyharris:
> > Looks like on linux-mips libibverbs did not flip the error code sign the correct number of times (that's the negation of `ENOSYS`):
> > ```
> > Error in pcap_findalldevs: ibv_get_device_list() failed: Unknown error -89
> > ```
> 
> What's wrong with this picture?
> 
> * https://github.com/linux-rdma/rdma-core/blob/master/libibverbs/ibdev_nl.c#L197
> * https://github.com/linux-rdma/rdma-core/blob/master/libibverbs/ibdev_nl.c#L214

Pull request linux-rdma/rdma-core#1712 submitted.

PR #1635, opened 2026-02-27
Autoconf: Move the --enable-remote warning at the end of configure sc…

fxlb:
…ript

It will be better seen at the end of the output.
Update the warning text.


infrastation:
This should convey the message better.

guyharris:
BTW, should the same be done with tcpdump and the SMB dissector?

fxlb:
> BTW, should the same be done with tcpdump and the SMB dissector?

Yes, I was planning to do that next.

fxlb:
Backported to 1.10.

Issue #1634, opened 2026-02-27
Misaligned Memory Access in USB Header Processing


FuzzAnything-ORG:
Dear libpcap developers,

I found an UndefinedBehaviorSanitizer (UBSan) error when parsing pcapng files with USB head processing in the `pcap_next_ex` API.

## Version
Found in: libpcap-1.8.1-4138-g0ddbca0d
Reproducible in main: libpcap-1.8.1-4223-ge0b74e21

## Description

The function `pcapint_post_process` in `pcap-util.c` performs an unaligned memory access when processing packet data from USB Linux mmapped captures. At line 543-544, the code casts a byte pointer (`u_char *data`) directly to a `pcap_usb_header_mmapped*` pointer:

```c
const pcap_usb_header_mmapped *usb_hdr =
    (const pcap_usb_header_mmapped *) data;
```

The `pcap_usb_header_mmapped` structure (defined in `pcap/usb.h`) requires 8-byte alignment because its first member is `uint64_t id`. However, the packet data buffer (`data`) may not be 8-byte aligned, particularly when reading from pcapng files where Enhanced Packet Block (EPB) data offsets are only guaranteed to be 4-byte aligned.

When the code subsequently accesses structure members at line 567:

```c
if (usb_hdr->ndesc > USB_MAXDESC)
    return;
```

## Reproduction Steps

1. Save the crash input file (`crash_data.h` contains the raw bytes, or use `crash-poc` binary file)

2. Compile the PoC:
```bash
clang++ -g -O1 -fno-omit-frame-pointer -fsanitize=address,undefined \
    -I<path-to-libpcap>/include \
    poc.cpp -o poc \
    <path-to-libpcap>/lib/libpcap.a
```

3. Run with UBSan stack trace enabled:
```bash
export UBSAN_OPTIONS="halt_on_error=1:print_stacktrace=1"
./poc
```

## Stack Trace

```
$ export UBSAN_OPTIONS="halt_on_error=1:print_stacktrace=1"
$ ./poc
/root/src/libpcap/pcap-util.c:567:16: runtime error: member access within misaligned address 0x51d00000009c for type 'const pcap_usb_header_mmapped' (aka 'const struct _usb_header_mmapped'), which requires 8 byte alignment
0x51d00000009c: note: pointer points here
[...]

PR #1639, opened 2026-03-01
Don't assume Linux USB headers are 64-bit aligned.

guyharris:
They might not be, especially in pcapng capture files, which only guarantee 32-bit alignment for packet data.

Fixes GitHub issue #1634.

Discovered by FuzzAnything Organization (https://github.com/FuzzAnything-ORG) (fuzzanything@gmail.com).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2843,6 +2843,12 @@ if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/.devel OR EXISTS ${CMAKE_BINARY_DIR}/.deve
         # XXX - add /Qspectre if that is really worth doing.
         #
         check_and_add_compiler_option(-wd5045)
+        #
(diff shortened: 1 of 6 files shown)


guyharris:
The PoC reports a misaligned access when built with an unmodified version of the tip of the main branch of libpcap, but does not report one when built with a version with this change.

guyharris:
Both the old code and the new code, when used in tcpdump, result in a big-endian USB capture being transformed into the same (bit-for-bit identical) little-endian USB capture when tcpdump reads the big-endian file and writes to a file on my little-endian Mac.

guyharris:
Reverted and then recommitted in 61542941feff19af0f801caafc690351450e4e78, to add the "Discovered by..." line to the commit message (is there some way to update a pull/merge request's initial comment *and* the corresponding commit message in one operation, so they don't get out of sync?).

PR #1640, opened 2026-03-05
Fix memory leak in daemon_unpackapplyfilter() error paths

nidu-ninja:
Fix a memory leak in daemon_unpackapplyfilter() where dynamically
allocated BPF instruction memory was not freed on early-return
error paths.

This change introduces a structured cleanup block to ensure
allocated memory is released when instruction reception,
validation, or filter application fails.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -66,6 +66,8 @@ DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
       rpcapd: Refine SSL options in printusage().
       Fix a possible buffer overflow (Coverity CID 1619148).
       Fix parameter name validation in the configuration file.
+      Fix a memory leak in daemon_unpackapplyfilter() (Coverity CID
+        1641537).
     Documentation:
       Add a README.hurd.md file.
       Cross-reference some man pages better.
--- a/rpcapd/daemon.c
+++ b/rpcapd/daemon.c
@@ -2342,6 +2342,7 @@ daemon_unpackapplyfilter(PCAP_SOCKET sockctrl, SSL *ctrl_ssl, struct session *se
 	struct bpf_insn *bf_insn;
 	struct bpf_program bf_prog;
 	unsigned int i;
+	int ret;
 
 	status = rpcapd_recv(sockctrl, ctrl_ssl, (char *) &filter,
 	    sizeof(struct rpcap_filter), plenp, errmsgbuf);
@@ -2385,11 +2386,13 @@ daemon_unpackapplyfilter(PCAP_SOCKET sockctrl, SSL *ctrl_ssl, struct session *se
 		    sizeof(struct rpcap_filterbpf_insn), plenp, errmsgbuf);
 		if (status == -1)
 		{
-			return -1;
+			ret = -1;
+			goto cleanup;
 		}
 		if (status == -2)
 		{
-			return -2;
+			ret = -2;
+			goto cleanup;
 		}
 
 		bf_insn->code = ntohs(insn.code);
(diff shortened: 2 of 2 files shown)


infrastation:
Please put the commit message into the commit rather than request description and keep the existing indentation style.

rpcapd/daemon.c
```diff
@@ -2336,86 +2336,95 @@ daemon_msg_endcap_req(uint8_t ver, struct daemon_slpars *pars,
 static int
 daemon_unpackapplyfilter(PCAP_SOCKET sockctrl, SSL *ctrl_ssl, struct session *session, uint32_t *plenp, char *errmsgbuf)
 {
-	int status;
```
guyharris, line 2339: As @infrastation said, please don't change the indentation style - it makes it difficult to see what the real changes are.

CHANGES
```diff
@@ -19,6 +19,7 @@ DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
         (EOA) as of December 31, 2017
       Make pcap_compile() error messages more uniform and consistent.
       Deprecate pcap_compile_nopcap().
+      Deprecate bpf_filter().
```
guyharris: That's a change for 1.10.7, not 1.11.0.

CHANGES
```diff
@@ -28,6 +29,7 @@ DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
       Fix DECnet packet filtering on big-endian hosts.
       Fix various failures to reject invalid DECnet primitives.
       Require "vpi" and "vci" values to be within valid ranges.
+      Initialize the scratch memory store to 0.
```
guyharris: This is the only change that you should be making to the CHANGES file.

CHANGES
```diff
@@ -52,9 +54,6 @@ DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
       In "gateway" negate the host(s), but not the protocol.
       Reject "gateway" within MPLS, VXLAN or Geneve.
       In "net <n> mask <m>" catch ENOMEM for the "m" too.
-      Match both byte orders for the AF_ value when filtering DLT_NULL
```
guyharris, line 54: This should not be removed.

CHANGES
```diff
@@ -76,8 +75,9 @@ DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
       Discuss Linux BPF extensions in the man pages.
       Note endianness in pcap_compile(3PCAP) and pcap_lookupnet(3PCAP).
       man: Document devices, interfaces and "any" better.
+      Remove list of OSes that support "ipv6-icmp"; all the ones we
```
guyharris: This is another 1.10.7 change that should not be added here.

CHANGES
```diff
@@ -76,8 +75,9 @@ DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
...
+        support appear to do so.
       Add a README.qnx.md file.
-      Improve dcumentation of TLS supprt in rpcapd man page.
```
guyharris, line 79: Don't remove existing CHANGES entries.

CHANGES
```diff
@@ -159,29 +159,6 @@ DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
     QNX:
       Disable zero-copy BPF to work around portability issues.
       Use "unix.h" instead of the missing <sysexits.h>.
-    RDMA:
```
guyharris, line 158: Don't remove existing CHANGES entries and don't remove the 1.10.7 set of changes.

fxlb:
Whitespaces change problem. Please fix.

infrastation:
After the clean-ups the code makes much more sense, thank you.  Please wrap the commit message to 72 characters, also there is no need to state the obvious ("Keep existing daemon.c indentation style and make only the relevant CHANGES entry for this fix").

guyharris:
> Fix a memory leak in daemon_unpackapplyfilter() where dynamically allocated BPF instruction memory was not freed on early-return error paths.

It wasn't freed on *any* return paths, ao it would leak memory even on *success* (when `pcap_setfilter()` returns, either the filter has been copied into kernel space or has been copied to a newly-allocated user-mode structure, so, if the tilter code has been allocated, it can and must be freed). Your change fixes that as well.

So it should say

> Fix a memory leak in daemon_unpackapplyfilter() where dynamically allocated BPF instruction memory was not freed before returning.

guyharris:
And the second paragraph just gives the details of how the leak was fixed; it can either be removed or changed to  be

> This change introduces a structured cleanup block to ensure allocated memory is released.

PR #1641, opened 2026-03-06
fuzz: fix stale daemon_serviceloop() prototype in fuzz_rserver

OwenSanzas:
## Problem

`fuzz_rserver.c` independently declares `daemon_serviceloop()` with 5 parameters, but the actual function in `rpcapd/daemon.c` has had 6 parameters since commit 493e050f (Aug 2022), when `char *data_port` was added. The fuzzer was written in 2019 (commit c4510b8b) and was never updated.

Including the real header confirms the conflict:

```
$ cc -I. -Ibuild -DHAVE_CONFIG_H -include rpcapd/daemon.h \
     -c testprogs/fuzz/fuzz_rserver.c

error: conflicting types for 'daemon_serviceloop'; have 'int(int, int, char *, int, int)'
note: previous declaration with type 'int(int, int, char *, int, char *, int)'
```

## Undefined behavior

The fuzzer calls `daemon_serviceloop(sock, 1, malloc(0), 1, 0)` — 5 arguments for a 6-argument function. On x86-64 System V ABI, the 6th integer/pointer argument is passed in r9. Since the caller only sets rdi–r8, the callee reads `uses_ssl` from r9, which is never set by the caller. This is undefined behavior per C11 §6.5.2.2.

Disassembly of the original call site confirms r9 is not set:

```asm
xor    %r8d,%r8d              ; arg5 (r8):  0 — caller thinks this is uses_ssl
call   daemon_serviceloop     ; arg6 (r9):  NOT SET
```

GDB on a non-sanitizer build shows r9 contains garbage at entry:

```
(gdb) break daemon_serviceloop
(gdb) run
(gdb) info registers r9
r9    0x555555fe7ae0    93825003322080
```

When built with OpenSSL (`HAVE_OPENSSL`), a nonzero `uses_ssl` causes `daemon_serviceloop` to enter the TLS handshake path and allocate OpenSSL structures that are never freed.

## Fix

Replace the hand-written forward declaration with `#include "rpcapd/daemon.h"` (as suggested by @fxlb and @guyharris), and pass `NULL` for `data_port` and `0` for `uses_ssl`.

Disassembly of the fixed call site confirms r9 is now explicitly zeroed:

```asm
xor    %r8d,%r8d              ; arg5 (r8):  data_port = NULL
xor    %r9d,%r9d              ; arg6 (r9):  uses_ssl = 0
call   daemon_serviceloop
```

### DIFF
--- a/testprogs/fuzz/fuzz_rserver.c
+++ b/testprogs/fuzz/fuzz_rserver.c
@@ -34,7 +34,7 @@ void rpcapd_log(log_priority priority, const char *message, ...)
 }
 
 void sock_initfuzz(const uint8_t *Data, size_t Size);
-int daemon_serviceloop(int sockctrl, int isactive, char *passiveClients, int nullAuthAllowed, char *data_port, int uses_ssl);
+#include "rpcapd/daemon.h"
 
 int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
     int sock;


testprogs/fuzz/fuzz_rserver.c
```diff
@@ -34,7 +34,7 @@ void rpcapd_log(log_priority priority, const char *message, ...)
 }
 
 void sock_initfuzz(const uint8_t *Data, size_t Size);
-int daemon_serviceloop(int sockctrl, int isactive, char *passiveClients, int nullAuthAllowed, int uses_ssl);
+int daemon_serviceloop(int sockctrl, int isactive, char *passiveClients, int nullAuthAllowed, char *data_port, int uses_ssl);
```
fxlb: Or:
#include "rpcapd/daemon.h"

OwenSanzas: Thanks for the guidance, change is done.

OwenSanzas:
Thanks for the guidance. It compiles cleanly and ensures the prototype stays in sync automatically if the signature changes again.                                                                                 
                                                                                                                                                                                                                                                              
Disassembly confirms the fix: the call site now explicitly sets r9 (xor %r9d,%r9d), so uses_ssl is guaranteed to be 0 instead of reading uninitialized register contents.

guyharris:
`daemon_serviceloop()` passes `data_port` to `daemon_msg_startcap_req()`. `daemon_msg_startcap_req()` assumes it's non-null; instead, "use a dynamically-chosen data port" is indicated by passing a null string - `""`.

See #1645.

guyharris:
> `daemon_serviceloop()` passes `data_port` to `daemon_msg_startcap_req()`. `daemon_msg_startcap_req()` assumes it's non-null; instead, "use a dynamically-chosen data port" is indicated by passing a null string - `""`.
> 
> See #1645.

If you want, you can just replace the `NULL` with `""` and update the comment - I'll close my equivalent pull request.

guyharris:
Or maybe I should just change the code to treat a null `data_port` as equivalent to an empty `data_port`.

infrastation:
It would be best to squash this into one commit.

guyharris:
Add

    Fixes #1642.

to the commit message; I think that'll cause GitHub to mark #1642 as fixed when this pull request is merged.

guyharris:
> Add
> 
> ```
> Fixes #1642.
> ```
> 
> to the commit message

...and to the first comment.

guyharris:
> Or maybe I should just change the code to treat a null `data_port` as equivalent to an empty `data_port`.

Done. Thanks for provoking me to check some bits of the rpcapd code and clean them up!

OwenSanzas:
Oh sorry for my late reply... Thanks for helping me with this. I really learnt a lot from your discussion.

PR #1659, opened 2026-03-20
plugin: add runtime plugin loader for external capture backends including dpdk

vjardin:
Dataplanes like DPDK expose unstable APIs that change across releases. Compiling backends for these dataplanes directly into libpcap ties the two release cycles together: a DPDK API change breaks the libpcap build, and fixing it requires a coordinated update of both projects. The same problem applies to any fast-moving dataplane (grout, VPP, etc.) whose capture interface evolves independently from libpcap.

Add a generic plugin loader (pcap-plugin.c) that discovers external capture backends at runtime. Each backend ships as a pcap-*.so shared module in a well-known directory. The loader dlopen's every matching file, looks up a "pcap_plugin_entry" symbol, validates the ABI version, and dispatches findalldevs/create calls through it. Backends are found in $PCAP_PLUGIN_DIR (colon-separated), /usr/lib/pcap/plugins, and /usr/local/lib/pcap/plugins.

Because pcap-int.h is not installed, plugins cannot access the pcap_t struct directly. A set of pcap_plugin_* accessor functions is exported with default visibility: handle allocation, ops registration, field getters/setters, BPF helpers, and device enumeration. This gives plugins a stable ABI (versioned via PCAP_PLUGIN_ABI_VERSION) without exposing libpcap internals.

With this mechanism, a dataplane project builds its own pcap-*.so against its own headers and ships it alongside its packages. Upgrading the dataplane only requires rebuilding the plugin, not libpcap. If the plugin is absent or fails to load, libpcap keeps working — no compile time dependency, no conditional #ifdefs.

### DIFF
--- a/pcap-plugin.c
+++ b/pcap-plugin.c
@@ -426,6 +426,30 @@ pcap_plugin_get_tstamp_type(pcap_t *p)
 }
 
 int
+pcap_plugin_get_tstamp_precision(pcap_t *p)
+{
+	return p->opt.tstamp_precision;
+}
+
+int
+pcap_plugin_get_promisc(pcap_t *p)
+{
+	return p->opt.promisc;
+}
+
+int
+pcap_plugin_get_buffer_size(pcap_t *p)
+{
+	return (int)p->opt.buffer_size;
+}
+
+int
+pcap_plugin_get_immediate(pcap_t *p)
+{
+	return p->opt.immediate;
+}
+
+int
 pcap_plugin_set_tstamp_type_list(pcap_t *p, const int *types, int count)
 {
 	u_int *list;
@@ -439,6 +463,47 @@ pcap_plugin_set_tstamp_type_list(pcap_t *p, const int *types, int count)
 	p->tstamp_type_list = list;
 	p->tstamp_type_count = count;
 	return 0;
+}
+
+int
+pcap_plugin_set_tstamp_precision_list(pcap_t *p, const int *precisions,
+    int count)
+{
+	u_int *list;
+
+	list = malloc(count * sizeof(u_int));
+	if (list == NULL)
+		return -1;
+	for (int i = 0; i < count; i++)
+		list[i] = (u_int)precisions[i];
+	free(p->tstamp_precision_list);
+	p->tstamp_precision_list = list;
+	p->tstamp_precision_count = count;
+	return 0;
+}
+
+int
+pcap_plugin_set_datalink_list(pcap_t *p, const int *dlts, int count)
+{
+	u_int *list;
+
+	list = malloc(count * sizeof(u_int));
+	if (list == NULL)
+		return -1;
+	for (int i = 0; i < count; i++)
+		list[i] = (u_int)dlts[i];
+	free(p->dlt_list);
+	p->dlt_list = list;
+	p->dlt_count = count;
+	return 0;
+}
+
+void
+pcap_plugin_set_selectable_fd(pcap_t *p, int fd)
+{
+#ifndef _WIN32
+	p->selectable_fd = fd;
+#endif
 }
 
 pcap_if_t *
--- a/pcap-plugin.c
+++ b/pcap-plugin.c
@@ -419,6 +419,28 @@ pcap_plugin_filter(const struct bpf_insn *pc, const unsigned char *pkt,
 	return pcapint_filter(pc, pkt, wirelen, caplen);
 }
 
+int
+pcap_plugin_get_tstamp_type(pcap_t *p)
+{
+	return p->opt.tstamp_type;
+}
+
+int
+pcap_plugin_set_tstamp_type_list(pcap_t *p, const int *types, int count)
+{
+	u_int *list;
+
+	list = malloc(count * sizeof(u_int));
+	if (list == NULL)
+		return -1;
+	for (int i = 0; i < count; i++)
+		list[i] = (u_int)types[i];
+	free(p->tstamp_type_list);
+	p->tstamp_type_list = list;
+	p->tstamp_type_count = count;
+	return 0;
+}
+
 pcap_if_t *
 pcap_plugin_add_dev(pcap_if_list_t *devlistp, const char *name,
     unsigned int flags, const char *description, char *errbuf)
--- a/pcap-plugin.c
+++ b/pcap-plugin.c
@@ -1,0 +1,428 @@
+/*
+ * Copyright (c) 2026 Vincent Jardin, Free Mobile, Iliad
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   1. Redistributions of source code must retain the above copyright
+ *      notice, this list of conditions and the following disclaimer.
+ *   2. Redistributions in binary form must reproduce the above copyright
+ *      notice, this list of conditions and the following disclaimer in the
+ *      documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS''AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+
+/*
+ * pcap-plugin: generic plugin loader for libpcap capture backends.
+ *
+ * Scans plugin directories for pcap-*.so shared modules, loads each
+ * via pcapint_load_code(), and dispatches findalldevs/create calls to
+ * loaded plugins.
+ *
+ * Security model (modeled after OpenSSL provider loading):
+ *   - $PCAP_PLUGIN_DIR is read through secure_getenv() (or equivalent),
+ *     so it is automatically ignored under elevated privileges (setuid,
+ *     setgid, file capabilities, LSM transitions).
+ *   - The hardcoded plugin directory (PCAP_PLUGIN_DIR) is always scanned.
+ *   - Filesystem permissions on the plugin directory are the security
+ *     boundary, same as PAM, NSS, OpenSSL, and Mesa.
+ *   - libpcap does NOT drop privileges itself; that is the app's job.
+ *
+ * Plugin search order:
+ *   1. $PCAP_PLUGIN_DIR (colon-separated, skipped under privilege)
+ *   2. PCAP_PLUGIN_DIR (compile-time default, e.g. /usr/lib/pcap/plugins)
+ */
+
+#include <config.h>
+
+#ifdef HAVE_SECURE_GETENV
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE	/* for secure_getenv() */
+#endif
+#endif
+
+#include <stdarg.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#ifndef _WIN32
+#include <dirent.h>
+#include <dlfcn.h>
+#include <unistd.h>
+
+#ifdef HAVE_GETAUXVAL
+#include <sys/auxv.h>
+#endif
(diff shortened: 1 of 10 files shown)


guyharris:
There's already a code-loading API, `pcapint_load_code()`, in pcap.c.

It currently has only a Windows implementation, as it was only used to load AirPcap support code, allowing the same code to work regardless of whether AirPcap support was installed. It's *currently* unused in the main branch, as we removed AirPcap support (Riverbed's AirPcap devices are no longer supported by the vendor).

You should probably add UN\*X support to that, and use it.

guyharris:
Loading plugins should be done with **_great_** care, as opening capture devices requires elevated privileges on many platforms, which means you may be loading plugins into a program running with elevated privileges.

At minimum, plugin-loading code should *require* that the plugin be loaded from a specific trusted directory (or maybe one of a small set of trusted directories) and that neither the directory nor the plugin file be writable by unprivileged users.

There may be - in fact, there probably *are* - more things to worry about.

See, for example:

- [this Stack Overflow question and answers thereto](https://stackoverflow.com/questions/34722352/how-can-i-secure-the-plugins-i-load);
- [this Medium post](https://medium.com/@amaraltohami30/ld-preload-privilege-escalation-linux-priv-escalation-c8abdf1a9bec);
- [this Reddit post](https://www.reddit.com/r/programming/comments/1c2h4rd/systemd_replacing_elf_dependencies_with_dlopen/) and [th Mastodon post to which it refers](https://mastodon.social/@pid_eins/112256363180973672) - it started with Lennart Poettering discussing having systemd load library dependencies at runtime with `dlopen()` rather than linking them in, with one rationale being that it can reduce the attack surface of systemd ("Recently, another reason for doing this came into focus: ibsystemd was used as a vehicle to get from the backdoored liblzma into the ssh process. There's hence value in reducing the ELF dependencies loaded into consumers of our library, if we can avoid it, to make it harder to use our code as exploit vehicle, even if we were neither the final target of the attack, nor directly attacked.";
- [another Reddit post asking about secure plugin loading](https://www.reddit.com/r/C_Programming/comments/1c77waq/can_a_safe_plugin_system_be_created_where_plugins/);
- a comment in [this Oracle document](https://docs.oracle.com/cd/E37838_01/html/E61050/gnclc.html) about secure use of `dlopen()`;

Those were found with [a Google search for "securely allowing dlopen for privileged code"](https://www.google.com/search?client=safari&rls=en&q=securely+allowing+dlopen+for+privileged+code&ie=UTF-8&oe=UTF-8); there may be others. A similar search, using LoadLibrary rather than dlopen, might also be useful.

Supporting plugin modules for libpcap would be useful in that it would, as you note, allow third parties to maintain their modules independently; the reason why I, at least, haven't done more than think about it is that it *really* needs to be secure.

(That also brings up the problem that *opening* a capture device may require privileges that you really don't want the program using. Perhaps the opening should be done by a program that runs with privileges and hands a file descriptor to libpcap over a UNIX-domain socket. The program still has some privileges, in that it has a descriptor for a capture device, but it's not as if it has *root* privileges if that's what opening the capture device required. But that's *another* discussion.)

fxlb:
Adding plugins into a security-sensitive library like libpcap need conducting an in-depth security risk assessment.

mcr:
So I think that we are all pretty wary about any kind of plugin system.  
I suggest that DPDK look at rpcapd.c, and consider bringing that into DPDK core, where the dependancy/API drift can be better dealt with within the DPDK CI.  The tcpdump/wireshark/libpcap interface is then rpcap protocol to localhost or Unix domain socket. 
Maybe that also facilitates better JIT/adaption of pcap filters into DPDK runtime, which might also involve pushing filters down to smart NICs.

rjarry:
Hey folks,

Thanks for the review and all the security refs, those are the right questions to ask.

On the rpcapd idea: the backends we're after ([OVS-DPDK](https://github.com/openvswitch/ovs/blob/main/lib/netdev-dpdk.c), [grout](https://github.com/DPDK/grout/pull/559), etc.) need to do packet capture into shared memory with no syscalls on the data path. The libpcap side would just mmap a file descriptor it gets from the dataplane over a unix socket. Going through rpcap's TCP/UDP sockets would negate the whole point -- you'd end up paying for copies and syscalls on every packet, which is exactly what these backends try to get rid of.

That said, the dlopen-in-a-privileged-process concern is legit and needs to be dealt with. I think there's a reasonable way to go about it though:

Before loading any plugin, libpcap could check if the process has elevated privileges (setuid/setgid, file capabilities on Linux). If so, either refuse to load plugins entirely, or drop the extra privileges first and only go ahead if the drop actually worked. That way a single `tcpdump` binary could still do `tcpdump -i eth0` (needs raw socket privileges) and `tcpdump -i $backend:port0` (just talks to a unix socket, no special privileges needed). These plugin backends won't need elevated privileges -- the dataplane process owns the capture, libpcap just reads from it.

In practice this probably means libpcap shouldn't load plugins silently on its own but instead expose something like `pcap_load_plugins()` that apps call explicitly. tcpdump would open its raw sockets, drop privileges, then call `pcap_load_plugins()`. Keeps the app in control instead of libpcap doing something unexpected behind its back.

Also `$PCAP_PLUGIN_DIR` should be ignored when running with elevated privileges, same way the dynamic linker drops `LD_LIBRARY_PATH` for setuid binaries. Only load from compiled-in system directories in that case, with ownership and permission checks on both the directory and the .so files.

Does that sound workable?

PS: if we want yet another layer on top, we could add optional signature verification before dlopen -- libpcap looks for a detached signature (e.g. `pcap-foo.so.sig`) next to each plugin and checks it against trusted public keys in a root-owned directory (say `/etc/libpcap/trusted-keys/`). Similar to how distros manage CA bundles or RPM signing keys -- the sysadmin decides what's trusted, not libpcap upstream. Though honestly, plugins installed via a package manager in root-owned system directories have already gone through the distro's own package signing, so it might be overkill. Worth discussing if the added complexity buys anything over the privilege checks and directory ownership validation alone.

Cheers

infrastation:
Thank you for attempting this development, it certainly has its pro et contra. The first committed version would need to be safe and maintainable, and it would take some more design and implementation work to get there. If there is a similar library that has already solved a very similar problem using plug-ins, it could be used as a model to relate the problem space and the solution space.

We are trying to make a few releases before long, so substantial changes and new features are currently mostly on hold. That said, why not consider this matter in the background meanwhile.

I agree that potentially there may be more sense in DPDK developing a stable tiny API just for packet capture (whether via libpcap or not), and hiding the internal churn behind that interface. That said, rpcap may be not the best interface/transport for that.

vjardin:
Thanks for all the feedback. I've scoped the plugin loader to mitigate the security concerns around loading external .so files

pcap-plugin.c
```diff
@@ -0,0 +1,571 @@
...
+	 * It covers setuid, setgid, file capabilities, and LSM
+	 * domain transitions (e.g. SELinux).
+	 */
+	return (getauxval(AT_SECURE) != 0);
+#else
+	/*
+	 * Fallback for other Unix systems: compare real and
+	 * effective uid/gid.
+	 */
+	return (getuid() != geteuid() || getgid() != getegid());
```
guyharris: This should use `issetugid()` if present on the OS.

fxlb:
> If there is a similar library that has already solved a very similar problem using plug-ins, it could be used as a model to relate the problem space and the solution space.

+1.

vjardin:
> > If there is a similar library that has already solved a very similar problem using plug-ins, it could be used as a model to relate the problem space and the solution space.
> 
> +1.

I did start checking.  If you have a taxonomy I could refer too, can you share those ? Any URLs on their git repo ?

vjardin:
Enclosed this minor update for the BSD systems.

mcr:
> So I think that we are all pretty wary about any kind of plugin system. I suggest that DPDK look at rpcapd.c, and consider bringing that into DPDK core, where the dependancy/API drift can be better dealt with within the DPDK CI. The tcpdump/wireshark/libpcap interface is then rpcap protocol to localhost or Unix domain socket. Maybe that also facilitates better JIT/adaption of pcap filters into DPDK runtime, which might also involve pushing filters down to smart NICs.

The more I think about it, the more this makes more sense.
Rather than throw a plugin at libpcap and hope the DPDK user/developer/operator has the most recent one, put this functionality with DPDK.  I suggest copying libpcap/rpcapd [yes, fork it] to DPDK, along with just enough of libpcap so that you can statically compile an "dpdk-pcapd" as part of DPDK.  That allows DPDK to evolve its capture mechanism(s), optimize your BPF (or another thing), deal with any multi-threading, locking or scaling issues, and also deal with any permissions issues that might arise.

vjardin:
How to get high performances with your proposal ?

vjardin:
Based on a taxonomy that @maxime-leroy shared with me in DM, I did simplify and limit the security surface of the plugin. See the latest update.

Moreover, I tried to enrich all the cases a DPDK plugin could need.

vjardin:
The ci seems to be flakky : choco times out or does not prepare properly the Windows environment ; s390x timeout on an unrelated topic.

vjardin:
@guyharris : please, can I get your review on it ?

guyharris:
> The ci seems to be flakky : choco times out or does not prepare properly the Windows environment

AppVeyor isn't the best. At one point I looked at another one - I think it was Cirrus CI - but they didn't have anything with Visual Studio pre-installed, and having to install it every time would impose a significant hit to CI time. Maybe it's changed, or maybe there's a way to get it done once and reuse it.

> s390x timeout on an unrelated topic.

For some reason it's currently prone to timeouts in the filter tests. The only thing noteworthy about it is that it's big-endian (which is a feature, not a bug, until the last big-endian machine running anything UN*Xy is shut down, and IBM doesn't seem to be in a hurry to get rid of either z/Architecture or Linux on Z); it's not as if those machines are exactly tiny machines with slow CPUs.

Maybe there's some issue with big-endian machines in the filter code.

guyharris:
> > The ci seems to be flakky : choco times out or does not prepare properly the Windows environment
> 
> AppVeyor isn't the best. At one point I looked at another one - I think it was Cirrus CI - but they didn't have anything with Visual Studio pre-installed, and having to install it every time would impose a significant hit to CI time. Maybe it's changed, or maybe there's a way to get it done once and reuse it.

I've hit "re-run incomplete" on the Appveyor build.

guyharris:
> For some reason it's currently prone to timeouts in the filter tests. The only thing noteworthy about it is that it's big-endian (which is a feature, not a bug, until the last big-endian machine running anything UN*Xy is shut down, and IBM doesn't seem to be in a hurry to get rid of either z/Architecture or Linux on Z); it's not as if those machines are exactly tiny machines with slow CPUs.
> 
> Maybe there's some issue with big-endian machines in the filter code.

Or maybe not, given that it's not reliably failing on any given test; it might just be random luck.

vjardin:
> Or maybe not, given that it's not reliably failing on any given test; it might just be random luck.

Can we increase the timeout ? 

How to get detailed logs/traces ?

infrastation:
Complete CI logs are available in respective CI checks associated with the pull request, if the obvious needs to be stated.

linux-s390x is hosted by OSU OSL, which are in the process of moving the data centre ([donations are welcome](https://www.beavsgive.org/organizations/open-source-lab) to cover the expenses), so some flakiness is expected. If this persists, it will be dealt with later as and if required.

One more time, releasing libpcap 1.11.0 is a more important and more urgent problem to solve compared to fixing of DPDK support, so these changes will have to wait until time is good for merging.

infrastation:
Apparently, the donations page is now [here](https://osuosl.org/donate/) (and requires manual input to donate specifically to OSL).

vjardin:
@infrastation : Denis, fyi, the following has to be selected:

image


When will the release 1.11.0 be tag'd ?

Meanwhile, can someone confirm the approach of this pull request ? We could use it to clean up the unsupported DPDK case of the libpcap.

infrastation:
A few urgent problems are in the way of 1.11.0 release, this is being worked on.

vjardin:
thanks @rjarry for the early wrap integrations on DPDK/grout.

vjardin:
Hi,

Sorry for not being "aware" of the release trains of libpcap, but what is the landing/when for the release for a possible merge of this pull request (assuming ok one) ? Behind the hood, it is to sync with grout and DPDK communities and to avoid the risks of "local" patches.

thank you,

infrastation:
Thank you for waiting Vincent. libpcap 1.11.0 has not been released yet because it got blocked by libpcap 1.10.7, which has not been released yet because it is still blocked by important bug fixes that took longer than expected. This is going to take a bit longer.

rjarry:
Hi folks,

just checking in to see if things are moving along.

@infrastation, do you have some visibility on the release of 1.11.0?

Cheers

infrastation:
Unfortunately, various things have been getting in the way of libpcap 1.11.0. I hope we will have enough attention to spend on this pull request in the coming months.

Let me remind that the main difficulty of this change comes from mixing two relatively independent pieces of work into one change: fixing/rewriting of DPDK support and introduction of plugins into libpcap. As discussed earlier, the latter part would require more work security- and maintenance-wise, so it may be better to attempt one improvement at a time.

fxlb:
> As discussed earlier, the latter part would require more work security- and maintenance-wise, so it may be better to attempt one improvement at a time.

Yes.

@rjarry "`Plugin loader <E2><80><94> always compiled, discovers pcap-*.so at runtime.`"
1) For security reasons, the loader must not be compiled by default.
2) Avoid non-ASCII characters in the code.

rjarry:
@fxlb: agreed, it should be disabled by default. NB: I didn't write the code :)

@infrastation: This work #1659 shouldn't even refer to "DPDK". It should just say "external backend plugin support". In fact, given the state of the existing DPDK backend, I have submitted another PR to remove it entirely from libpcap in #1697

Would you like to explore what needs to be done to limit the attack surface of such a loadable plugin system? How about dropping all privileges before loading plugins?

External plugins should *not* require any special capabilities. In fact, the example @vjardin submitted https://github.com/DPDK/grout/pull/565, the only thing the plugin needs is read/write access to a UNIX socket to start/stop the capture using the Grout API. Then, packets are written on an ephemeral shared memory segment.

PR #1664, opened 2026-03-30
Proactive Hardening: Add Rust Secure-by-Design Scaffolding for pcapng Parser State (No Public ABI Changes)

Noamismach:
## Summary
This PR introduces Secure-by-Design Scaffolding for the pcapng parser path by adding a Rust-backed opaque parser state and a narrow C FFI boundary.  
The intent is proactive modernization and hardening of a high-risk data parsing surface while preserving current behavior and compatibility.

## What This PR Changes
- Adds a Rust parser-state component (`pcapng_rs`) and integrates it into pcapng parsing internals.
- Replaces manual `realloc`-based growth for pcapng block buffers with Rust-managed dynamic storage.
- Replaces manual interface-array growth and multiplication arithmetic with safe `Vec` growth semantics.
- Exposes a strict C ABI boundary that is `catch_unwind`-protected, preventing Rust panics from crossing into C.
- Keeps all public libpcap APIs and externally visible ABI unchanged.

## Concrete Hardening Improvements Included
- Fixes a C-side memory leak in the precision error path by ensuring cleanup on the `goto fail` path.
- Adds safeguards around implicit numeric downcasts to reduce truncation and overflow-risk transitions at FFI and parser boundaries.

## Validation and Compatibility
- CMake integration is seamless and builds cleanly with the Rust static library integration.
- Public ABI remains unchanged; this is an internal implementation hardening and modernization.
- A 10-minute libFuzzer run against the parser path did not surface a shallow crash, which is consistent with the parser being mature, while still motivating proactive memory-safety modernization.

## Why This Direction
This approach is collaborative and incremental: it introduces a safe foundation for future parser evolution without disrupting users, while reducing maintenance burden from manual allocation and arithmetic-heavy state management.

PR #1665, opened 2026-03-30
Hardening: Fix memory leak in pcapng fail path and secure realloc logic

Noamismach:
This is a pure C99 follow-up to the discussions in PR #1664, isolating the parser hardening and memory leak fixes as recommended by @guyharris and @mcr.

### What this PR changes:
* **Fixes Memory Leaks in `pcap_ng_check_header`:** Reroutes early error returns (such as unknown timestamp precision or `p->buffer` allocation failures) through the `goto fail` path. This ensures that `ps->ifaces` and `p->buffer` are properly freed before returning `NULL`.
* **Hardens `realloc` state in `read_block`:** Ensures that when `p->buffer` is successfully expanded via `realloc` into a temporary pointer, the metadata (`p->bufsize`) is correctly updated to track the new capacity. This prevents stale size tracking and repeated, unnecessary reallocation attempts.

This patch strictly adheres to the project's C99 requirements and introduces no changes to the build systems.

### DIFF
--- a/sf-pcapng.c
+++ b/sf-pcapng.c
@@ -343,6 +343,7 @@ read_block(FILE *fp, pcap_t *p, struct block_cursor *cursor, char *errbuf)
 			return (-1);
 		}
 		p->buffer = bigger_buffer;
+		p->bufsize = bhdr.total_length;
 	}
 
 	/*
@@ -902,9 +903,7 @@ pcap_ng_check_header(const uint8_t *magic, FILE *fp, u_int precision,
 	default:
 		snprintf(errbuf, PCAP_ERRBUF_SIZE,
 		    "unknown time stamp resolution %u", precision);
-		free(p);
-		*err = 1;
-		return (NULL);
+		goto fail;
 	}
 
 	p->opt.tstamp_precision = precision;
@@ -931,9 +930,7 @@ pcap_ng_check_header(const uint8_t *magic, FILE *fp, u_int precision,
 	p->buffer = malloc(p->bufsize);
 	if (p->buffer == NULL) {
 		snprintf(errbuf, PCAP_ERRBUF_SIZE, "out of memory");
-		free(p);
-		*err = 1;
-		return (NULL);
+		goto fail;
 	}
 	ps->max_blocksize = INITIAL_MAX_BLOCKSIZE;
 


sf-pcapng.c
```diff
@@ -902,9 +903,7 @@ pcap_ng_check_header(const uint8_t *magic, FILE *fp, u_int precision,
 	default:
 		snprintf(errbuf, PCAP_ERRBUF_SIZE,
 		    "unknown time stamp resolution %u", precision);
-		free(p);
-		*err = 1;
-		return (NULL);
+		goto fail;
```
guyharris, line 906: Not sure there's a leak there, as nothing other than the `pcap_t`-plus-private-data was allocated, and it's freed, but this is good practice anyway in case the code changes - the resource deallocation is, at least, centralized.

(This is not one of C's strengths.)

sf-pcapng.c
```diff
@@ -931,9 +930,7 @@ pcap_ng_check_header(const uint8_t *magic, FILE *fp, u_int precision,
 	p->buffer = malloc(p->bufsize);
 	if (p->buffer == NULL) {
 		snprintf(errbuf, PCAP_ERRBUF_SIZE, "out of memory");
-		free(p);
-		*err = 1;
-		return (NULL);
+		goto fail;
```
guyharris, line 933: Ditto.

sf-pcapng.c
Noamismach: Agreed. Even if the secondary allocations havent occurred yet at this specific point, funneling everything through a centralized `goto fail` block removes a major footgun for future modifications. And yes, definitely not one of C's strengths!

infrastation:
Please make the commit message describe, using the conventional 72-column formatting, the problem(s) the commit fixes.

guyharris:
> Ensures that when `p->buffer` is successfully expanded via `realloc` into a temporary pointer, the metadata (`p->bufsize`) is correctly updated to track the new capacity. This prevents stale size tracking and repeated, unnecessary reallocation attempts.

Wireshark has a data structure called a `Buffer` that includes a buffer pointer and a size, and has routines to expand a buffer to a certain size, so reallocating the data and updating the size are centralized. We should probably have something similar to that (but simpler, with *only* the buffer pointer and size - libpcap doesn't need the other stuff).

I'll look at that after this is merged.

We should probably backport this.

Noamismach:
@infrastation - Done. I've amended the commit message to include a detailed explanation of the fixes, wrapped at 72 columns, and force-pushed the update.

@guyharris - That makes total sense. Having a unified struct for the buffer and its capacity would definitely eliminate this entire class of state-tracking bugs natively. Glad this patch helps secure things in the meantime, and honored to hear it might be backported!

PR #1667, opened 2026-04-15
pcap-linux: fix EFD_NONBLOCK rejected in old kernel

SmileyLemon:
pcap-linux: fix EFD_NONBLOCK rejected in old kernel
Some older kernels don't accept EFD_NONBLOCK in eventfd(), this fix would fall back to plain eventfd and set O_NONBLOCK via fcntl() when it happends.
If eventfd isn't supported, continue without it; breakloop won't interrupt blocking poll(), but capture still works.

### DIFF
--- a/pcap-linux.c
+++ b/pcap-linux.c
@@ -232,6 +232,7 @@ static int pcap_setfilter_linux(pcap_t *, struct bpf_program *);
 static int pcap_setdirection_linux(pcap_t *, pcap_direction_t);
 static int pcap_set_datalink_linux(pcap_t *, int);
 static void pcap_cleanup_linux(pcap_t *);
+static int pcap_linux_open_breakloop_fd(pcap_t *);
 
 union thdr {
 	struct tpacket2_hdr		*h2;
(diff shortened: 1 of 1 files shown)


infrastation:
Please describe the problem and the solution in more detail.

1. What are the steps to reproduce the problem?
2. Which old Linux kernels does this change address? libpcap currently requires at least 2.6.27, and [eventfd(2)](https://www.man7.org/linux/man-pages/man2/eventfd.2.html) says:
   ```
   EFD_NONBLOCK (since Linux 2.6.27)
                 Set the O_NONBLOCK file status flag on the open file
                 description (see open(2)) referred to by the new file
                 descriptor.  Using this flag saves extra calls to fcntl(2)
                 to achieve the same result.
   ```
3. Where does `ENOSYS` come from? Neither eventfd(2) nor fcntl(2) document this as a possible value.

infrastation:
If you do not describe the problem this change fixes, this pull request will be closed in a few weeks.

infrastation:
As discussed.

PR #1388, opened 2024-11-24
Indicate no-capture and no-inject devices.

infrastation:
This is a proof of the concept discussed in tcpdump-workers, it may require more work.

PR #1676, opened 2026-05-01
FreeBSD: fallback to the old mode of finding all devs

glebius:
in case if finddevs_bpf() failed.  A usual failure case is lack of privileges to open /dev/bpf.  However, traditionally `tcpdump -D` or wireshark were able to display possible list of capture devices even if started by an unprivileged user.

### DIFF
--- a/pcap-bpf.c
+++ b/pcap-bpf.c
@@ -43,7 +43,7 @@
 #endif
 #include <sys/utsname.h>
 
-#if defined(__FreeBSD__) && __FreeBSD_version < 1600006
+#if defined(__FreeBSD__)
 /*
  * Add support for creating FreeBSD usbusN interfaces as necessary.
  */
@@ -2936,7 +2936,12 @@ finddevs_bpf(pcap_if_list_t *devlistp, char *errbuf)
 	free(bi.bi_ubuf);
 	return (0);
 }
-#else
+#endif
+
+/*
+ * Guess possible USB tap points on FreeBSD < 1600006 or when finddevs_bpf()
+ * failed due to lack of privileges to open /dev/bpf.
+ */
 static int
 finddevs_usb(pcap_if_list_t *devlistp, char *errbuf)
 {
@@ -3011,7 +3016,6 @@ finddevs_usb(pcap_if_list_t *devlistp, char *errbuf)
 	closedir(usbdir);
 	return (0);
 }
-#endif
 #endif	/* FreeBSD */
 
 /*
@@ -3156,13 +3160,12 @@ pcapint_platform_finddevs(pcap_if_list_t *devlistp, char *errbuf)
 #endif
 
 #if defined(__FreeBSD__)
+	if (
 #if __FreeBSD_version >= 1600006
-	if (finddevs_bpf(devlistp, errbuf) == -1)
-		return (-1);
-#else
-	if (finddevs_usb(devlistp, errbuf) == -1)
-		return (-1);
+	(finddevs_bpf(devlistp, errbuf) == -1) &&
 #endif
+	(finddevs_usb(devlistp, errbuf) == -1))
+		return (-1);
 #endif
 
 	return (0);


fxlb (approved):
Seems good to me.

infrastation:
I currently do not have means to verify this change, excuse me.

glebius:
@guyharris maybe you can get this moving? Thanks!

guyharris (approved):
On FreeBSD 16, we probably want to use the old mode of finding all devices and then use the list of BPF-attached devices to mark devices *not* on the list as neither supporting capturing nor injecting, as per pull request #1388, so that an application will show the device (avoiding user questions of "why isn't XXX showing up?) but mark it as not supporting capturing or injecting packets (so the user knows that it won't work - hopefully we can indicate that in a fashion that lets the user knows to complain to whoever provides device support, not to tcpdump.org).

But we can do that later; FreeBSD 16 hasn't been released yet.

infrastation:
Let me get #1388 into shape then.

guyharris:
Works on FreeBSD 15.1 and 16.0-CURRENT, NetBSD 11.0, OpenBSD 7.9, DragonFly BSD 6.4.2, CupertinoBSD 26.4.1, and Solaris 11.4. Merge it?

infrastation:
This needs a comprehensible change log entry because it affects user-visible behaviour. Can be rebased in the same go.

guyharris:
> This needs a comprehensible change log entry because it affects user-visible behaviour.

What user-visible behavior does it change?

infrastation:
The output of `tcpdump -D`, as far as I managed to gather from the discussion.

guyharris:
> The output of `tcpdump -D`, as far as I managed to gather from the discussion.

About the only thing I see would change is that non-USB, non-network-interface devices that can be bound to and captured on via BPF would show up. I'm not sure that there are any such devices.

infrastation:
Alright, maybe the new comment could be a bit more specific, this way the logic would be a bit easier to follow when things change again in future or get ported to other systems that use FreeBSD network stack.

fxlb:
> The output of `tcpdump -D`, as far as I managed to gather from the discussion.

Yes, also with findalldevstest (test on cfarm440).
Without patch:
```
testprogs/findalldevstest
Error in pcap_findalldevs: Attempt to open /dev/bpf failed - root privileges may be required
```
With patch:
```
vtnet0
        Flags: UP, RUNNING (connected)
        Address Family: AF_LINK (18)
                Interface Index: 1
                Type: 6 (IFT_ETHER)
                Length: 6
                Address: 00:15:5d:xx:xx:xx
...
```

fxlb:
With `tcpdump -D` (same host).
Without patch:
```
$ ./tcpdump -D
tcpdump: Attempt to open /dev/bpf failed - root privileges may be required
```
With patch:
```
$ ./tcpdump -D
1.vtnet0 [Up, Running, Connected]
2.lo0 [Up, Running, Loopback]
3.dbus-system (D-Bus system bus) [none]
4.dbus-session (D-Bus session bus) [none]
```

fxlb:
> > This needs a comprehensible change log entry because it affects user-visible behaviour.
> 
> What user-visible behavior does it change?

See above..

guyharris:
> > > This needs a comprehensible change log entry because it affects user-visible behaviour.
> > 
> > 
> > What user-visible behavior does it change?
> 
> See above..

Those are on 16.0-CURRENT. That was with the current top-of-tree libpcap compiled on 16.0, so that it calls `finddevs_bpf()`, which fails if you don't have permission to open BPF devices, so the *entire call* fails, even though it found network interfaces. If you compile the version with this pull request, it only fails if `finddevs_bpf()` *and* `finddevs_usb()` fail, so if the former fails but the latter doesn't fail, it succeeds. `finddevs_usb()` doesn't require permission to open BPF devices, so it succeeds.

This would *not* happen if built on a pre-1.60 system, as it wouldn't call `finders_bpf()`, so this isn't a bug on any *released* version of FreeBSD, so it only represents a visible change there.

If we completely revert to the existing code, and remove `finddevs_bpf()` entirely, that should also fix the "permission denied" error. The only downside is that, if there are any non-USB, non-network-interface devices, they won't show up - but, if that was the case *prior* to FreeBSD 16, that's an *existiing* bug and should be fixed, if possible, in a way that *doesn't* involve using `finders_bpf()`, so that it fixes the problem on, say, FreeBSD 15, the current release.

I'll look into that. It *might* be that devices for pflog and ipfw taps are non-USB, non-network-interface devices. If they exist as BPF-tappable devices in pre-FreeBSD-16 versions (or on non-FreeBSD OSes), we should find a way of enumerating them without using the new BPF ioctl, to fix that problem on those OSes.

fxlb:
> Those are on 16.0-CURRENT.

Yes, linked to ea5317adaf0e4e308543fbd66977f89344f9ef7f.
(see my 2026/03/26 email.)

glebius:
> I'll look into that. It _might_ be that devices for pflog and ipfw taps are non-USB, non-network-interface devices. If they exist as BPF-tappable devices in pre-FreeBSD-16 versions (or on non-FreeBSD OSes), we should find a way of enumerating them without using the new BPF ioctl, to fix that problem on those OSes.

Before FreeBSD 16 all bpf taps were associated with interfaces. Even USB in the kernel were interfaces, but APIs were hiding them. The ipfw and pf taps were interfaces.

Same statement other way around: a non-USB, non-network interface taps do not exist on versions where `finddevs_bpf()` exists.

guyharris:
> Before FreeBSD 16 all bpf taps were associated with interfaces. Even USB in the kernel were interfaces, but APIs were hiding them. The ipfw and pf taps were interfaces.
> 
> Same statement other way around: a non-USB, non-network interface taps do not exist on versions where finddevs_bpf() exists.

So the issue is that FreeBSD 16 doesn't tie BPF capture devices to network interfaces.

As FreeBSD 16 hasn't been released yet, how about providing a mechanism that *doesn't* require special privileges and that enumerates devices with BPF taps?

infrastation:
I was going to suggest a read-only file somewhere under `/proc/`, but apparently the latter is not mounted by default.

glebius:
> So the issue is that FreeBSD 16 doesn't tie BPF capture devices to network interfaces.

This is not an issue, it is a feature :) Now you can have many different tapping points and interfaces are just a subset of them.

> As FreeBSD 16 hasn't been released yet, how about providing a mechanism that _doesn't_ require special privileges and that enumerates devices with BPF taps?

To be fair, I see very little value in that. What's the point in seeing available bpf taps, when you can't neither read them nor write to them?

Theoretically this is possible, of course. That would require relaxing permissions on /dev/bpf and then doing privilege checks at per-ioctl level. This would be a change that would require more thorough reviewing and of course will widen the attack surface (amount of kernel code directly available to unpriveleged syscall).

IMHO, falling back to the old way of deducting possible list of taps through listing interfaces is good enough for the strange case where "I want to see them albeit I won't be able to use them".

glebius:
> I was going to suggest a read-only file somewhere under `/proc/`, but apparently the latter is not mounted by default.

Roughly speaking FreeBSD doesn't have procfs. We can export list of taps as a sysctl, though!

glebius:
Also, I have a feature in mind, that is not implemented yet - a dynamic templated taps. E.g. the tap name would be "ipfw%u" meaning that you can attach to any rule of ipfw. Don't know how this will fit into listing.

guyharris:
> What's the point in seeing available bpf taps, when you can't neither read them nor write to them?

At least with network interfaces, the point is "to get users to ask 'why don't I have permission to capture on these devices?', for which the correct answer is fairly simple to figure out, as opposed to 'why am I not seeing the interfaces on my machine?', which requires a bunch of extra work to figure out if the problem is a permissions problem or something else.

(Yes, that's a thing.)

It *might* be the case that, for things other than network interfaces, that won't turn out to be an issue. For network interfaces, we can just enumerate all of them and flag the ones for which the user doesn't have capture permission as such.

infrastation:
`sysctl` allows to enumerate various physical and logical devices without elevating privileges, so the pattern has already been established and BPF could just reuse it.

glebius:
> It _might_ be the case that, for things other than network interfaces, that won't turn out to be an issue. For network interfaces, we can just enumerate all of them and flag the ones for which the user doesn't have capture permission as such.

Right. So what the existing patch does, it basically reverts to the old mode when encounters EPERM and it will list **all of the interfaces and USB devices** that were listed before. Nothing that was present before won't be missed! That's why I'm just suggesting to include the patch as is and then we can discuss if we want or not to disclose list of all taps including new and future features to unprivileged users.

glebius:
> `sysctl` allows to enumerate various physical and logical devices without elevating privileges, so the pattern has already been established and BPF could just reuse it.

The sysctl needs to be discussed with other FreeBSD network folks. For example right now it would give some insight to an unprivileged user into ipfw rules. Unprivileged users can not see list of rules, can not even estimate their count. Suggested sysctl will create a little bit of information leak on rules. If we add more new types of taps that would also might be sensitive information.

That's why I'm again asking to include this patch as is to get back to the status quo, where only interfaces and USB devices are seen.

PR #1684, opened 2026-05-19
fuzz: fix unused parameter compiler warning in fuzz_filter.c

JoaoNasMonteiro:
When building with -Wall -Wextra a warning is triggered in the fuzzing test suite:

```
libpcap/testprogs/fuzz/fuzz_filter.c: In function ‘fuzz_openFile’:
libpcap/testprogs/fuzz/fuzz_filter.c:7:33: warning: unused parameter ‘name’ [-Wunused-parameter]
    7 | void fuzz_openFile(const char * name){
      |                    ~~~~~~~~~~~~~^~~~
```

Since the `name` parameter in `fuzz_openFile()` is intentionally unused this PR casts the unused parameter to `void` to explicitly signal to the compiler that this is intentional.

Since this is my first PR please let me know if there is something that i've got wrong!

### DIFF
--- a/testprogs/fuzz/fuzz_filter.c
+++ b/testprogs/fuzz/fuzz_filter.c
@@ -5,7 +5,7 @@
 #include <pcap/pcap.h>
 
 void fuzz_openFile(const char * name){
-    //do nothing
+    (void)name;//do nothing
 }
 
 int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
--- a/testprogs/fuzz/fuzz_filter.c
+++ b/testprogs/fuzz/fuzz_filter.c
@@ -1,11 +1,12 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include "varattrs.h"
 
 #include <pcap/pcap.h>
 
-void fuzz_openFile(const char * name){
-    (void)name;//do nothing
+void fuzz_openFile(const char * name _U_){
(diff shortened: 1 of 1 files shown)


fxlb:
Thanks.

Or with `_U_` [__attribute__((unused))], we use in some places:

```patch
diff --git a/testprogs/fuzz/fuzz_filter.c b/testprogs/fuzz/fuzz_filter.c
index d234a022..0067f324 100644
--- a/testprogs/fuzz/fuzz_filter.c
+++ b/testprogs/fuzz/fuzz_filter.c
@@ -1,10 +1,11 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include "varattrs.h"
 
 #include <pcap/pcap.h>
 
-void fuzz_openFile(const char * name){
+void fuzz_openFile(const char * name _U_){
     //do nothing
 }
 

```

JoaoNasMonteiro:
Interesting, I was not familiar with that header. 

Would you recommend that I change my fix to use the `_U_` macro instead?

fxlb:
> Would you recommend that I change my fix to use the `_U_` macro instead?

Yes, that's what we do elsewhere.

JoaoNasMonteiro:
Thanks for the review! 

I've updated fuzz_filter.c to use the _U_ macro as you suggested, let me know if everything looks good now!

fxlb:
Merged, thanks.

PR #1686, opened 2026-05-28
rpcapd: free BPF instruction buffer in daemon_unpackapplyfilter()

TristanInSec:
daemon_unpackapplyfilter() allocates a bpf_insn array to receive filter
instructions from the client.  pcap_setfilter() copies the filter via
install_bpf_program() and does not take ownership of the caller's buffer.

The buffer was never freed on any return path, leaking up to 64 KB per
RPCAP_MSG_UPDATEFILTER_REQ message.

Fix: free bf_prog.bf_insns before every return after the allocation.


infrastation:
This is an equivalent of #1640, but the proposed changes are not as easy to maintain. Closing.

infrastation:

PR #1687, opened 2026-05-28
bpf: reject constant shift amounts >= 32 in validator and interpreter

TristanInSec:
The BPF validator accepted LSH/RSH instructions with a constant operand
(BPF_K) of 32 or greater.  Shifting a uint32_t by >= 32 is undefined
behavior (C11 6.5.7p3).  The register-based variant (BPF_X) already
checked X < 32, but the constant variant did not.

This caused platform-dependent filter results: on x86-64 the shift
masked to 5 bits (no-op for k=32), while on ARM32 it zeroed the
accumulator.

Fix: reject k >= 32 in the validator, and add bounds checking in the
interpreter to match the existing BPF_X behavior.

### DIFF
--- a/bpf_filter.c
+++ b/bpf_filter.c
@@ -349,11 +349,17 @@ DIAG_ON_DEFAULT_ONLY_SWITCH
 			continue;
 
 		case BPF_ALU|BPF_LSH|BPF_K:
-			A <<= pc->k;
+			if (pc->k < 32)
+				A <<= pc->k;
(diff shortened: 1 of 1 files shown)


infrastation:
This is a known issue, and it already has a pending solution that includes an equivalent of the proposed changes, as well as tests to prove it works as expected.  The pending solution is going to be committed later. Therefore closing.

Issue #1670, opened 2026-04-20
vxlan: "vxlan" filter do not support "vxlan and arp"


satanaileo:
## Problem Description

libpcap does not properly support filtering ARP packets within VXLAN encapsulation using the expression `vxlan and arp`. When using this filter combination, the generated BPF code fails to correctly match ARP packets inside VXLAN tunnels.

## Steps to Reproduce

1. Run tcpdump with VXLAN and ARP filter:
   ```bash
   ./tcpdump -i net_api vxlan and arp
   ```

2. Observe the warning (in older versions):
   ```
   Warning: Kernel filter failed: Invalid argument
   ```

3. Check the generated BPF instructions:
   ```bash
   ./tcpdump -i net_api vxlan and arp -d
   ```

## Current Behavior

The generated BPF code has incorrect offset calculations for accessing the encapsulated Ethernet frame's EtherType field. The current code in `gen_ether_linktype()` directly compares `OR_LINKTYPE` without validating that there is a valid encapsulated link-layer header present.

### Generated BPF (before fix):
```
(000) ldh      [12]
(001) jeq      #0x800           jt 2        jf 13
(002) ldb      [23]
(003) jeq      #0x11            jt 4        jf 24
(004) ldh      [20]
(005) jset     #0x1fff          jt 24       jf 6
(006) ldxb     4*([14]&0xf)
(007) ldh      [x + 16]
(008) jeq      #0x12b5          jt 9        jf 24
(009) ldb      [x + 22]
(010) jeq      #0x8             jt 11       jf 24
(011) txa
(012) jeq      x                jt 20       jf 24
...
(020) ldx      M[1]
(021) ldh      [x + 0]
(022) jeq      #0x806           jt 23       jf 24
```

Notice that instruction (021) attempts to read from `[x + 0]` which may not correctly point to the encapsulated frame's EtherType field.

## Expected Behavior

The BPF filter should correctly:
1. Identify VXLAN packets (UDP port 4789 or other VXLAN ports)
2. Verify there is a valid encapsulated Ethernet frame
3. Check if the encapsulated EtherType is ARP (0x0806)

### Generated BPF (after fix):
```
(000) ldh      [12]
(001) jeq      #0x800           jt 2        jf 17
(002) ldb      [23]
(003) jeq      #0x11            jt 4        jf 46
(004) ldh      [20]
(005) jset     #0x1fff          jt 46       jf 6
(006) ldxb     4*([14]&0xf)
[...]

PR #1688, opened 2026-05-30
fix BPF optimizer JT==JF bypass dropping memory stores needed by targ…

satanaileo:
[https://github.com/the-tcpdump-group/libpcap/issues/1670](url) 
Fix BPF optimizer JT==JF bypass dropping memory stores needed by target block (issue #1670)                                                                   
                                                                                                                                                          
  Found via "vxlan and arp" filter producing incorrect results: the optimizer's                                                                           
  JT==JF bypass path only checked use_conflict(target->out_use) but missed
  target->in_use, causing blocks storing M[0]/M[1]/M[2] offsets to be                                                                                     
  incorrectly eliminated when the target block itself needs those values.

### DIFF
--- a/optimize.c
+++ b/optimize.c
@@ -1696,7 +1696,8 @@ opt_j(opt_state_t *opt_state, struct edge *ep)
 		 * from what the blocks before it did and isn't
 		 * doing any tests the results of which matter.
 		 */
-		if (!use_conflict(ep->pred, JT(ep->succ))) {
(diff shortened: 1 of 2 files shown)


infrastation:
Thank you for preparing this change. I would need to study the optimizer more to verify it fixes a bug and does not introduce new bugs. In any case, it will need a good commit message and new tests because right now obviously it does not change anything that the current tests cover. Also it would be useful to understand what other syntax is subject to the same bug.

Issue #1441, opened 2025-01-16
Files with BSD advertising clause for Lawrence Berkeley Laboratory


johnthacker:
Some of the include files, notably pcap.h, have a modified version of the original 4-clause BSD license where the advertising clause refers to the Computer Systems Engineering Group at Lawrence Berkeley Laboratory instead of the University of California, Berkeley and its contributors.

Lawrence Berkeley Laboratory is operated by the University of California, Berkeley, but for the US Department of Energy. The official UC Berkeley statements referring to the deletion of the clause in 1999 and restated in 2012 specifically detail the original version of the clause.

  ftp://ftp.cs.berkeley.edu/pub/4bsd/README.Impt.License.Change
  https://ipira.berkeley.edu/sites/default/files/amendment_of_4-clause_bsd_software_license.pdf

Does this mean that the advertising provision is in force when using those files? IANAL, but I would guess so unless a specific statement has been made to the contrary. Is it possible to remove the advertising clause from those files, so that it is compatible with other licenses? Could that be done possibly by contacting either the Berkeley Lab, or Steve McCanne, Craig Leres, and Van Jacobson?

The public [Software Disclosure and Distribution policy](https://commons.lbl.gov/display/rpm2/Software+Disclosure+and+Distribution) for the Berkeley Lab does not appear to address the issue, though it does provide some contact information.

infrastation:
AFAIR, Guy quite a while ago was going to clarify this with the intent to switch to a more common BSD licence variant, but I do not know how it went.

guyharris:
> AFAIR, Guy quite a while ago was going to clarify this with the intent to switch to a more common BSD licence variant, but I do not know how it went.

I sent mail to some people at LBL about this, but nobody said "yes, we'll change the license", and didn't make it sufficiently clear to me what doing so would involve.

My original email was (modulo some reformatting to survive being put into Github-flavored Markdown):

> I'm not sure who are the right people to contact about this; if none of you are, but you know who are the right people, perhaps you can forward this question to them:
> 
> Various releases of the Berkeley Software Distribution had, in the source code files, a copyright notice that read
> 
> ```
> 	Copyright (c) <years>, The Regents of the University of California.
> 	All rights reserved.
> 
> 	Redistribution and use in source and binary forms, with or without
> 	modification, are permitted provided that the following conditions
> 	are met:
> 	1. Redistributions of source code must retain the above copyright
> 	   notice, this list of conditions and the following disclaimer.
[...]

PR #1689, opened 2026-06-03
Add SPDX license identifier tags to files

illuusio:
Add SPDX license identifier tags to files. Each SPDX license identifier is in own commit to make easier to review this.

### DIFF
--- a/bpf_filter.c
+++ b/bpf_filter.c
@@ -1,4 +1,6 @@
 /*-
+ * SPDX-License-Identifier: BSD-4-Clause-UC
+ *
  * Copyright (c) 1990, 1991, 1992, 1993, 1994, 1995, 1996, 1997
  *	The Regents of the University of California.  All rights reserved.
  *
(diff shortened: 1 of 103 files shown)


infrastation:
Thank you for preparing these changes. The licence boilerplates went out of sync over many years and require a bit more work than just adding the identifiers to become accurate, see #1441 for the context. It will be a better time to look into this after the 1.11.0 release, I guess.

illuusio:
Actually I read that one before submitting. Would be easier to just reduce to easy one's like `BSD-[23]-Clause` for the start?

infrastation:
To me it would make the most sense to complete the tcpslice part of the problem first in order to work a reference method out, and then to apply the method to libpcap and tcpdump — this way any remaining/introduced discrepancies will be easier to trace in future if required. I just got distracted by more urgent matters in the middle of it, as often happens. Let's keep this pull request open for a while.

For any subsequent revisions please add `[skip ci]` to the commit summary of the last commit if all changes concern comments only.

PR #1691, opened 2026-06-20
rpcapd: fix a memory leak in daemon_unpackapplyfilter()

matejsmycka:
The BPF filter program allocated in daemon_unpackapplyfilter() was never freed, leaking memory on every RPCAP_MSG_UPDATEFILTER request as well as on each of the function's error paths.


infrastation:
The changes are exactly as in #1686, but the commit message is empty, closing.

infrastation:

PR #1692, opened 2026-06-20
fuzz: make fuzz_rclient reach the rpcap packet-read path

matejsmycka:
The fuzz_rclient target never reached the rpcap packet-reading code because the fuzzing sock_open() stub returned INVALID_SOCKET and getpeername()/getnameinfo() failed on the unconnected fuzzing socket; this gives the stub a usable socket, skips those connection-dependent calls under the fuzzing macro, and has the harness read the captured bytes so sanitizers can catch over-reads.

### DIFF
--- a/testprogs/fuzz/fuzz_rclient.c
+++ b/testprogs/fuzz/fuzz_rclient.c
@@ -43,7 +43,11 @@ int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
     //loop over packets
     r = pcap_next_ex(pkts, &header, &pkt);
     while (r > 0) {
+        unsigned int i, sum = 0;
         fprintf(outfile, "packet length=%d/%d\n",header->caplen, header->len);
+        for (i = 0; i < header->caplen; i++)
+            sum += pkt[i];
+        fprintf(outfile, "checksum=%u\n", sum);
         r = pcap_next_ex(pkts, &header, &pkt);
     }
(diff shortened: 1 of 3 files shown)


matejsmycka:
I think the failure is a CI/CD issue, not a PR issue.

guyharris:
> I think the failure is a CI/CD issue, not a PR issue.

If by "the failure" you mean the Appveyor failure, yeah, it's a CI/CD issue, the issue being "Appveyor has, for some unknown reason, immense difficulty opening some TLS connections in order to download tools or libraries".

I've restarted the build, just for the lulz, but, unless the fuzzing code can be built and run on Windows, this change is unlikely to make anything worse on Windows.

testprogs/fuzz/fuzz_rclient.c
```diff
@@ -43,7 +43,11 @@ int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
     //loop over packets
     r = pcap_next_ex(pkts, &header, &pkt);
     while (r > 0) {
+        unsigned int i, sum = 0;
         fprintf(outfile, "packet length=%d/%d\n",header->caplen, header->len);
+        for (i = 0; i < header->caplen; i++)
```
guyharris, line 48: This doesn't look as if it's related to this issue. Is that debugging code that leaked out into the pull request, or is it part of *another* change you want to submit?

PR #1690, opened 2026-06-20
rpcap: fix an out-of-bounds read in pcap_read_nocb_remote()

matejsmycka:
The captured length in an rpcap packet message was validated against the full payload length instead of the payload length minus the rpcap packet header, letting a malicious or compromised server make a consumer read past the end of the capture buffer.

PR #1693, opened 2026-06-25
rpcap: validate caplen against payload size in pcap_read_nocb_remote

Samin061:
pcap_read_nocb_remote checks a packet message's caplen against the full payload length plen, but plen also covers the 20-byte struct rpcap_pkthdr that precedes the captured data, so a malicious server can report a caplen up to sizeof(struct rpcap_pkthdr) bytes larger than the data actually received. The packet-data pointer is buffer + sizeof(rpcap_header) + sizeof(rpcap_pkthdr) and plen may be as large as bufsize minus the rpcap header, so the caplen-byte packet handed to the caller reads past the end of p->buffer. Reject payloads shorter than a packet header and bound caplen by plen minus sizeof(struct rpcap_pkthdr).

### DIFF
--- a/pcap-rpcap.c
+++ b/pcap-rpcap.c
@@ -576,7 +576,22 @@ static int pcap_read_nocb_remote(pcap_t *p, struct pcap_pkthdr *pkt_header, u_ch
 		return 0;	/* Return 'no packets received' */
 	}
 
-	if (ntohl(net_pkt_header->caplen) > plen)
+	/*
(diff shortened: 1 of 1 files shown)


infrastation:
See #1690.

Samin061:
Missed that one when I went looking, sorry for the dupe. #1690 predates this and lands the same bound on caplen, so closing in favor of it.

infrastation:

PR #1694, opened 2026-06-27
rpcap: bound userinfo parsing in rpcap_doauth_userinfo

Samin061:
rpcap_doauth_userinfo() copies the URI userinfo into 256-byte username and password stack buffers one character at a time with no limit on the index, so a userinfo with more than 256 decoded bytes writes past the end of those buffers. The FIXME on the loop and the RPCAP_USE_USERINFO=0 guard in pcap-int.h already flag this as a stack overflow to be reimplemented or removed. This bounds the write index by sizeof(username) so the loop stops before overflowing and rejects an over-long username or password. Confirmed with ASAN: a 600-byte userinfo triggers a stack-buffer-overflow on the unpatched loop and is cleanly rejected after the change, while normal user:pass and full 255-byte fields still parse.

### DIFF
--- a/pcap-rpcap.c
+++ b/pcap-rpcap.c
@@ -2266,8 +2266,16 @@ static int rpcap_doauth_userinfo(PCAP_SOCKET sockctrl, SSL *ssl, uint8_t *ver,
 
 	if ((ptr = userinfo) != NULL)
 	{
-		// FIXME: 'pos' is not limited to the sizes of 'username' and 'password'.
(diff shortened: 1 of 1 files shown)


infrastation:
Thank you for preparing this change, but this problem requires a different, more substantial fix.  Commit e3e47ff explains it in more detail.

Samin061:
Makes sense. I was only bounding the write, but rejecting malformed credentials before the connection is set up is the right fix, and that means untangling the URL parsing from the protocol code rather than patching the loop in place. Thanks for the pointer to e3e47ff, I'll leave this closed.

PR #1695, opened 2026-06-29
rpcapd: free the BPF program in daemon_unpackapplyfilter

Samin061:
daemon_unpackapplyfilter() mallocs the bf_insns array and hands it to pcap_setfilter(), which installs its own malloc()+memcpy() copy and does not take ownership of the caller's program, so the allocation is never released. An authenticated client leaks bf_len * sizeof(struct bpf_insn) bytes on the server for each RPCAP_MSG_UPDATEFILTER_REQ and for the filter carried in RPCAP_MSG_STARTCAP_REQ, on the success path and on every error return after the malloc. Free bf_prog.bf_insns before each of those returns.


infrastation:
See #1691.

Samin061:
Right, this is the same leak already covered by #1691 and the earlier chain. Didn't catch that it was tracked before opening this. Closing as a duplicate.

infrastation:

PR #1697, opened 2026-07-10
dpdk: remove capture backend

rjarry:
The current DPDK integration model is not useful in practice. Since libpcap/tcpdump takes exclusive ownership of DPDK ports via rte_eth_dev_configure(), it cannot capture traffic from an existing DPDK application. This makes it useless for the primary use case of observing packets in a running DPDK-based network stack.

The support added in commit 8d7f0e39da08 ("enable dpdk for libpcap, test with DPDK v18.11") was never updated to follow the evolution of the public DPDK API. This led the backend to be disabled by default in commit 06fbfb73255d ("DPDK: don't enable it by default.") because the maintenance cost became too high.

Remove all DPDK capture support from the source, build systems (autoconf, CMake) and documentation.

This is preliminary cleanup before introducing external loadable plugin support. Userspace networking capture backends like DPDK do not belong in libpcap itself and are better handled as external plugins.

Link: https://github.com/the-tcpdump-group/libpcap/pull/1659


infrastation:
No objections, except this part of the commit message is misleading:
```
This is preliminary cleanup before introducing external loadable plugin
support. Userspace networking capture backends like DPDK do not belong
in libpcap itself and are better handled as external plugins.
```
It gives an impression there is a consensus on whether libpcap should use plug-ins in future, and how exactly it should do it, neither of which is true at the time of this writing. Please make the comment say everything that needs to be said about this particular change, and nothing else.

rjarry:
Done :+1:

infrastation:
Thank you, this is going to be merged tomorrow if nobody objects. Meanwhile you can add an entry to the change log as well if you wish.

rjarry:
Thanks, I can update the change log. Should I target 1.11.0?

infrastation:
Yes.

rjarry:
Done, let me know what you think.

infrastation:
Thank you.

PR #1700, opened 2026-07-13
translatetest: expand the explanation of the "if" in the option processing

guyharris:
Explain in more detail why we're using an "if", and note that if we add any command-line options that *don't* cause us to do something quick and exit, we should turn the "if" back into a "while".

### DIFF
--- a/testprogs/translatetest.c
+++ b/testprogs/translatetest.c
@@ -261,6 +261,30 @@ main(int argc, char **argv)
 	{
 		int op;
 		opterr = 0;
+
+		/*
+		 * Currently, the only option we support is "-h",
+		 * which causes us to print a help message and exit
+		 * with a success error code.
+		 *
+		 * All other options are invalid, which causes us
+		 * to print a usage message and exit with an "invalid
+		 * usage" error code.
+		 *
+		 * This means that no code after the switch statement
+		 * is ever executed, which generates a warning from Sun C
+		 * if the outer control statement is "while", because
+		 * there would be "continue the loop" machine code after
+		 * the switch statement.
+		 *
+		 * Therefore, we use an "if" statement, to suppress that
+		 * warning.
+		 *
+		 * If we ever add options that do *not* cause us to
+		 * exit, the "if" should become a "while", as the warning
+		 * should no longer occur, and we may want to support
+		 * processing more than one option.
+		 */
 		if ((op = getopt(argc, argv, "h")) != -1) {
 			switch (op) {
 			case 'h':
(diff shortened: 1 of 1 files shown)


testprogs/translatetest.c
```diff
@@ -261,6 +261,30 @@ main(int argc, char **argv)
...
+		/*
+		 * Currently, the only option we support is "-h",
+		 * which causes us to print a help message and exit
+		 * with a success error code.
+		 *
+		 * All other options are invalid, which causes us
+		 * to print a usage message and exit with an "invalid
+		 * usage" error code.
+		 *
+		 * This means that no code after thw switch statement
```
infrastation: "thw"

testprogs/translatetest.c
```diff
@@ -261,6 +261,30 @@ main(int argc, char **argv)
...
+		 * is ever executed, which generates a warning from Sun C
+		 * if the outer control statement is "while", because
+		 * there would be "continue the loop" machine code after
+		 * the switch statement.
+		 *
+		 * Therefore, we use an "if" statment, to suppress that
```
infrastation: "statment"

testprogs/translatetest.c
```diff
@@ -261,6 +261,30 @@ main(int argc, char **argv)
...
+		 * warning.
+		 *
+		 * If we ever add options that do *not* cause us to
+		 * exit, the "if" should becone a "while", as the warning
```
infrastation: "becone"

testprogs/translatetest.c
```diff
@@ -270,11 +294,7 @@ main(int argc, char **argv)
 				usage_short(stderr);
 				exit(EX_USAGE);
 			}
-			/*
-			 * The end of this block is not reachable, which
-			 * generates a warning from Sun C if the outer control
-			 * statement is "while", hence use "if".
-			 */
+			/* NOTREACHED */
```
infrastation: `/* NOTREACHED */` usually works as a hint for the compiler, which Sun C documentation confirms, but specifically in this case it did not suppress the warning so long as the outer statement remained "while", as far as I remember.

testprogs/translatetest.c
guyharris: Fixed.

testprogs/translatetest.c
guyharris: Fixed.

testprogs/translatetest.c
guyharris: Fixed.

testprogs/translatetest.c
guyharris: I put back the "hint for humans" comment, and had it just refer to the warning mentioned above.

(`/* NOTREACHED */` could also serve as a hint for humans, but if it doesn't serve as a hint for the compiler, we might as well just write it for humans.)

PR #1705, opened 2026-07-20
Rename SWAPSHORT(), SWAPLONG(), SWAPLL() to PCAP_BSWAP_({16,32,64})

fxlb:
In particular, 'LONG' was ambiguous because it is 4 bytes in a 32-bit build and 8 bytes in a 64-bit build.

Don't use BSWAP_({16,32,64}) to avoid redefine on Solaris/illumos.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -21,6 +21,7 @@ DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
       Deprecate pcap_compile_nopcap().
       Remove unmaintained DPDK capture backend (was disabled by default).
       Return more useful strings from pcap_statustostr().
+      Rename SWAPSHORT(), SWAPLONG(), SWAPLL() to PCAP_BSWAP_({16,32,64}).
     Link-layer types:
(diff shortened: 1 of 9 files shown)


guyharris:
Should it be `PCAP_` or `PCAPINT_`? If we export it to code using libpcap, `PCAP_` is appropriate, but if it's only for libpcap modules (for which we *currently* have no API for external modules), maybe it should be called `PCAPINT_`, as we do for routines not exported to third-party code.

fxlb:
> Should it be `PCAP_` or `PCAPINT_`?

Good question!  `PCAPINT_` is indeed more appropriate here. I’ll update it.

infrastation:
Please go ahead and do not forget about `EXTRACT_SHORT()` and `EXTRACT_LONG()`. Also for the unconditional byte-swapping macros it would be useful to have tests, which could be added later or in the same commit as the renaming.

infrastation:
I could add the tests later if it helps.

fxlb:
> I could add the tests later if it helps.

Yes, it helps, thanks.

PR #1706, opened 2026-07-20
Preserve statements when collapsing root branch chains

### LINKED ISSUE #1699: BPF optimizer miscompilation: equal constant sub-expressions collapse a comparison into a constant
## Summary

`pcap_compile()` with `optimize=1` miscompiles filter expressions in which the
two operands of a relational comparison are independent sub-expressions that the
optimizer proves have the same value. The optimizer aliases the two operands and
rewrites the comparison as a register-against-itself test (e.g. `A > A`, i.e.
`0 > 0`), which folds to a constant. The resulting program classifies **every**
packet incorrectly. Compiling the identical filter with `optimize=0` produces a
correct program, so the two optimization levels disagree.

This is a correctness defect. It is reachable from any caller that lets users supply a filter
string, e.g. `tcpdump <expr>`.

## Affected version

- libpcap `1.11.0-PRE-GIT`, commit `c28d595f4a9e932df08a5063e8c62378035a6fe9`.
- Component: BPF optimizer, `optimize.c` (value numbering).

## Minimal reproducer

Filter: `1/2=1/3&&1>4/5&&1>6/7`

Every sub-expression is a constant: `1/2 == 0`, `1/3 == 0` (so `1/2 = 1/3` is
true), `4/5 == 0` and `6/7 == 0` (so `1 > 4/5` and `1 > 6/7` are true). The
filter is therefore **true for every packet** and must match everything.

```c
/* cc -O2 repro.c -lpcap -o repro */
#include <stdio.h>
#include <pcap.h>
int main(void) {
    const char *filter = "1/2=1/3&&1>4/5&&1>6/7";
    unsigned char pkt[64] = {0};           /* contents irrelevant */
    struct pcap_pkthdr h = {0};
    h.caplen = h.len = sizeof pkt;
    for (int opt = 0; opt <= 1; opt++) {
        pcap_t *pd = pcap_open_dead(DLT_EN10MB, 65535);
        struct bpf_program bp;
        if (pcap_compile(pd, &bp, filter, opt, PCAP_NETMASK_UNKNOWN) != 0) {
            printf("optimize=%d: compile error: %s\n", opt, pcap_geterr(pd));
        } else {
            printf("optimize=%d: match = %u\n", opt,
                   pcap_offline_filter(&bp, &h, pkt));
            pcap_freecode(&bp);
        }
        pcap_close(pd);
    }
    return 0;
}
```

Output:

```
optimize=0: match = 65535     <- correct (filter is a tautology)
optimize=1: match = 0         <- WRONG (matches nothing)
```

## What the optimizer does

`optimize=0` (correct) computes the three comparisons and accepts:

```
(000) ld       #0x1
(001) st       M[0]
...                       ; computes 1/2 into M[1], 1/3 into M[3]
(018) jeq      x          jt 19 jf 46     ; 1/2 == 1/3  -> true
...                                        ; 1 > 4/5, 1 > 6/7 -> true
(045) ret      #65535     ; accept
(046) ret      #0
```

`optimize=1` (wrong) collapses the whole thing to a single self-comparison:

```
(000) ldx      #0x0
(001) txa                 ; A = 0
(002) jgt      x          jt 3 jf 4    ; 0 > 0  -> false
(003) ret      #65535
(004) ret      #0         ; reject everything
```

The optimizer constant-folded the operands to the same value, assigned them the
same value number, and then treated the comparison as a comparison of a value
with itself — emitting `0 > 0` and folding it to "reject all".

## Root cause

Value numbering in `optimize.c` (`vmap` / the constant-folding in `fold_op()`
and the block-condition handling) assigns equal value numbers to the two
independently-computed operands, then the comparison is optimized as if the two
operands were the *same* run-time quantity. The relational operator is thus
evaluated against a single value (`X ⋈ X`) instead of against the two distinct
sub-expression results.

This is not specific to division or to the value zero: it also reproduces with
`+`, `^`, `%`, and non-zero folded values, wherever two distinct sub-expressions
fold to the same constant on the two sides of a relation.

## Impact

- `tcpdump`/libpcap callers that compile with optimization (the common case,
  including `tcpdump` by default) get a filter that silently matches all packets
  or no packets, diverging from the filter's actual meaning.

carrerasdarren-cell:
Fixes #1699.

`opt_root()` follows chains of jump blocks whose true and false successors are identical. Each block still executes its statements, but the old loop retained statements only from the first and final blocks. Statements on intermediate blocks were discarded, so the surviving comparison could run with stale BPF register values.

This change preserves every traversed statement list in execution order and adds one-packet execution regressions for both undercapture and overcapture.

On unmodified current master, the optimized true case returns 0 instead of 1508 and the optimized false case returns 1508 instead of 0. Both unoptimized controls pass. With optimizer debugging enabled, the graph after `opt_root()` shows that the old code drops the intermediate `ld #0x1`; the patched graph preserves it immediately before the final `jgt x`.

Testing: `TESTRUN_JOBS=4 make check` (14,613 passed, 6,585 platform skips, 0 failed).

### DIFF
--- a/optimize.c
+++ b/optimize.c
@@ -2157,14 +2157,23 @@ opt_root(struct block **b)
 {
 	struct slist *tmp, *s;
 
-	s = (*b)->stmts;
-	(*b)->stmts = 0;
-	while (BPF_CLASS((*b)->s.code) == BPF_JMP && JT(*b) == JF(*b))
-		*b = JT(*b);
+	/*
+	 * A block with identical successors still executes its statements.
+	 * Preserve each statement list in order while collapsing the chain.
+	 */
+	s = 0;
+	for (;;) {
+		tmp = (*b)->stmts;
+		(*b)->stmts = 0;
+		if (s == 0)
+			s = tmp;
+		else if (tmp != 0)
+			sappend(s, tmp);
 
-	tmp = (*b)->stmts;
-	if (tmp != 0)
-		sappend(s, tmp);
+		if (BPF_CLASS((*b)->s.code) != BPF_JMP || JT(*b) != JF(*b))
+			break;
+		*b = JT(*b);
+	}
 	(*b)->stmts = s;
 
 	/*
--- a/testprogs/TESTrun
+++ b/testprogs/TESTrun
@@ -25810,6 +25810,18 @@ my @filter_accept_blocks = (
 # * results (mandatory, array): the list of program filter results to expect
 my @filter_apply_blocks = (
 	{
+		name => 'constant_true_subexpressions_on_pppoed',
+		savefile => 'pppoe.pcap',
+		expr => '1/2=1/3&&1>4/5&&1>6/7',
+		results => [1508],
+	},
+	{
+		name => 'constant_false_subexpressions_on_pppoed',
+		savefile => 'pppoe.pcap',
+		expr => '1/2=1/3&&1>4/5&&1=6/7',
+		results => [0],
+	},
+	{
 		name => 'pppoed_nullary_on_ctp',
 		savefile => 'loopback.pcap',
 		expr => 'pppoed',


infrastation:
Thank you for preparing these changes. This pull request changes a function in a very common code path, so it will require more time and work to become obviously correct and safe. Two easy clean-ups would be adding a right-sized commit message to the commit and limiting the test to one packet (there is no point in computing the same constant 10 times).

A more difficult thing would be to prove that `opt_root()` is indeed incorrect before the change and correct after the change. Unfortunately, the testing framework currently does not provide means to test the optimiser in a granular enough fashion, so this facility would have to be developed first. In practical terms, this is not going to be done before the 1.11.0 release, but this function has been this way since at least 1999, so this problem is not new.

carrerasdarren-cell:
Thank you. I updated the existing commit and branch:

- expanded the commit message
- rebased onto current master
- reduced both regressions to one packet
- documented the `opt_root()` invariant beside the loop

I also rebuilt with `--enable-optimizer-dbg` to isolate the optimizer step. On unmodified master, the graph immediately after `opt_root()` is:

```
ldx #0x0
txa
jgt x
```

With this patch it is:

```
ldx #0x0
txa
ld #0x1
jgt x
```

Thus the old function drops the intermediate statement while collapsing the chain; the patch retains it in execution order. The one-packet negative controls fail in both directions on unmodified master and pass with the patch. The full suite reports 14,613 passed, 6,585 skipped, and 0 failed.

infrastation:

PR #1711, opened 2026-07-28
Validate both named services in portrange filters

### LINKED ISSUE #1440: portrange parsing and validation discrepancies
The `port` qualifier accepts either a number or a name, and the name must match the protocol if a protocol is specified (this is what "If a name is used, both the port number and protocol are checked." means, as far as I understand):
```
$ ./testprogs/filtertest IPV4 udp port http
filtertest: port 'http' is tcp
```
Likewise, the `portrange` qualifier also accepts a number or a name for either of its two arguments and seems to follow the same logic, but not quite entirely:
* `port 25` works as expected
* `port smtp` works as expected
* `portrange 25` works (undocumented syntax)
* `portrange smtp` returns an error ("port range 'smtp' contains no hyphen")
* `portrange 25-25` works as expected
* `portrange 25-smtp` works as expected
* `portrange smtp-25` works as expected
* `portrange smtp-smtp` works as expected
* `udp port 25` works as expected
* `udp port smtp` returns an error as expected ("port 'smtp' is tcp")
* `udp portrange 25` works (undocumented syntax)
* `udp portrange smtp` returns an error about a missing hyphen, not an error about protocol mismatch
* `udp portrange 25-25` works as expected
* `udp portrange 25-smtp` compiles without an error, which is a failure to fail
* `udp portrange smtp-25` compiles without an error, which is a failure to fail
* `udp portrange smtp-smtp` returns an error as expected ("port in range 'smtp-smtp' is tcp")
* `udp portrange smtp-domain` compiles without an error, which is a failure to fail

carrerasdarren-cell:
## Summary

Preserve the resolved protocol for each endpoint of a named `portrange`, then validate both endpoints against an explicit `tcp`, `udp`, or `sctp` qualifier.

Previously, resolving the second endpoint overwrote the first endpoint's protocol. If the two endpoints differed, the result became unspecified, so expressions such as `udp portrange smtp-25` and `udp portrange 25-smtp` compiled even though `udp port smtp` was rejected on systems where `smtp` is TCP-only.

Numeric endpoints and service names available for both TCP and UDP keep their existing behavior. Unqualified ranges also keep the previous combined-protocol behavior.

The regression tests use `getservbyname()` to run only when the host service database exposes an unambiguous service name.

Fixes #1440.

## Validation

- Linux Autotools build
- Full Linux `make check`: 14,629 passed, 6,569 skipped, 0 failed
- macOS CMake build
- Focused Linux regressions for mismatched services in both endpoint positions
- Focused valid-range smoke tests for numeric, dual-protocol service, and unqualified ranges
- `perl -c testprogs/TESTrun`

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -7114,12 +7114,11 @@ stringtoport(compiler_state_t *cstate, const char *string, size_t string_size,
  */
 static void
 stringtoportrange(compiler_state_t *cstate, const char *string,
-    bpf_u_int32 *port1, bpf_u_int32 *port2, int *proto)
+    bpf_u_int32 *port1, bpf_u_int32 *port2, int *proto1, int *proto2)
 {
 	const char *hyphen_off;
 	const char *first, *second;
 	size_t first_size, second_size;
-	int save_proto;
 
 	if ((hyphen_off = strchr(string, '-')) == NULL)
 		bpf_error(cstate, "port range '%s' contains no hyphen", string);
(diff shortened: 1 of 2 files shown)


infrastation:

PR #1713, opened 2026-07-28
pcapng: avoid binary timestamp scaling overflow

soul-sol:
Binary pcapng timestamps multiply the fractional timestamp by the requested resolution before dividing by the file resolution. For high binary resolutions that multiplication wraps modulo 2^64 and produces an incorrect timestamp.

For example, the included pcapng file uses `if_tsresol=2^-63` and a fractional timestamp of `2^63-1`. With microsecond output, the expected fraction is 999999, but the old calculation produces 1.

This change keeps the existing fast multiplication path when it cannot overflow. For the overflow case, it builds the quotient one multiplier bit at a time while retaining a reduced remainder. The input constraints (`file_tsresol <= 2^63`, `frac < file_tsresol`) keep every intermediate value within `uint64_t`.

The internal scaler is now exercised directly through `translatetest` with exact expected values for zero, the smallest fraction, half, the largest fraction in microsecond and nanosecond modes, both sides of the direct-multiplication overflow boundary, and a small binary resolution. The pcapng fixture remains as an end-to-end packet-reading and sanitizer regression.

Validation:

- baseline Clang unsigned-overflow/UB sanitizer: aborts on the fixture at `sf-pcapng.c`
- patched sanitizer build: fixture exits 0
- eight direct expected-value timestamp tests: pass
- short full `TESTrun`: 4,475 passed, 0 failed, 942 skipped
- exhaustive 12,800,000 arithmetic comparisons against a `__uint128_t` reference: all passed


sf-pcapng.c
```diff
@@ -248,6 +248,48 @@ struct pcap_ng_sf {
...
+ * Scale a fractional timestamp from a power-of-2 resolution to the
+ * user-requested resolution without overflowing.
+ */
+static uint64_t
+scale_binary_timestamp(uint64_t frac, uint64_t user_tsresol,
+    uint64_t file_tsresol)
+{
+	uint64_t scaled, remainder, bit;
+
+	if (frac <= UINT64_MAX / user_tsresol)
```
infrastation, line 261: If `UINT64_MAX / user_tsresol` is a constant in `pcap_ng_next_packet()`, it would be more effective to test the condition once in the latter function instead of doing it for every packet.

sf-pcapng.c
```diff
@@ -248,6 +248,48 @@ struct pcap_ng_sf {
...
+			scaled++;
+		}
+		if (user_tsresol & bit) {
+			remainder += frac;
+			if (remainder >= file_tsresol) {
+				remainder -= file_tsresol;
+				scaled++;
+			}
+		}
+	}
```
infrastation, line 289: What method does this block implement and how much performance drop does it mean compared to the current implementation?

testprogs/TESTrun
```diff
@@ -25809,6 +25809,12 @@ my @filter_accept_blocks = (
 #   "filtertest -r", this should not have too many packets
 # * results (mandatory, array): the list of program filter results to expect
 my @filter_apply_blocks = (
+	{
+		name => 'pcapng_binary_timestamp_resolution',
+		savefile => 'pcapng-tsresol-bin-overflow.pcapng',
+		expr => 'len >= 0',
```
infrastation, line 25815: The new test's purpose is substantially different from the existing tests in this list, so it would need a comment to say which code path it tests, and that without a certain sanitiser flag it is a no-op.

infrastation:
Thank you for preparing these changes.

Because this change affects a busy code path, it would need to be tested more substantially. (One invalid result is sufficient to prove the presence of a bug with a 100% confidence, more than one valid result and a test method are required to prove the absence of the bug with confidence that is acceptably close to 100%.) Note that neither `fuzz_pcap` break/fix evidence nor "timestamp reproducer" nor "12,800,000 arithmetic comparisons" are a part of the commit. The only test included in the commit can prove — under non-default conditions — that `pcap_ng_next_packet()` for one specific packet instead of triggering a sanitiser error produces a value, but it does not prove that the value is correct for all possible packets or even the specific packet. To test the modified code path better, it would be necessary to define multiple tests, each of which would correlate specific input arguments with a specific result value, and the tests would need to cover a reasonable dataset, including nearly-minimal/nearly-maximal/edge/corner/off-by-one/etc cases.

One potential way to do this would be via `translatetest` by driving the new function directly, although this would cover only a part of the affected code path. Another potential way would be using a test helper (a new binary or an option of `filtertest`) that reads every packet from a file and prints the packet timestamp, although this way it would be difficult to correlate the timestamp components with each final value. This needs a bit more thinking.

Meanwhile it would be useful to provide exact steps to reproduce the problem and to make sure the commit has a right-sized commit message (currently empty).

soul-sol:
Thank you. I updated the commit in response to this review.

The scaler is now an internal `pcapint_` helper driven directly by `translatetest`. Eight committed cases assert exact values for zero, the minimum fraction, half, the maximum fraction in microsecond and nanosecond modes, the last non-overflowing fast-path input, the first wide-path input, and a small 2^-2 resolution. The original pcapng fixture remains to exercise `pcap_next_ex()` end to end. I also replaced the one-line commit text with a full rationale and test description.

Exact sanitizer reproduction from a clean clone (the fixture can first be saved from commit `9248af5e`):

```sh
git show 9248af5e:tests/filter/pcapng-tsresol-bin-overflow.pcapng > /tmp/pcapng-tsresol-bin-overflow.pcapng
git switch --detach a9b2c25e
cmake -S . -B build-san -DCMAKE_C_COMPILER=clang -DBUILD_SHARED_LIBS=OFF -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_FLAGS="-fsanitize=undefined,unsigned-integer-overflow -fno-sanitize-recover=all -fno-omit-frame-pointer" -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=undefined,unsigned-integer-overflow"
cmake --build build-san --target fuzz_pcap -j4
UBSAN_OPTIONS=halt_on_error=1 build-san/run/fuzz_pcap /tmp/pcapng-tsresol-bin-overflow.pcapng
```

On `a9b2c25e`, that exits 134 with:

```text
sf-pcapng.c:1485:8: runtime error: unsigned integer overflow: 9223372036854775807 * 1000000 cannot be represented in type uint64_t
```

Repeating the build and command at `9248af5e` exits 0. The eight new `translatetest` cases all pass, and `TESTRUN_SHORT=1 TESTRUN_JOBS=8 testprogs/TESTrun` reports 4,475 passed, 0 failed, 942 skipped on macOS ARM64.

infrastation:

PR #1718, opened 2026-08-08
Place all long jumps in one pass over the flowgraph

### LINKED ISSUE #1255: Simple filter of 1000 hosts takes O(n²) to compile
When giving tcpdump a filter of 1000 hosts to ignore, the compilation of the filter takes at least O(n²) time and RAM.

I would have expected it to take O(n) or worst case O(n log n) for such a simple filter.

The memory usage is also drastic in the compilation phase (8GB for 3000 hosts, 1 GB for 1000 hosts).

After compilation of the filter both CPU usage and memory usage are as expected.

## Build
~~~
$ git clone https://github.com/the-tcpdump-group/libpcap
$ cd libpcap
$ ./autogen.sh 
$ ./configure 
$ make -j5
$ sudo make install
$ cd ..

$ git clone https://github.com/the-tcpdump-group/tcpdump
$ cd tcpdump
$ ./autogen.sh 
$ ./configure 
$ make -j5
$ sudo make install
$ cd ..
~~~
## Version
~~~
$ tcpdump  --version
tcpdump version 5.0.0-PRE-GIT
OpenSSL 3.0.2 15 Mar 2022
~~~
## MCVE
~~~
$ filter() {
    # $1 = How many entries?                                                             
    perl -e '                                                                            
        $pre = "not (";                                                                  
        $post = ")";                                                                     
        $join = " and ";                                                                 
        sub hostport {                                                                   
          $host = sprintf "%d.%d.%d.%d", rand()*255,rand()*255,rand()*255,rand()*255;    
          $port = sprintf "%d", rand()*65535;                                            
          return "(host $host and port $port)";                                          
        }                                                                                
        print $pre, join($join,map { hostport() } 1..shift), $post;                      
    ' $1
}
$ export -f filter
$ sudo time -v tcpdump -c 1 `filter 300`
$ sudo time -v tcpdump -c 1 `filter 1000`
$ sudo time -v tcpdump -c 1 `filter 3000`
~~~
Measure the time to compile the filter for 300, 1000, 3000 hosts.

cicku:
A branch whose target is more than 255 instructions away needs an extra `BPF_JA`, and reserving room for one means laying the program out again. `convert_code_r()` returned on the first such branch it found, so a program needing N long jumps took N passes over the whole flowgraph to place them.  That is most of the time spent compiling a filter with a few hundred "or"ed host/port pairs, with or without the optimizer.

Attempt to fix https://github.com/the-tcpdump-group/libpcap/issues/1255.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -58,6 +58,9 @@ DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
         and DLT_ENC savefiles. (issue #74)
       Match all AF_INET6 values when filtering DLT_PFLOG savefiles.
       Fix optimization of "jset #0xffffffff".
+      Place all long jumps in one pass, not one pass per long jump, so
+        that compiling a large filter is no longer quadratic in the
+        number of them. (issue #1255)
     Capture file reading:
       Fix misaligned accesses in processing Linux USB captures (issue
         #1634, reported by FuzzAnything Organization
--- a/optimize.c
+++ b/optimize.c
@@ -2676,10 +2676,24 @@ int bids[NBIDS];
 #endif
 
 /*
- * Returns true if successful.  Returns false if a branch has
- * an offset that is too large.  If so, we have marked that
- * branch so that on a subsequent iteration, it will be treated
- * properly.
+ * Returns true if successful.  Returns false if any branch has an
+ * offset that is too large.  If so, we have marked every such branch
+ * so that on a subsequent iteration, they will be treated properly.
+ *
+ * Marking all of them in one pass, rather than returning as soon as the
+ * first one turns up, is what keeps this from costing one pass over the
+ * whole flowgraph per long jump.  It is safe because a long jump only
+ * ever makes the code longer, so a branch that does not fit now will
+ * not fit in a later pass either.
+ *
+ * A branch marked in this pass does not get its extra jump until the
+ * next pass, so the offsets computed here do not yet account for the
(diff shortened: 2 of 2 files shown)


infrastation:
Thank you for preparing this change. The commit message conceptually makes sense. The comment before the function makes sense except "A pass that marked something computes offsets that are too small, so a later pass can find more branches, but in practice only a few passes are needed." — this would need a bit more context or rewording to be comprehensible. The substantial part of the code changes looks roughly consistent with the comments, but I am not ready to reason whether the modified code would correctly handle all input in all scenarios and edge cases.

Unfortunately, there is no direct coverage of `optimize.c` at this point, and none of the existing tests are complex enough to exercise the target scenario indirectly even once. This is a very common code path and it would be highly desirable to cover this code with at least some tests before changing it.

As far as I understand, that would be generally difficult because it would require serialisation of graphs that are the input, the output and the intermediate data within iterations and recursions. However, in this specific case what `convert_code_r()` does with the current block at one recursion level looks mostly independent of what it does at other recursion levels with other blocks, so if block-scope code could be made an independent function or several functions, it would be much easier to test each such independent function using one sequence of statements as input and another as output. Of course, it would take some prototyping to confirm whether this approach works as expected. I may be able to do that later.

cicku:
I updated the message a bit. Let me know what coverage test do you wish to have. If you want to implement, I can definitely wait.

infrastation:
Thank you, let me have a look later, I am currently busy with other work.

PR #1719, opened 2026-08-09
Fix BPF optimizer handling of VLAN metadata offsets

### LINKED ISSUE #1513: BPF optimizer produces incorrect offsets around "vlan" when the check involves BPF extensions.
Input filter is "(ether proto 0x8892 or (vlan and ether proto 0x8892))". Output BPF:

```
(000) ldh      [12]
(001) jeq      #0x8892          jt 9    jf 2
(002) jeq      #0x8100          jt 7    jf 3
(003) jeq      #0x88a8          jt 7    jf 4
(004) jeq      #0x9100          jt 7    jf 5
(005) ldb      [-4092]
(006) jeq      #0x1             jt 7    jf 10
(007) ldh      [16]
(008) jeq      #0x8892          jt 9    jf 10
(009) ret      #65536
(010) ret      #0
```

Instruction 7 uses offset 16 (12 + 4) even though the SKF_AD_VLAN_TAG_PRESENT extension was used in instruction 5, which would mean that there is no tag present in the packet data. It should use 12 in that case. The non-aware version (with handle obtained from `pcap_compile_dead()`) produces this instead, which I presume is the input prior to injecting the extension code:

```
(000) ldh      [12]
(001) jeq      #0x8892          jt 7	jf 2
(002) jeq      #0x8100          jt 5	jf 3
(003) jeq      #0x88a8          jt 5	jf 4
(004) jeq      #0x9100          jt 5	jf 8
(005) ldh      [16]
(006) jeq      #0x8892          jt 7	jf 8
(007) ret      #262144
(008) ret      #0
```

The other relevant compiler transformation is on the input "vlan and ether proto 0x8892", which is handled correctly. Here we get the appropriate code to check for SKF_AD_VLAN_TAG_PRESENT first, then for one of the VLAN EtherType values, saving the 4-byte offset adjustment to memory:

```
(000) ld       #0x0
(001) st       M[1]
(002) ldb      [-4092]
(003) jeq      #0x1             jt 10   jf 4
(004) ld       #0x4
(005) st       M[1]
(006) ldh      [12]
(007) jeq      #0x8100          jt 10   jf 8
(008) jeq      #0x88a8          jt 10   jf 9
(009) jeq      #0x9100          jt 10   jf 14
(010) ldx      M[1]
(011) ldh      [x + 12]
(012) jeq      #0x8892          jt 13   jf 14
(013) ret      #65536
(014) ret      #0
```

I would expect the first expression ("ether proto 0x8892 or (vlan and ether proto 0x8892)") to produce something like the first 2 instructions, followed by the entirety of that previous output. I am already aware of the [FAQ entry on the "vlan" keyword](https://www.tcpdump.org/faq.html#q13), which is why I knew not to expect something like "(vlan or not vlan) and ether proto 0x8892)" to work.

moodyastra:
Fixes #1513.

`or_pullup()` can move a block that defines scratch-memory values across a path where those values are already live. With Linux VLAN metadata, this moves the four-byte offset adjustment onto the metadata-present path, so the optimized filter reads the EtherType at byte 16 instead of byte 12 and can produce a false match.

Prevent that pull-up when the moved block defines scratch words used by the crossed block. Add bytecode and execution regressions covering metadata-stripped VLAN, inline VLAN, positive matches, the false-positive probe, and an ordinary nonmatch.

Validation:

- focused optimized and unoptimized regression tests
- `TESTRUN_JOBS=4 perl testprogs/TESTrun`
- `TESTRUN_JOBS=4 make -s check`

AI assistance was used for analysis and drafting. The issue was reproduced locally against current master, and the final patch was verified with the project test suite.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -58,6 +58,7 @@ DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
         and DLT_ENC savefiles. (issue #74)
       Match all AF_INET6 values when filtering DLT_PFLOG savefiles.
       Fix optimization of "jset #0xffffffff".
+      Fix optimizer handling of Linux VLAN metadata offsets (issue #1513).
     Capture file reading:
       Fix misaligned accesses in processing Linux USB captures (issue
         #1634, reported by FuzzAnything Organization
--- a/optimize.c
+++ b/optimize.c
@@ -1929,19 +1929,17 @@ or_pullup(opt_state_t *opt_state, struct block *b, struct block *root)
 		if ((*samep)->val[A_ATOM] == val)
 			break;
 
-		/* XXX Need to check that there are no data dependencies
-		   between dp0 and dp1.  Currently, the code generator
-		   will not produce such dependencies. */
 		samep = &JF(*samep);
 	}
-#ifdef notdef
-	/* XXX This doesn't cover everything. */
-	for (i = 0; i < N_ATOMS; ++i)
-		if ((*samep)->val[i] != pred->val[i])
-			return;
-#endif
-	/* Pull up the node. */
+	/*
+	 * Pulling this block ahead of diffp makes its statements run on paths
+	 * that did not previously execute them.  Do not move definitions that
+	 * are live on any of those paths.
+	 */
 	pull = *samep;
+	if ((pull->def & (*diffp)->in_use) != 0)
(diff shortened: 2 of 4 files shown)


infrastation:

PR #1720, opened 2026-08-11
pcap-bpf: guard monitor_mode() against negative ifm_count from SIOCGIFMEDIA

Alb3e3:
## Problem

`req.ifm_count` is declared as `int` in `struct ifmediareq`. The existing check in `monitor_mode()` only rejects zero:

```c
if (req.ifm_count == 0) {
    /* No media types. */
    close(sock);
    return (PCAP_ERROR_RFMON_NOTSUP);
}
/* ... */
media_list = malloc(req.ifm_count * sizeof(*media_list));
```

A negative value from `SIOCGIFMEDIA` (e.g. due to a kernel bug or a malformed response on a virtualised or emulated NIC) passes the guard. The implicit conversion from `int` to `size_t` in the `malloc()` call then wraps the negative value to a very large `size_t`, typically causing `malloc()` to fail and return `NULL` — which is caught by the subsequent NULL check. The failure is therefore unlikely to be exploitable in practice, but produces a misleading `errno`-based diagnostic rather than a clear indication that the ioctl returned an implausible count.

## Fix

Change the guard from `== 0` to `<= 0` so that any non-positive `ifm_count` — whether zero (no media types) or negative (unexpected kernel response) — takes the `PCAP_ERROR_RFMON_NOTSUP` path before the problematic `malloc()` is reached. The comment is updated to explain the additional case.

This code is only compiled on platforms that define `HAVE_BSD_IEEE80211` (BSD and macOS).

### DIFF
--- a/pcap-bpf.c
+++ b/pcap-bpf.c
@@ -3199,9 +3199,12 @@ monitor_mode(pcap_t *p, int set)
 			return (PCAP_ERROR);
 		}
 	}
-	if (req.ifm_count == 0) {
+	if (req.ifm_count <= 0) {
 		/*
-		 * No media types.
+		 * No media types, or negative count (guard against a kernel
+		 * bug or unexpected ioctl return value: ifm_count is int,
+		 * so a negative value would wrap to a huge size_t in the
+		 * malloc() call below).
 		 */
 		close(sock);
 		return (PCAP_ERROR_RFMON_NOTSUP);


guyharris:
> The failure is therefore unlikely to be exploitable in practice

It's unlikely to occur in practice, as it requires either that the kernel code either stuff a ridiculously large count (> 2^31-1) or a negative count there.

They really *should* have made the `ifm_count` field an `unsigned int`.

Alb3e3:
Agreed on both points — the  field should have been unsigned from the start. The check is defensive belt-and-suspenders: since the field is `int` in the kernel ABI and we control neither the kernel nor older versions, treating a negative value the same as zero costs one comparison and avoids relying on the caller to never see a negative from the kernel. Happy to add a comment to the code making that rationale explicit if that would help.

guyharris:
> Happy to add a comment to the code making that rationale explicit if that would help.

It might.

Not sure whether the right response is to say "no monitor mode support" or "bogus media type count".

infrastation:

PR #1721, opened 2026-08-12
enumeratetest

infrastation:
More tests and a couple minor improvements.


infrastation:
DLT enumeration depends on the OS-specific preprocessing of `pcap/dlt.h`, so the expected result needs to be composed using similar conditionals.

infrastation:
Also the new Perl function is off. This is going to require at least one other revision.

infrastation:
This should work better. Tested manually on Haiku, GNU/Hurd, macOS 25.5.0, Solaris 10 and Solaris 11.4 CBE.

infrastation:
This is going to be merged soon (needs another minor fixup first).

PR #1725, opened 2026-08-18
pcap-util: clean up the byte-swapping macros.

guyharris:
Use the same code style as PCAP_BSWAP_64() for PCAP_BSWAP_32() and PCAP_BSWAP_16().

For all of them:

- use uintNN_t types rather than u_short or u_int;
- use UINTnn_C macros for the byte masks, so that the appropriate suffix is added (for example, if uint64_t is unsigned long, it will append UL rather than ULL to the constant if necessary);
- put the (uintNN_t)(y) in parentheses;
- put a final cast to the desired type around the entire expression, just as we were already doing for PCAP_BSWAP_16().

Remove a no-longer-applies comment and and a now-unnecessary cast from testprogs/translatetest.c.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2843,6 +2843,48 @@ endif()
 # To get a correct builtin test result, the source must be compiled and linked,
 # which in CMake means check_c_source_compiles().
 #
+# Different builtins require different methods of testing, see the comments
+# about HAVE_BUILTIN_BSWAPnn in pcap-util.h and HAVE_BUILTIN_CTZ in optimize.c.
+#
+check_c_source_compiles(
+"int main(int argc, char **argv)
+{
+	switch (argc) {
+	case __builtin_bswap16(0):
+		return 0;
+	default:
+		return 1;
+	}
+}
+"
+    HAVE_BUILTIN_BSWAP16)
+
+check_c_source_compiles(
+"int main(int argc, char **argv)
+{
+	switch (argc) {
+	case __builtin_bswap32(0):
+		return 0;
+	default:
+		return 1;
+	}
+}
+"
+    HAVE_BUILTIN_BSWAP32)
+
+check_c_source_compiles(
+"int main(int argc, char **argv)
+{
+	switch (argc) {
+	case __builtin_bswap64(0):
+		return 0;
+	default:
+		return 1;
+	}
+}
+"
+    HAVE_BUILTIN_BSWAP64)
+
 check_c_source_compiles("int main(void) {return __builtin_ctz(1);}" HAVE_BUILTIN_CTZ)
 
 #
--- a/configure.ac
+++ b/configure.ac
@@ -2350,6 +2350,72 @@ AC_LBL_DEVEL(V_CCOPT)
 # To get a correct builtin test result, the source must be compiled and linked,
 # which in Autoconf means AC_LINK_IFELSE().
 #
+# Different builtins require different methods of testing, see the comments
+# about HAVE_BUILTIN_BSWAPnn in pcap-util.h and HAVE_BUILTIN_CTZ in optimize.c.
+#
+AC_MSG_CHECKING([whether the compiler implements __builtin_bswap16()])
+AC_LINK_IFELSE(
+	[AC_LANG_SOURCE([[
+int main(int argc, char **argv)
+{
+	switch (argc) {
+	case __builtin_bswap16(0):
+		return 0;
+	default:
+		return 1;
+	}
+}
+	]])],
+	[
+		AC_MSG_RESULT(yes)
+		AC_DEFINE(HAVE_BUILTIN_BSWAP16, 1,
+			[Define to 1 if you have the '__builtin_bswap16' compiler builtin.])
+	],
+	[AC_MSG_RESULT(no)]
+)
+
+AC_MSG_CHECKING([whether the compiler implements __builtin_bswap32()])
+AC_LINK_IFELSE(
+	[AC_LANG_SOURCE([[
+int main(int argc, char **argv)
+{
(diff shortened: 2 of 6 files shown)


infrastation:
Would you prefer my byte-swapping builtins change to be a separate commit in this pull request, or a separate pull request or kept out of scope for the time being?

guyharris:
> Would you prefer my byte-swapping builtins change to be a separate commit in this pull request, or a separate pull request or kept out of scope for the time being?

If you have that change working, might as well add it to this pull request.

guyharris:
According to the [Oracle of Matt Godbolt](https://godbolt.org), the latest versions of MSVC for IA-32 ("x86"), x86-64 ("x64"), and ARM64 ("arm64"), when run with `/O1` recognize the idioms from the macros and generate inline machine code to do the swaps without shifting and masking, but older versions don't recognize the 64-bit swap.

However, the older versions *do* recognize the functions `_byteswap_ushort()`, `_byteswap_ulong()`, and `_byteswap_uint64()` and generate inline machine code, even for the 64-bit version, with `/O1`, but generate out-of-line calls to the library implementations without the optimizer. (The library routines are C-callable C++ functions that do the exact same shifting  and masking as our macros, but have a bunch of statements that add the results together, so they do the same thing as the macros.)

guyharris:
GCC, at least, doesn't care about optimization, it recognizes the builtins even with *no* optimization enabled.

MSVC's behavior... varies, by compiler version and optimization level:

- at `/O2`, it doesn't recognize the 64-bit version of our macros, until Visual Studio 17.7;
- at `/O1`, it doesn't recognize the 16-bit or 64-bit versions of our macros until VS 17.4 and 17.7, respectively, *and* doesn't recognize the `_byteswap` routines and generate inline machine code;
- with no optimization, it doesn't recognize any of our macros, *and* doesn't recognize the `_byteswap` routines and generate inline machine code.

So it looks as if the *only* benefit of using Microsoft's not-really-100%-builtin routines is that, with older versions of Visual Studio will generate optimized machine code for the 64-bit byte swap, and the downside is that out-of-line procedure calls are generated for *all* of them unless `/O2` is specified - `/O1` doesn't suffice (plus it's more configure-time crud or `#if` compiler tests and additional gunk to use them in the header).

I vote for not using the Microsoft routines unless and until Microsoft promotes them to real builtins recognized even at lower optimization levels (or unless and until they just decide to go with as much LLVM as possible and end up supporting the GCC-style builtins).

guyharris:
It seems to pass all the tests so far, with only OpenBSD-on-MIPS still running; if that finishes and passes (or dies from a random memory error), is there anything more to do here, or should we just merge this?

infrastation:
Re MSVC, I would prefer not to explore its dimensions. Re merging, one potential thing to add could be comments to say that all branches of a byte-swapping macro must be usable as a `case` label (hence the need for real built-ins) and that even the default branch in specific use cases can get automatically optimised into fast machine code.

Actually, this made me realise that the build systems must test the byte-swapping builtins as a case label, so my commit will need another revision.

guyharris:
> Re MSVC, I would prefer not to explore its dimensions. 

Yup, added complexity for minimal gain and some loss.

Plus...

> ll branches of a byte-swapping macro must be usable as a `case` label (hence the need for real built-ins)

...which completely rules the Microsoft not-100%-builtins out.

guyharris:
> Re merging, one potential thing to add could be comments to say that all branches of a byte-swapping macro must be usable as a case label (hence the need for real built-ins) and that even the default branch in specific use cases can get automatically optimised into fast machine code.

Done, in a separate commit.

infrastation:
Thank you, let me redo my part after a good sleep.

CMakeLists.txt
guyharris, line 2842: You might want to add a note to the comment indicating that the `switch` statements are there to test whether the `_builtin_bswapN()` functions are builtin *and*, when passed a compile-time constant argument, evaluate to a compile-time constant, as we have code that requires that to be the case.

configure.ac
guyharris, line 2349: And add that comment here too.

infrastation:
The commit message says that. Perhaps the comment in `pcap-util.h` could be another place to explain this nuance exactly once.

guyharris:
> The commit message says that. Perhaps the comment in `pcap-util.h` could be another place to explain this nuance exactly once.

The comment currently says:

```
 * These macros are used to generate case labels in switch statements,
 * so they can't be done as inline functions, and, when passed a
 * compile-time constant argument, must produce a compile-time constant
 * result. This means that they must either use builtin functions that
 * are recognized by the compiler as producing constant results when
 * passed a constant argument, or must be done with expressions that
 * evaluate to compile-time constants when passed a compile-time
 * constant.
 *
 * Newer versions of GCC, Clang, Visual Studio C/C++, and possibly
 * some other compilers recognize the expressions as byte-swapping
 * idioms and generate optimized machine-specific code for architectures
 * that include instructions that can be used to swap bytes.   
```

guyharris:
But my concern is that somebody looking at the configure script or CMakeLists.txt file might not have seen the comment and might be wondering "so why are we doing this `switch`-statement stuff?", and possibly even removing it as apparent overkill.

infrastation:
Let me prepare yet another revision tomorrow. Would "see the comment in pcap-util.h" do the job?

guyharris:
> Let me prepare yet another revision tomorrow. Would "see the comment in pcap-util.h" do the job?

That should suffice.

infrastation:
Does this revision look good enough?

guyharris:
> Does this revision look good enough?

Looks good to me. Merge?

infrastation:
Thank you.

PR #1728, opened 2026-08-20
Harden libpcap against integer overflow and bounds errors

mohammidharaz:
Here is the description in English for your Pull Request:

Title:
Harden libpcap against integer overflow and bounds errors

Description:

This pull request hardens multiple code paths across libpcap against integer overflow, buffer overflow, and bounds-validation issues to improve security and overall stability.

Key Changes:

Allocation Checks: Added NULL checks following memory allocations (malloc, strdup) to safely handle failure cases and prevent null pointer dereferences.

Integer Overflow Safeguards: Introduced explicit bounds checks on size calculations—such as for ring buffer allocations, BPF filter calculations, and OID structures—to prevent arithmetic overflows before allocation.

Buffer & String Safety:

Replaced unsafe string functions like strcpy and strcat with bounded alternatives (pcapint_strlcpy, pcapint_strlcat).

Added validation for control message lengths (cmsg_len) and netfilter attribute offsets (nfattr) to prevent out-of-bounds reads and writes.

Clamped copy operations (memcpy) against maximum allowed snapshot limits to protect against memory overru

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -7031,6 +7031,10 @@ stringtoport(compiler_state_t *cstate, const char *string, size_t string_size,
 		 * Not a valid number; try looking it up as a port.
 		 */
 		cpy = malloc(string_size + 1);	/* +1 for terminating '\0' */
+		if (cpy == NULL) {
+			longjmp(cstate->top_ctx, 1);
+			/*NOTREACHED*/
+		}
 		memcpy(cpy, string, string_size);
 		cpy[string_size] = '\0';
 		tcp_port = nametoport(cstate, cpy, IPPROTO_TCP);
--- a/pcap-bpf.c
+++ b/pcap-bpf.c
@@ -2017,8 +2017,8 @@ pcap_activate_bpf(pcap_t *p)
 					status = PCAP_ERROR;
 					goto bad;
 				}
-				strcpy(wltdev, "wlt");
-				strcat(wltdev, p->opt.device + 2);
+				pcapint_strlcpy(wltdev, "wlt", strlen(p->opt.device) + 2);
+				pcapint_strlcat(wltdev, p->opt.device + 2, strlen(p->opt.device) + 2);
 				free(p->opt.device);
 				p->opt.device = wltdev;
 			}
@@ -2800,8 +2800,8 @@ check_bpf_bindable(const char *name)
 			    errno, "malloc");
 			return (-1);
 		}
-		strcpy(en_name, "en");
-		strcat(en_name, name + 3);
+		pcapint_strlcpy(en_name, "en", en_name_len + 1);
+		pcapint_strlcat(en_name, name + 3, en_name_len + 1);
 		fd = bpf_open_and_bind(en_name, errbuf);
 		free(en_name);
 	} else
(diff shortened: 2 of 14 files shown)


mohammidharaz:
> **Follow-up — Security Hardening**
>
> I have completed an additional review of the affected code paths and addressed the portability issues identified by the CI builds.
>
> The latest changes harden `libpcap` against integer-overflow conditions and unsafe size calculations, including:
>
> * BPF filter size calculations in `pcap-linux.c` and `pcap-rpcap.c`
> * Frame-header ring allocation size calculations
> * Remote-capture snapshot length validation
> * Compatibility with platforms where integer types have narrower ranges
>
> I also verified the changes locally with:
>
> ```text
> git diff --check
> make -s clean
> make -s CFLAGS=-Werror
> ```
>
> The NetBSD/aarch64 build initially exposed two `-Werror` portability issues, which have now been addressed.
>
> The NetBSD/mips failure observed in build 624 was caused by:
>
> ```text
> No space left on device
> ```
>
> and appears to be an infrastructure/worker issue rather than a source-code failure.
>
> I have pushed the fixes to the `security-hardening` branch and will continue monitoring the CI results.
>
> **This PR is intended as proactive security hardening and defensive validation.**

gencode.c
```diff
@@ -7031,6 +7031,10 @@ stringtoport(compiler_state_t *cstate, const char *string, size_t string_size,
 		 * Not a valid number; try looking it up as a port.
 		 */
 		cpy = malloc(string_size + 1);	/* +1 for terminating '\0' */
+		if (cpy == NULL) {
```
guyharris, line 7034: This needs to call `bpf_set_error()` to set the error string.

pcap-bpf.c
```diff
@@ -2017,8 +2017,8 @@ pcap_activate_bpf(pcap_t *p)
 					status = PCAP_ERROR;
 					goto bad;
 				}
-				strcpy(wltdev, "wlt");
```
guyharris, line 2020: The `malloc()` allocates enough space to hold the resulting string, and both `"wlt"` and `p->opt.device` are known to be `'\0'`-terminated, so the calls to `pcapint_strlpy()` aren't necessary - `strcpy()` and `strcat()` are safe.

The code should be rewritten to make it more clear that it's large enough for "wlt", followed by all but the first two characters of the device name, plus a terminating `'\0'`.

pcap-bpf.c
```diff
@@ -2800,8 +2800,8 @@ check_bpf_bindable(const char *name)
 			    errno, "malloc");
 			return (-1);
 		}
-		strcpy(en_name, "en");
```
guyharris, line 2803: This is similar, and the code should be changed to make it clearer.

(This is why programming languages, even "system programming languages", should at least have character strings as data types.)

sf-pcap.c
```diff
@@ -857,7 +857,7 @@ pcap_dump(u_char *user, const struct pcap_pkthdr *h, const u_char *sp)
 	 * attempted, but it is better than nothing.
 	 */
 	if (fwrite(&sf_hdr, sizeof(sf_hdr), 1, f) == 1) {
-		(void)fwrite(sp, h->caplen, 1, f);
+		fwrite(sp, h->caplen, 1, f);
```
guyharris, line 860: That cast makes no difference, other causing warnings to be reported.

The underlying problems that the `pcap_dump()` API does not return a value, and making it do so would mean that it couldn't be used as a callback for `pcap_dispatch()` or `pcap_loop()`, which would be a huge API breakage.

So, sadly, we don't support checking for write errors, and we can't fix the warnings by checking the return value of `fwrite()`. Thus, the cast should not be removed.

We will, at some point, have new APIs - in order to, for example, fully support reading pcapng files and support writing them - and they *will* allow detecting write errors.

infrastation:

PR #1729, opened 2026-08-21
gencode: Quiet a sanitizer warning in gen_ncode

fxlb:
Use a temporary 64-bit variable for the shift result.

The error was like:

$ testprogs/filtertest RAW 'net 10 mask 255'
gencode.c:7463:11: runtime error: left shift of 4294967295 by 8 places
  cannot be represented in type 'bpf_u_int32' (aka 'unsigned int')

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior gencode.c:7463:11

### DIFF
--- a/gencode.c
+++ b/gencode.c
@@ -7457,11 +7457,14 @@ gen_ncode(compiler_state_t *cstate, const char *s, bpf_u_int32 v, struct qual q)
 		} else {
 			mask = 0xffffffff;
 			if (s == NULL && q.addr == Q_NET) {
+				uint64_t mask64 = mask;
+
 				/* Promote short net number */
 				while (v && (v & 0xff000000) == 0) {
 					v <<= 8;
-					mask <<= 8;
+					mask64 <<= 8;
 				}
+				mask = (bpf_u_int32)mask64;
 			} else {
 				/* Promote short ipaddr */
 				v <<= 32 - vlen;


guyharris:
I'm not sure what the issue is.

C99 says, in 6.5.7 "Bitwise shift operators":

> The integer promotions are performed on each of the operands. The type of the result is
that of the promoted left operand. ...
> 
> ...
> 
> The result of **E1 << E2** is **E1** left-shifted **E2** bit positions; vacated bits are filled with
zeros. If **E1** has an unsigned type, the value of the result is **E1×(2^E2)**, reduced modulo
one more than the maximum value representable in the result type.

The "integer promotions" are the way C deals with types shorter than `int`/`unsigned int`, but those aren't involved here - the left-hand operand of the shift is `unsigned int` and the right-hand operand of the shift is `int`.

So "left shift of 4294967295 by 8 places" is 4294967295*256 = 1099511627520, and "one more than the maximum value representable in the result type" is 4294967296. 1099511627520 modulo 4294967296 is 4294967040, which is 0xFFFFFF00.

fxlb:
Thus it's more "Silent a Clang warning about a left shift" than "Fix an undefined behavior".
Thank you.

guyharris:
I'm trying to reproduce this to figure out what the issue is.

On what platform (OS, OS version, CPU type) was this done, and with what compiler and compiler version was this done?

Not seeing it with a fairly recent Clang on macOS Tahoe ARM64 and not seeing it on Ubuntu 24.04 x86-64 with either the builtin GCC or the builtin Clang.

(I *am* seeing a syntax error, although it looks as if it gets to the code that fails before the parser fails.)

fxlb:
It is on Debian Bookworm with x86-64 build by Clang 19.
```
$ ./configure CC=clang-19 CFLAGS='-fsanitize=address,pointer-overflow,pointer-compare,pointer-subtract,null,undefined,integer,signed-integer-overflow -fsanitize-recover=address,pointer-overflow,pointer-compare,pointer-subtract,null,undefined,integer,signed-integer-overflow -fno-omit-frame-pointer -fno-optimize-sibling-calls -g3 -ggdb3 -O0'
$ make check
$ testprogs/filtertest RAW 'net 10 mask 255'
```

guyharris:
OK, I upgraded my x86-64 Debian 12 VM to 12.15, installed clang-19, and ran your configure command and `make check`, and it got a lot of test failures.

I don't know whether the difference is the Clang version, the set of sanitize options, or the optimization level.

I'll try it with the latest Clang version that Debian package for bookworm.

guyharris:
> I'll try it with the latest Clang version that Debian package for bookworm.

That's clang-22, and that does *not* report any failures with `make check`, so it looks as if that's a Clang bug that was fixed in a later version.

guyharris:
> That's clang-22, and that does _not_ report any failures with `make check`, so it looks as if that's a Clang bug that was fixed in a later version.

No,, it looks as if I didn't run the configure script with the correct `CFLAGS` value. With the value, it does.

Google's AI claims that this is a result of `-fsanitize=integer`, in which case there's still a less-serious bug/misfeature, namely that it's incorrectly claiming the behavior is "undefined" as opposed to "you're going to lose bits, and I'm not sure that's intentional, so I'm letting you know".

I'll see if there's some way to convince the integer sanitizer that we know what we're doing, perhaps with a cast.

guyharris:
> Google's AI claims that this is a result of `-fsanitize=integer`,

Yup, removing that got rid of the warnings.

> I'll see if there's some way to convince the integer sanitizer that we know what we're doing, perhaps with a cast.

Casting doesn't seem to help. `__attribute__` might allow suppressing the "shift" sanitizer, either in specific lines of code or in specific functions.

infrastation:
I had a look at a number of UBSan errors a few days ago, the one that popped up most often went away using this change:
```diff
--- a/optimize.c
+++ b/optimize.c
@@ -689,7 +689,7 @@ F(opt_state_t *opt_state, int code, bpf_u_int32 v0, bpf_u_int32 v1)
        bpf_u_int32 val;
        struct valnode *p;
 
-       hash = (u_int)code ^ (v0 << 4) ^ (v1 << 8);
+       hash = (u_int)code ^ ((v0 & 0x0fffffff) << 4) ^ (v1 << 8);
        hash %= MODULUS;
 
        for (p = opt_state->hashtbl[hash]; p; p = p->next)
```
Did not commit because it looks wrong to complicate a busy code path that has already been working for decades, only to appease UBSan.

guyharris:
> I had a look at a number of UBSan errors a few days ago, the one that popped up most often went away using this change:

And that was sufficient - it didn't require you to mask `v1` with `0x00ffffff`? That's a bit surprising.

infrastation:
At the run time of `make check` on Linux on that line of that file only `v0` happens to have set bits that UBSan does not like. Unlike compiler warnings, for run-time checks upfront there is not even a complete list of issues to triage.

guyharris:
Putting `__attribute__((no_sanitize("unsigned-shift-base")))` in front of `optimize.c: F()` and `gencode.c:gen_ncode()` squelches these warnings and the similar warnings for `F()`.

I'm thinking of defining a local header file, e.g. "sanitizer-attrs.h", and have it include a `#define` called `UNSIGNED_SHIFT_OK` or something such as that, as appropriate for the compiler being used, and use that for the two functions.

infrastation:
If, for the sake of argument, the build systems verified that the C compiler produces the expected traditional behaviour for bit shifts, integer overflow, pointer overflow etc. before disabling UBSan around respective pieces of code, what would be the probability of actually running into a compiler that implements the standard exactly and fails the checks and thus cannot be used for compiling libpcap?

guyharris:
> If, for the sake of argument, the build systems verified that the C compiler produces the expected traditional behaviour for bit shifts, integer overflow, pointer overflow etc. before disabling UBSan around respective pieces of code, what would be the probability of actually running into a compiler that implements the standard exactly and fails the checks and thus cannot be used for compiling libpcap?

Depends on the issue.

For this *particular* issue, any compiler that implements the standard exactly will implement `<<` as shifting the upper bits into nothingness and not having an overflow error, as per https://github.com/the-tcpdump-group/libpcap/pull/1729#issuecomment-5374659055, so, for *this* issue, the technical term for a compiler that doesn't give the expected behavior is "non-standard-conforming".

infrastation:
Just as llvm/llvm-project#60712 goes about it.

guyharris:
> Just as [llvm/llvm-project#60712](https://github.com/llvm/llvm-project/issues/60712) goes about it.

See also llvm/llvm-project#218300.

guyharris:
> Putting `__attribute__((no_sanitize("unsigned-shift-base")))` in front of `optimize.c: F()` and `gencode.c:gen_ncode()` squelches these warnings and the similar warnings for `F()`.
> 
> I'm thinking of defining a local header file, e.g. "sanitizer-attrs.h", and have it include a `#define` called `UNSIGNED_SHIFT_OK` or something such as that, as appropriate for the compiler being used, and use that for the two functions.

See #1736

fxlb:
Fixed by #1736. Closed.

Issue #110, opened 2013-04-15
ineffective filter optimization


guyharris:
Converted from [SourceForge issue 1889037](http://sourceforge.net/support/tracker.php?aid=1889037), submitted by amotin

Using libpcap 0.9.8 I have found that it in some cases generates very ineffective code

For example:
```
%tcpdump -i lo0 -d 'src net 10.0.1.0/24 or src net 10.0.2.0/24'                 (000) ld       [0]
(001) jeq      #0x2000000       jt 2    jf 9
(002) ld       [16]
(003) and      #0xffffff00
(004) jeq      #0xa000100       jt 8    jf 5
(005) ld       [16]
(006) and      #0xffffff00
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(007) jeq      #0xa000200       jt 8    jf 9
(008) ret      #96
(009) ret      #0
```

In this case instructions 5 and 6 could be easily removed.

At the same time if expression becomes some more complicated optimizer is able to handle that:
tcpdump -i lo0 -d 'src port 25 and ( src net 10.0.1.0/24 or src net 10.0.2.0/24)'
```
(000) ld       [0]
(001) jeq      #0x1c000000      jt 17   jf 2
(002) jeq      #0x2000000       jt 3    jf 17
(003) ldb      [13]
(004) jeq      #0x84            jt 7    jf 5
(005) jeq      #0x6             jt 7    jf 6
(006) jeq      #0x11            jt 7    jf 17
(007) ldh      [10]
(008) jset     #0x1fff          jt 17   jf 9
(009) ldxb     4*([4]&0xf)
(010) ldh      [x + 4]
(011) jeq      #0x19            jt 12   jf 17
(012) ld       [16]
(013) and      #0xffffff00
(014) jeq      #0xa000100       jt 16   jf 15
(015) jeq      #0xa000200       jt 16   jf 17
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(016) ret      #96
(017) ret      #0
```

I think such optimization should significantly increase filter execution performance.

If needed I will be glad to help you as developer, but as optimizer code is not simple, could somebody give me any descripton of it's algorithm to save some of my investigation time?

infrastation:
The proposed patch:
```patch
--- optimize.c.prev	2008-02-09 02:11:54.000000000 +0200
+++ optimize.c	2008-02-09 02:11:44.000000000 +0200
[...]

PR #1732, opened 2026-08-25
optimizer clean-ups

infrastation:
These are various by-products of my yet another attempt to comprehend #110.

### DIFF
--- a/optimize.c
+++ b/optimize.c
@@ -33,6 +33,7 @@
 #include <limits.h> /* for SIZE_MAX */
 #include <errno.h>
 #include <stdbool.h>
+#include <stdint.h>
 
 #include "pcap-int.h"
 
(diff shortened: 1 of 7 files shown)


infrastation:
One type cast is missing.

guyharris:
> One type cast is missing.

Meaning "I'll add that as another commit to this pull request, don't merge it" or meaning "I'll do that later in another pull request?"

If the latter, is there any reason not to merge this now? (The AppVeyor failures are probably the result of AppVeyor's difficulty in making secure connections to sites from which it downloads various software packages for Windows; I'll do my usually prodding to get it through the test.)

optimize.c
```diff
@@ -2317,6 +2336,36 @@ mark_code(struct icode *ic)
 	make_marks(ic, ic->root);
 }
 
+// Return 1 iff opcode is valid and uses the 'k' field.
+bool
+pcapint_opcode_without_k(const uint16_t opcode)
+{
+	static const bool without_k[UINT8_MAX + 1] = {
+		[BPF_LD   | BPF_LEN         ] = 1,
```
guyharris: This is a `bool` array, so should the `= 1` be `= true`?

guyharris:
> > One type cast is missing.

...

> (The AppVeyor failures are probably the result of AppVeyor's difficulty in making secure connections to sites from which it downloads various software packages for Windows; I'll do my usually prodding to get it through the test.)

Nope, it's a real error:

```
C:\projects\libpcap\optimize.c(2386): warning C4242: 'function': conversion from 'int' to 'const uint16_t', possible loss of data [C:\projects\libpcap\build\pcap_static.vcxproj]
```

so is that the place where the type cast is missing? If so, I'm not sure where that conversion is being done.

optimize.c
infrastation: Exactly.

infrastation:
In `struct stmt` the opcode is `int`, so it can be set to `NOP` (-1) whilst still in the optimizer.

guyharris:
> In `struct stmt` the opcode is `int`, so it can be set to `NOP` (-1) whilst still in the optimizer.

So the conversion is done when passing it as an argument.

Is it known not to be `NOP` in the call on line 2386?

infrastation:
Yes, this revision includes a comment to clarify that. Thank you for having a look.

infrastation:
This is going to be merged in a few days.

Issue #1702, opened 2026-07-16
RFC: a differentially-certified memory-safe Rust port of the pcap savefile parser — of interest?


anthonyadame:
Hi,

I'd like to offer a memory-safe Rust port of libpcap 1.10.5's **classic-pcap savefile parser**
— `pcap_check_header` and `pcap_next_packet` from `sf-pcap.c`, the code that runs first on an
untrusted `.pcap` file — and ask whether it's a direction libpcap would want to explore.

**What exists today**

- A zero-dependency, `#![forbid(unsafe_code)]` Rust crate covering the classic-pcap file-header
  and packet-record parser, ported from the pinned 1.10.5 tarball and BSD-3-Clause identical to
  upstream.
- A byte-differential certification harness: the upstream C parser (sliced **verbatim** from
  your `sf-pcap.c` and compiled with a thin shim + a tmpfile-backed `FILE*`) and the Rust port
  are driven by the same op-script driver over a 217-case envelope, and their complete stdout —
  parsed header fields, per-packet caplen/len/timestamp/data-hash, **and the verbatim
  diagnostic string on every rejection** — is byte-identical. The envelope is adversarial:
  every version and magic variant, both byte orders, snaplen edges, the caplen/len swap
  heuristics, the caplen>max and caplen>snapshot bounds checks, single-byte corruption at every
  header offset, truncation at every field boundary, and pure-garbage fuzz.
- The golden outputs regenerate from your C build, so the claim is independently checkable —
  and because the trace includes your exact error strings, it certifies the diagnostics too,
  not just accept/reject.

**Why this parser specifically**

The savefile header/record path is the untrusted-input front door, and it's had its share of
CVE-class bugs (malformed caplen/snaplen handling). This port reproduces the C parser's exact
accept/reject/bounds behavior in safe Rust — a rejected or fuzzed file is a clean `Err`, never
an out-of-bounds read.

**Disclosure on how it was made**

The port was produced by an automated, LLM-assisted translation pipeline whose acceptance gate
is the differential certification above — not human review of the generated code. Stating that
plainly. The trust anchor is the byte-exact behavioral match against your reference over an
adversarial envelope, plus the compiler-enforced absence of `unsafe`. Scope is honest: this is
the classic-pcap savefile *parser* only — not live capture, not pcapng, not the encoder.

**What I'm asking**

1. Is a Rust savefile parser something libpcap would want (as a sibling crate, an optional
[...]

PR #1734, opened 2026-08-29
pcapng: decode file fields through checked cursors

QuantumStonks:
## Summary

- treat serialized pcapng fields as immutable byte storage
- decode every fixed SHB, IDB, EPB, SPB, PB, and option-header field through
  one bounds-checked cursor using alignment-safe scalar loads
- use explicit pcapng wire sizes instead of C structure sizes, while preserving
  the existing public ABI and ordinary C99 build

## Rationale

`sf-pcapng.c` currently casts offsets in the file buffer to several C structure
types, mutates those structures in place when byte swapping, and uses their
compiler layout as the serialized wire layout. The existing cursor checks the
size before each cast, but the representation still depends on alignment,
effective-type, constness, and ABI-layout assumptions.

This change separates wire bytes from decoded C values. A fixed record is first
claimed as an exact-size child cursor, then decoded one scalar at a time with
`memcpy()` and the existing byte-swap helpers. The parent cursor remains the
single owner of pointer and remaining-length state. Packet bytes remain
zero-copy; the sole mutable escape occurs only after the captured-length slice
has passed the cursor check, as required by `pcapint_post_process()`.

This is a focused C implementation of the bounds-checked field-access direction
discussed in #1702. It adds no toolchain, generated source, or link requirement.

## Validation

- Clang ASan+UBSan full suite: 21,209 total, 14,640 passed, 6,569 skipped,
  0 failed
- fully instrumented libFuzzer: 885,020 executions across two workers,
  560 corpus entries, no crash or sanitizer finding
- base-vs-patch differential: identical exit status and byte-for-byte output on
  all 560 generated corpus inputs
- deterministic 200,000-run throughput comparison: 2,857 exec/s for both base
  and patch
- GCC 13.3 Release shared/static build: all 56 targets built
- native big-endian s390x GCC 13.3 build and execution: paired BE/LE valid
  fixtures produced identical four-packet output; paired invalid fixtures were
  rejected identically
- Clang `-Wall -Wextra -Werror -fsyntax-only`: clean

### DIFF
--- a/sf-pcapng.c
+++ b/sf-pcapng.c
@@ -46,20 +46,10 @@
  * Block types.
  */
 
-/*
- * Common part at the beginning of all blocks.
- */
-struct block_header {
-	bpf_u_int32	block_type;
-	bpf_u_int32	total_length;
-};
-
-/*
- * Common trailer at the end of all blocks.
- */
-struct block_trailer {
-	bpf_u_int32	total_length;
-};
+/* Sizes of serialized pcapng fields; these are format properties, not C ABIs. */
+#define BLOCK_HEADER_LENGTH	8U
+#define BLOCK_TRAILER_LENGTH	4U
+#define OPTION_HEADER_LENGTH	4U
 
 /*
  * Common options.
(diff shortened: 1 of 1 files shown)


infrastation:

PR #1737, opened 2026-08-30
Fix building libpcap with CMake without libnl-genl-3-dev

### LINKED ISSUE #1703: Error building libpcap with CMake without libnl-genl-3-dev

fxlb:
libpcap now require libnl-genl-3 only because the latter requires libnl-3.

The error was:
```
$ MATRIX_CC=gcc MATRIX_CMAKE=yes MATRIX_REMOTE=no MATRIX_PROTOCHAIN=no \
  ./build_matrix.sh
...
/usr/bin/ld: cannot find -lnl-genl-3: No such file or directory 
collect2: error: ld returned 1 exit status
[ 61%] Building C object CMakeFiles/pcap_static.dir/sf-pcapng.c.o 
make[2]: *** [CMakeFiles/pcap.dir/build.make:530: libpcap.so.1.11.0-PRE-GIT] Error 1 
make[1]: *** [CMakeFiles/Makefile2:166: CMakeFiles/pcap.dir/all] Error 2
```


fxlb:
Should fix #1703.

infrastation:
Then Autoconf should do the same?

infrastation:
And the function being tested should be from libnl-genl-3 rather than libnl-3?

fxlb:
> And the function being tested should be from libnl-genl-3 rather than libnl-3?

Yes, done.

infrastation:
The Autoconf side still makes different checks.

Also I vaguely remember a class of linking difficulties that is a result of some linkers correctly guessing indirect dependencies and other linkers requiring an explicit list of all dependencies, in a particular order. Guy is familiar with this problem space much better.

fxlb:
> The Autoconf side still makes different checks.
> 
> Also I vaguely remember a class of linking difficulties that is a result of some linkers correctly guessing indirect dependencies and other linkers requiring an explicit list of all dependencies, in a particular order. Guy is familiar with this problem space much better.

I missed that comment. This merge will probably need a follow-up...

PR #1743, opened 2026-09-08
netmap: count only packets that pass the filter.

### LINKED ISSUE #1733: netmap module counts received packets before filtering
In my setup this reproduced using a cross-compiled build of tcpdump+libpcap for QNX, but it is not specific to QNX and should be reproducible the same on FreeBSD and Linux.

One way to reproduce the problem is to use an Ethernet segment with a few hosts and on one of the hosts to replay the `pppoes.pcap` file from libpcap tests in a loop. Since the packets are unknown unicast, every host on the segment will be receiving a few PPPoE session packets per second. If the user needs to capture these packets only, and only one packet, they would run tcpdump in a manner like this:
```
# tcpdump -ni eth0 -c 1 pppoes
listening on eth0, link-type EN10MB (Ethernet), snapshot length 262144 bytes
14:18:50.104602 PPPoE  [ses 0x3b] LCP, Echo-Request (0x09), id 103, length 14
1 packet captured
2 packets received by filter
0 packets dropped by kernel
```
The above represents a Linux host with a conventional Ethernet network interface, and in this case the result is exactly as expected. On a QNX host with a cross-compiled build (to enable netmap) this produces a different result:
```
# tcpdump -ni netmap:cgem0 -c 1 pppoes
listening on netmap:cgem0, link-type EN10MB (Ethernet), snapshot length 262144 bytes
0 packets captured
1 packet received by filter
0 packets dropped by kernel
```
Using a higher packet limit, it is easier to see what is wrong:
```
# tcpdump -ni netmap:cgem0 -c 10 pppoes
listening on netmap:cgem0, link-type EN10MB (Ethernet), snapshot length 262144 bytes
13:23:30.428575 PPPoE  [ses 0x3b] LCP, Echo-Request (0x09), id 103, length 14
13:23:30.428664 PPPoE  [ses 0x17] LCP, Echo-Request (0x09), id 106, length 14
13:23:31.196596 PPPoE  [ses 0x3b] LCP, Echo-Request (0x09), id 103, length 14
13:23:31.196658 PPPoE  [ses 0x17] LCP, Echo-Request (0x09), id 106, length 14
4 packets captured
10 packets received by filter
0 packets dropped by kernel
```
The root cause of this seems to be this code in `pcap_netmap_dispatch()`:
```c
                ret = nm_dispatch((void *)d, cnt, (void *)pcap_netmap_filter, (void *)p);
                if (ret != 0)
                        break;
```
Even though in `pcap_netmap_filter()` the libpcap callback function invocation is conditional on the result of userland filtering, netmap invokes its callback function unconditionally and increments the packet counter every time, so every packet received by netmap ends up counted as received by libpcap user regardless if it matched the filter.

guyharris:
I'm not sure why they deliberately chose to count packets received *before* the filter, but most if not all capture mechanisms report only packets that passed the filter.

Fix #1733.

### DIFF
--- a/pcap-netmap.c
+++ b/pcap-netmap.c
@@ -64,7 +64,7 @@ struct pcap_netmap {
 	pcap_handler cb;	/* callback and argument */
 	u_char *cb_arg;
 	int must_clear_promisc;	/* flag */
-	uint64_t rx_pkts;	/* # of pkts received before the filter */
+	uint64_t rx_pkts;	/* # of pkts that passed the filter */
 };
 
 
@@ -88,7 +88,6 @@ pcap_netmap_filter(u_char *arg, struct pcap_pkthdr *h, const u_char *buf)
 	const struct bpf_insn *pc = p->fcode.bf_insns;
 	u_int snaplen = (u_int)p->snapshot; /* guaranteed not to be negative */
 
-	++pn->rx_pkts;
 	if (pc == NULL ||
 	    (snaplen = pcapint_filter(pc, p->fcode.bf_len,
 	                              buf, h->len, h->caplen)) != 0) {
@@ -97,6 +96,7 @@ pcap_netmap_filter(u_char *arg, struct pcap_pkthdr *h, const u_char *buf)
 		 */
 		if (h->caplen > snaplen)
 			h->caplen = snaplen;
+		++pn->rx_pkts;
 		pn->cb(pn->cb_arg, h, buf);
 	}
 }


infrastation:
This revision makes the received packet counter consistent with the number of packets received by the user callback, but the latter remains incorrect. Tested in a virtual Ethernet segment of the following configuration:
* host0 10.20.230.240/24
* host3 10.20.230.243/24
* host4 10.20.230.244/24

host0 and host3 are running `ping -n 10.20.230.244` and host 4 (MidnightBSD) is capturing. Because netmap disconnects network interfaces from the usual network stack, there is another virtual Ethernet segment (not shown) for SSH access.

Not using netmap:
```
# ./tcpdump -pni em1 -c 10 'icmp and src host 10.20.230.243'
listening on em1, link-type EN10MB (Ethernet), snapshot length 262144 bytes
09:53:44.266037 IP 10.20.230.243 > 10.20.230.244: ICMP echo request, id 2, seq 819, length 64
09:53:45.290321 IP 10.20.230.243 > 10.20.230.244: ICMP echo request, id 2, seq 820, length 64
09:53:46.314927 IP 10.20.230.243 > 10.20.230.244: ICMP echo request, id 2, seq 821, length 64
09:53:47.339301 IP 10.20.230.243 > 10.20.230.244: ICMP echo request, id 2, seq 822, length 64
09:53:48.363701 IP 10.20.230.243 > 10.20.230.244: ICMP echo request, id 2, seq 823, length 64
09:53:49.388727 IP 10.20.230.243 > 10.20.230.244: ICMP echo request, id 2, seq 824, length 64
09:53:50.412916 IP 10.20.230.243 > 10.20.230.244: ICMP echo request, id 2, seq 825, length 64
09:53:51.437708 IP 10.20.230.243 > 10.20.230.244: ICMP echo request, id 2, seq 826, length 64
09:53:52.461854 IP 10.20.230.243 > 10.20.230.244: ICMP echo request, id 2, seq 827, length 64
09:53:53.486499 IP 10.20.230.243 > 10.20.230.244: ICMP echo request, id 2, seq 828, length 64
10 packets captured
20 packets received by filter
0 packets dropped by kernel
```

Using netmap, without this change:
```
# ./tcpdump -pni netmap:em1 -c 10 'icmp and src host 10.20.230.243'
listening on netmap:em1, link-type EN10MB (Ethernet), snapshot length 262144 bytes
09:55:13.398078 IP 10.20.230.243 > 10.20.230.244: ICMP echo request, id 2, seq 906, length 64
09:55:14.422276 IP 10.20.230.243 > 10.20.230.244: ICMP echo request, id 2, seq 907, length 64
2 packets captured
10 packets received by filter
0 packets dropped by kernel
```

Using netmap, with this change:
```
# ./tcpdump -pni netmap:em1 -c 10 'icmp and src host 10.20.230.243'
listening on netmap:em1, link-type EN10MB (Ethernet), snapshot length 262144 bytes
09:56:08.720628 IP 10.20.230.243 > 10.20.230.244: ICMP echo request, id 2, seq 960, length 64
09:56:09.745972 IP 10.20.230.243 > 10.20.230.244: ICMP echo request, id 2, seq 961, length 64
09:56:10.769621 IP 10.20.230.243 > 10.20.230.244: ICMP echo request, id 2, seq 962, length 64
09:56:11.794555 IP 10.20.230.243 > 10.20.230.244: ICMP echo request, id 2, seq 963, length 64
09:56:12.818929 IP 10.20.230.243 > 10.20.230.244: ICMP echo request, id 2, seq 964, length 64
5 packets captured
5 packets received by filter
0 packets dropped by kernel
```

guyharris:
My inclination is *not* to commit this for 1.11.0; the problem is that both the current behavior and the behavior with this change match the behavior of at least one significant platform (existing matches \*BSD/Darwin/Solaris behavior, behavior with this change matches Linux/Windows with Npcap).

This needs more thought. The problem isn't helped by platform limitations make it difficult if not impossible to change either of the two behaviors on platforms that provide the other behavior. I'm also not sure which behavior is more useful.

infrastation:
This is not a recent regression: the counter has been broken since the initial implementation in 2017 and nobody complained.

guyharris:
> This is not a recent regression: the counter has been broken since the initial implementation in 2017 and nobody complained.

More correctly, the counter has behaved like the BPF counter rather than the Linux/NPF counter since the initial implementation in 2017.

What's broken is that various kernel capture mechanisms behave differently; if that's somebody's bug, I'd call it a Linux bug, as they ended up implementing a different behavior.

PR #1744, opened 2026-09-10
pcap_fileno, pcap_get_selectable_fd man pages: various fixes.

### LINKED ISSUE #1722: Clarify use of file descriptors returned by pcap_fileno() and pcap_get_selectable_fd()
It is not clear what operations an application can safely perform on the file descriptors returned by these two functions. In particular, whether the application is responsible for closing them when done.

A [mailing list discussion](https://seclists.org/tcpdump/2026/q3/9) clarified that an application should _only_ use these fds for detection of read/write events as returned by select/poll/kqueue et al and can then call pcap_loop() and pcap_dispatch() without blocking (within the bounds of what the OS supports). Furthermore, any other operations on the fd, including close() are prohibited and result in undefined (and likely catastrophic) behaviour within pcap.

It is suggested that the manpages be amended to include this clarification, perhaps along the lines of:

"The returned file descriptor can only be used for detecting readable/writable state as returned by select(2)/poll(2)/kqueue(2) and similar. No other operations should be performed on the file descriptor including close(2)."

guyharris:
Fix #1722 by warning that you must *not* close either of those file descriptors out from under libpcap.

Add the "if p refers to..." clause from the pcap_fileno page to the pcap_get_selectable_fd page.

Make some other improvements to the pcap_get_selectable_fd page.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -87,6 +87,9 @@ DayOfTheWeek, Month DD, YYYY / The Tcpdump Group
       Produce cleaner DOT in the optimizer debugging code.
       man: Use "null pointer" rather than "NULL" and fix some other
         issues.
+      Note in pcap_fileno(3PCAP) and pcap_get_selectable_fd(3PCAP) that
+        the descriptors returned by those routinesmust not be closed.
+        (Issue #1722)
     Building and testing:
       Apply GNU Hurd support patch from the Debian package.
       CI: Introduce and use LIBPCAP_CMAKE_TAINTED.
--- a/pcap_fileno.3pcap
+++ b/pcap_fileno.3pcap
@@ -17,7 +17,7 @@
 .\" WARRANTIES, INCLUDING, WITHOUT LIMITATION, THE IMPLIED WARRANTIES OF
 .\" MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE.
 .\"
-.TH PCAP_FILENO 3PCAP "25 July 2018"
+.TH PCAP_FILENO 3PCAP "10 September 2026"
 .SH NAME
 pcap_fileno \- get the file descriptor for a live capture
 .SH SYNOPSIS
@@ -33,7 +33,7 @@ int pcap_fileno(pcap_t *p);
 .SH DESCRIPTION
 If
 .I p
-refers to a network device that was opened for a live capture using
+refers to a capture device that was opened for a live capture using
 a combination of
 .BR pcap_create (3PCAP)
 and
@@ -42,6 +42,10 @@ or using
 .BR pcap_open_live (3PCAP),
 .BR pcap_fileno ()
 returns the file descriptor from which captured packets are read.
+.LP
+A
+.BR close (2)
+must not be performed on the file descriptor.
 .LP
 If
 .I p
--- a/pcap_get_selectable_fd.3pcap
+++ b/pcap_get_selectable_fd.3pcap
@@ -17,7 +17,7 @@
 .\" WARRANTIES, INCLUDING, WITHOUT LIMITATION, THE IMPLIED WARRANTIES OF
 .\" MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE.
 .\"
-.TH PCAP_GET_SELECTABLE_FD 3PCAP "6 September 2026"
+.TH PCAP_GET_SELECTABLE_FD 3PCAP "10 September 2026"
 .SH NAME
 pcap_get_selectable_fd \- get a file descriptor on which a select() can
 be done for a live capture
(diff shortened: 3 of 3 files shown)


infrastation:
These changes look substantial enough to bump the `.TH` date.

guyharris:
> These changes look substantial enough to bump the `.TH` date.

Forgot that. Thanks. Fixed.

cicku:
Hmm...`pcap_fileno` man page says:

> If *p* refers to a ``savefile`` [...], a "dead'' `pcap_t` opened using `pcap_open_dead(3PCAP)`, or a `pcap_t` that was created with `pcap_create()` but that has not yet been activated with `pcap_activate()`, it returns `PCAP_ERROR`.

Currently on master, a dead handle returns 0 instead. From both `pcap_fileno()` and `pcap_get_selectable_fd()`.

`pcap_open_dead_with_tstamp_precision()` builds its handle with `malloc()` + `memset(p, 0, sizeof(*p))` rather than going through `pcap_alloc_pcap_t()`, and `pcap_alloc_pcap_t()` is what sets `p->fd = -1` and `p->selectable_fd = -1`. On UN*X `pcap_fileno()` is `return (p->fd);` with no error path of its own, so the zeroed field comes straight back.

Should the page describe what it does, or should `pcap_open_dead()` set the two fields?

guyharris:
> Should the page describe what it does, or should `pcap_open_dead()` set the two fields?

Option 2 - set the two fields.

infrastation:
This uncovers another confusion. pcap_fileno(3PCAP) and pcap_get_selectable_fd(3PCAP) used to say the functions return -1 until my commit 6e15abc in 2018 changed both to `PCAP_ERROR`. Guy's commit a973128 in 2020 changed pcap_get_selectable_fd(3PCAP) to say -1 again, but pcap_fileno(3PCAP) still says `PCAP_ERROR`. Perhaps the latter man page should say -1 again too to reduce the chance of being "improved" by accident to `PCAP_ERROR_NOT_ACTIVATED` in the "not yet opened" case, as was done elsewhere in commits 77bd024, d842000 and fdb5e7c earlier.

Also it would be nice to have a number of tests in the C code space (`testprogs/opentest.c` seems appropriate) to verify that various functions return the expected values for dead and not-yet-activated handles.

Anyway, this is not a recent regression and there is no need for an urgent fix.

guyharris:
> This uncovers another confusion. pcap_fileno(3PCAP) and pcap_get_selectable_fd(3PCAP) used to say the functions return -1 until my commit [6e15abc](https://github.com/the-tcpdump-group/libpcap/commit/6e15abcfff9504c9c26c249db56179c4c26c1b16) in 2018 changed both to `PCAP_ERROR`. Guy's commit [a973128](https://github.com/the-tcpdump-group/libpcap/commit/a973128a85d7dd75c7ea6fdcf746fc143a987d03) in 2020 changed pcap_get_selectable_fd(3PCAP) to say -1 again, but pcap_fileno(3PCAP) still says `PCAP_ERROR`. Perhaps the latter man page should say -1 again too to reduce the chance of being "improved" by accident to `PCAP_ERROR_NOT_ACTIVATED` in the "not yet opened" case, as was done elsewhere in commits [77bd024](https://github.com/the-tcpdump-group/libpcap/commit/77bd02481ae819edc48f24a9f44b32bb20f54a47), [d842000](https://github.com/the-tcpdump-group/libpcap/commit/d84200082656227a82e1996b148c814e1da0dd41) and [fdb5e7c](https://github.com/the-tcpdump-group/libpcap/commit/fdb5e7c141d4d444cd920fa4889d17f2f921aff8) earlier.

Yes, the rationale for changing to -1 was

> Update the documentation to reflect this, and make various other cleanups (including documenting the error return value for pcap_get_selectable_fd() to -1 rather than PCAP_ERROR - it's not an error code, it's just a specific error value).

My inclination is to do that for pcap_fileno(3PCAP) as well.

This is the problem with APIs that don't provide separate return values and success/failure indications.  I *guess* we could define them to return either valid file descriptors or `PCAP_ERROR_` values, but, if they *do* return a value other than `PCAP_ERROR`, that might break code that checks only for -1 or for `PCAP_ERROR`.

(These days, I think of programming languages that failure should not be an option, it should be a built-in feature of the language. :-) I.e., the language should allow a function that can fail to perform the requested function to report the failure in a fashion independent of the return value. But here we are.)

So any reason not to commit this now for 1.11.1?

infrastation:
Looks about right to me.

guyharris:
> Hmm...`pcap_fileno` man page says:
> 
> > If _p_ refers to a `savefile` [...], a "dead'' `pcap_t` opened using `pcap_open_dead(3PCAP)`, or a `pcap_t` that was created with `pcap_create()` but that has not yet been activated with `pcap_activate()`, it returns `PCAP_ERROR`.
> 
> Currently on master, a dead handle returns 0 instead. From both `pcap_fileno()` and `pcap_get_selectable_fd()`.
> 
> `pcap_open_dead_with_tstamp_precision()` builds its handle with `malloc()` + `memset(p, 0, sizeof(*p))` rather than going through `pcap_alloc_pcap_t()`, and `pcap_alloc_pcap_t()` is what sets `p->fd = -1` and `p->selectable_fd = -1`. On UN*X `pcap_fileno()` is `return (p->fd);` with no error path of its own, so the zeroed field comes straight back.
> 
> Should the page describe what it does, or should `pcap_open_dead()` set the two fields?

Setting the fields in #1749.

PR #1746, opened 2026-09-12
rpcap: Allocate a sockaddr_storage for deserialized addresses.

### LINKED ISSUE #1738: pcap-rpcap.c:340 [-Walloc-size] from Clang 22
```
pcap-rpcap.c:340:40: warning: allocation of insufficient size '28' for type 'struct sockaddr' with size '32' [-Walloc-size]
  340 |                 (*sockaddrout) = (struct sockaddr *) malloc(sizeof(struct sockaddr_in6));
      |                                                      ^

> uname -srvm
Haiku R1~beta6 hrev59866+79 Aug 25 2026 03:02: x86_64

> clang --version
clang version 22.1.8
Target: x86_64-unknown-haiku
Thread model: posix
InstalledDir: /boot/system/bin
```
The Haiku R1/beta6 installation ISO can be downloaded from [here](https://www.haiku-os.org/get-haiku/r1beta6/), after the default installation the following command is sufficient to run a complete build matrix:
```
pkgman install cmake llvm22_clang
```
The two structures are as follows:
```
/system/develop/headers/posix/sys/socket.h
struct sockaddr {
        uint8_t         sa_len;
        sa_family_t     sa_family;
        uint8_t         sa_data[30];
};
sizeof == 32

/system/develop/headers/posix/netinet6/in6.h
struct sockaddr_in6 {
        uint8_t         sin6_len;
        uint8_t         sin6_family;
        uint16_t        sin6_port;
        uint32_t        sin6_flowinfo;
        struct in6_addr sin6_addr;
        uint32_t        sin6_scope_id;
};
sizeof == 28
```

JohannesGezachew:
Fixes #1738.

`rpcap_deseraddr()` allocates `sizeof(struct sockaddr_in)` or `sizeof(struct sockaddr_in6)` and returns the result through a `struct sockaddr *`. Where `struct sockaddr` is larger than the address structure allocated, the object is smaller than the type it is used as. On Haiku (`struct sockaddr` 32 bytes, `struct sockaddr_in6` 28) Clang 22 reports:

```
pcap-rpcap.c:340:35: warning: allocation of insufficient size '28' for type 'struct sockaddr' with size '32' [-Walloc-size]
```

This allocates a `struct sockaddr_storage` in both branches. It is defined to be large enough for any supported address family, and libpcap already uses it for this purpose in `sockutils.c`, `fad-getad.c`, `rpcapd/daemon.c` and others. The IPv4 branch has the same pattern as the IPv6 one the warning names, so both are changed.

The callers (`pcap_findalldevs_ex()` in the same file) only store these pointers in a `struct pcap_addr` and later free them, so nothing depends on the allocation size.

### Testing

Built on macOS 26 (arm64, Apple clang 21) with `cmake -DENABLE_REMOTE=ON`: builds clean, with no new warnings.

I could not reproduce the original diagnostic locally: Apple clang 21 does not support `-Walloc-size`, and on macOS `struct sockaddr` is 16 bytes (`sockaddr_in` 16, `sockaddr_in6` 28, `sockaddr_storage` 128), so an allocation is never smaller than `struct sockaddr`. The change is therefore argued from the sizes rather than from a reproduced warning.

### DIFF
--- a/pcap-rpcap.c
+++ b/pcap-rpcap.c
@@ -308,7 +308,7 @@ rpcap_deseraddr(struct rpcap_sockaddr *sockaddrin, struct sockaddr **sockaddrout
 		struct rpcap_sockaddr_in *sockaddrin_ipv4;
 		struct sockaddr_in *sockaddrout_ipv4;
 
-		(*sockaddrout) = (struct sockaddr *) malloc(sizeof(struct sockaddr_in));
+		(*sockaddrout) = (struct sockaddr *) malloc(sizeof(struct sockaddr_storage));
 		if ((*sockaddrout) == NULL)
 		{
 			pcapint_fmt_errmsg_for_errno(errbuf, PCAP_ERRBUF_SIZE,
@@ -337,7 +337,7 @@ rpcap_deseraddr(struct rpcap_sockaddr *sockaddrin, struct sockaddr **sockaddrout
 		struct rpcap_sockaddr_in6 *sockaddrin_ipv6;
 		struct sockaddr_in6 *sockaddrout_ipv6;
 
-		(*sockaddrout) = (struct sockaddr *) malloc(sizeof(struct sockaddr_in6));
+		(*sockaddrout) = (struct sockaddr *) malloc(sizeof(struct sockaddr_storage));
 		if ((*sockaddrout) == NULL)
 		{
 			pcapint_fmt_errmsg_for_errno(errbuf, PCAP_ERRBUF_SIZE,


infrastation:
This change silences the warning on Haiku, but I did not check if it needs to be accompanied by any other changes to be correct.

infrastation:
To reproduce the break/fix locally, it is sufficient to have a hypervisor with a graphics card (VirtualBox works), to [download](https://www.haiku-os.org/get-haiku/r1beta6/) and to [install](https://www.haiku-os.org/guides/installing) Haiku, then to follow libpcap-specific steps in `doc/README.haiku.md`, and to run:
```
CC=clang-22 ./configure --enable-remote
make -s
```

infrastation:
Regardless of whether this change is ready in the memory management sense, if it addresses the warning, it should also remove the lines that commit 8d6fce2 added to build.sh.

guyharris:
Yeah, the problem is that, back in 4.2BSD, `struct sockaddr` was chosen to be big enough for everything they were expecting to use. The modification date on my copy of the 4.2BSD source is 1983-07-28; this predates both IPv6 and the OSI stack work at Berkeley, so it's not big enough for all the stuff they ended up using. 4.2BSD had `sockaddr_XXX` structures for IPv4 and... [PUP](https://en.wikipedia.org/wiki/PARC_Universal_Packet). I think it just used `struct sockaddr` for UNIX-domain sockets. I guess that had 14 bytes of data because 1) that was "big enough" and 2) it meant that the structure was 16 bytes long, making it a power of 2.

So we have an annoying mess wherein the "generic" socket address type is `struct sockaddr`, pointers to which are passed to various networking APIs, and the "big enough to hold an arbitrary socket address" socket address type is `struct sockaddr_storage`, and we're somewhat stuck with that.

Haiku, I guess, decided to double the size of `struct sockaddr` to 32 bytes, increasing the data length to 30 bytes; that made it bigger than a `struct sockaddr_in`.

And here we are.

I *suspect* that, had `struct sockaddr` been defined to have the last element be a flexible array element, this wouldn't be an issue, but 1) it was turning into a Project to try to test that on Haiku, 2) they didn't *have* flexible array elements at the time, and 3) this is all a "what if".

So I guess this is probably the best fix. I don't think it requires other changes; it does allocate more memory than is needed, but we can probably live with that.

infrastation:
I am going to have another look tomorrow.

infrastation:

@guyharris, you understand the problem space much better, please fix this when how and if you see fit.

Issue #1509, opened 2025-04-22
at the build time verify that atexit() does not crash


infrastation:
This is a by-product of #1507. Since it is a corner case and the bug seems to be not in this code base, it should be fine to keep it unresolved for a while.

When using TinyCC version 0.9.27 or a recent development snapshot, the `atexit()` function causes a segmentation fault on 32-bit ARM platforms ([steps to reproduce](https://lists.nongnu.org/archive/html/tinycc-devel/2025-04/msg00003.html)). The only code that uses `atexit()` directly or indirectly is in the following files:
* `pcap-dag.c` (which does not apply to 32-bit ARM)
* `pcap-bpf.c` if `BIOCGDLTLIST` is defined (which is the case on NetBSD)
* `pcap-linux.c` if netlink library is used for 802.11 monitor mode

A crude workaround could be to have `pcapint_do_addexit()` always fail if `__TINYC__` is defined. Alternatively, Autoconf and CMake could test whether `atexit()` works as expected and if that is not the case, error the build or avoid defining `HAVE_ATEXIT`, which `pcapint_do_addexit()` would normally expect to be defined.

In the latter case a simple way to test the workaround would be using Raspberry Pi Zero, Zero W or 2B rev 1.1 (all are 32-bit with an easy way to install Linux, Zero W has an built-in 802.11 interface, but it does not support monitor mode as discussed in #1041, so in any case it would have to be an external 802.11 adapter).

infrastation:
One potential way to remove `atexit()` from libpcap scope could be registering all pending clean-up work in the `pcap_t` and doing it in `pcap_close()`, then it would be a responsibility of the application to call `pcap_close()` no matter which exit route.

infrastation:
If this problem cannot be addressed by means of a new test that runs `pcapint_do_addexit()`, it will need build-time checks.

infrastation:
Fixed, closing.

PR #1748, opened 2026-09-16
tests, bug fixes and a man page update

infrastation:
This includes a workaround for #1509 among other things.


infrastation:
```
bpf_filter.c:582:47: error: pointer comparison always evaluates to false [-Werror,-Wtautological-compare]
  582 |         if (len < 1 || len > BPF_MAXINSNS || f + len < f)
      |                                                      ^
```
Looks like a pointer overflow is no longer a concept in Clang 21.1.8...

infrastation:
The same warning from Clang 20, so this revision does not include the commits with the offending code clean-ups.

Issue #1730, opened 2026-08-23
bpf_filter: Unsigned integer overflow


fxlb:
```
./configure CC=clang-22 CFLAGS='-fsanitize=unsigned-integer-overflow -fsanitize-recover=unsigned-integer-overflow -fno-omit-frame-pointer -fno-optimize-sibling-calls -g3 -ggdb3 -O0'
(Debian Bookworm, x86_64)

One example:

$ perl ./testprogs/TESTrun --one apply_ip6_protochain_108_tunnel_opt

INFO: Using a test timeout of 5.
INFO: This Perl supports threads, using 4 tester thread(s).
INFO: S = skipped, . = passed, F = failed, T = timed out            
F                                                                   1 / 1 (100%)

Failed tests:
    apply_ip6_protochain_108_tunnel_opt     : diff error
--- /dev/shm/libpcap_TESTrun_m13UJ0Xf/job000-expected.txt       2026-08-23 13:23:31.697757871 +0200
+++ /dev/shm/libpcap_TESTrun_m13UJ0Xf/job000-stdout.txt 2026-08-23 13:23:31.841760980 +0200
@@ -1,3 +1,12 @@
+bpf_filter.c:496:14: runtime error: unsigned integer overflow: 19 + 4294967281 cannot be represented in type 'u_int' (aka 'unsigned int')
+    #0 0x5642266c2292 in pcapint_validate_filter .../libpcap/bpf_filter.c:496:14
+    #1 0x5642266c240a in bpf_validate .../libpcap/bpf_filter.c:534:9
+    #2 0x564226699d99 in main .../libpcap/testprogs/./filtertest.c:484:7
+    #3 0x7fc5549e5249 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
+    #4 0x7fc5549e5304 in __libc_start_main csu/../csu/libc-start.c:360:3
+    #5 0x56422666cbb0 in _start (.../libpcap/testprogs/filtertest+0x10bb0) (BuildId: e009bc05c5cdbd35346991806d06966473859ac0)
+
+SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior bpf_filter.c:496:14
 0
 0
 65535

------------------------------------------------
0 tests skipped
1 tests failed
0 tests passed
```

infrastation:
On the current master branch using Clang 19.1.7:
```
[...]

PR #1751, opened 2026-09-20
Avoid unsigned overflow when validating BPF_JA targets

Uchio7:
## Summary

Fixes the unsigned-integer-overflow reported in #1730 while validating `BPF_JA` jump targets.

- calculate the `BPF_JA` target using a signed 32-bit offset in a 64-bit temporary
- reject targets outside the filter program without relying on unsigned addition wraparound
- preserve libpcap's signed interpretation for backward `protochain` jumps
- add a changelog entry

## Testing

I could not run the full libpcap test suite in my current Windows environment because no C compiler is installed (`gcc`, `clang`, and MSVC `cl` are unavailable).

Static check performed:

```text
git diff --check
```

Assisted-by: OpenAI Codex


infrastation:

PR #1752, opened 2026-09-21
Prevent optimizer stack exhaustion on deeply nested CFGs

### LINKED ISSUE #27: stack overflow

tokenjunkielabs:
Fixes #27.

Add an iterative CFG-depth preflight before the optimizer's recursive graph walkers. Pathologically deep control-flow graphs are rejected before they can exhaust the process stack; shared-DAG revisits retain the greatest observed depth, and the existing level computation still owns the later optimizer state.

The change is confined to `optimize.c` and keeps the existing optimizer error path for fail-closed rejection.


infrastation:

